PR #7, opened 2015-02-05
ocf_mail: add mail metadata logging

chriskuehl:
We used to log all mail that went through sandstorm for ~7 (?) days so
that it could be examined if we suspected spam. When I rebuilt the mail
server as anthrax, I didn't re-implement this.

Instead, I added full mail logging for "nomail"'d users, and no logging
for normal users.

This adds more logging without being overly intrusive. It logs time
sent, from/to, sender UID, and subject. Almost all of this information
was already logged by default to /var/log/mail.log, but in an
inconvenient-to-parse-or-search format.

This produces nice JSON logs:

{"date": "2015-02-04T17:40:06.049154",
 "to": [null, "ocfstats@ocf.berkeley.edu"],
 "from": ["Cron Daemon", "root@ocf.berkeley.edu"],
 "uid": "999",
 "subject": "Cron <ocfstats@dementors> /opt/stats/bin/toner > /opt/stats/toner"}

Eventually we can parse this every day or so to automatically flag users
sending too much mail.

The only additional info logged is subject and UID, which I think is a
reasonable compromise for privacy reasons, especially given that we used
to log *everything*.

### DIFF
--- a/modules/ocf_mail/files/site_ocf/aliases
+++ b/modules/ocf_mail/files/site_ocf/aliases
@@ -63,3 +63,4 @@ ocfapt: /dev/null
 
 # misc
 archive: /var/mail/archives/archive
+ocflog: "| /usr/local/bin/log-mail"
--- a/modules/ocf_mail/files/site_ocf/postfix/main.cf
+++ b/modules/ocf_mail/files/site_ocf/postfix/main.cf
@@ -14,6 +14,9 @@ mydestination = int.ocf.berkeley.edu
 virtual_mailbox_domains =
 virtual_alias_domains = ocf.berkeley.edu
 
+# logging
+always_bcc = ocflog
+
 # aliases
 virtual_alias_maps = hash:/etc/postfix/ocf/aliases-local,
 	proxy:ldap:/etc/postfix/ldap-aliases.cf
--- a/modules/ocf_mail/files/spam/logging/log-mail
+++ b/modules/ocf_mail/files/spam/logging/log-mail
@@ -1,0 +1,63 @@
+#!/usr/bin/env python3
+# Appends JSON dictionary of NSA-style metadata about a message read from stdin
+# to /var/log/ocfmail.log for the purposes of monitoring spam. Should be run as
+# the postfix user.
+#
+# Note that `date` returned is the current date (and not one parsed from
+# message headers), since we don't really trust headers more than necessary.
+
+import email
+import email.utils
+import json
+import re
+import sys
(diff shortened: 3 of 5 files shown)


chriskuehl:
Does this seem like a reasonable compromise for everyone? I too don't want to log more info than needed, but we need a better way to recognize users sending spam earlier, and we're still not storing much info at all.

Still to do:
* Maybe log relay host? We can get it from the `Received` header (same place we get uid)
* Prevent sending to anthrax directly on at least death and tsunami, require to use local MTA (sendmail interface) instead. We can use a simple iptables rule with uid matching for this.

matthew-mcallister:
What do we need subjects for? To decide if email is spam or not?

chriskuehl:
@matthew-mcallister yeah, basically. or what kind of spam.

It's useful to see e.g. if the mails are generated by WordPress comments, user registrations, selling viagra, etc.

matthew-mcallister:
I guess it's useful to be able to tell people where their spam is coming from, but is there a solution where we don't log subjects?

chriskuehl:
@matthew-mcallister yes, but I think it's more work than we care for. The solution in that case is to stick the users on the `nomail` list after we notice them sending what might be spam. Then all their mails are rejected and get logged in full (headers, subject, full body). But this is a lot more work, because now we have to wait another day or two for mail to accumulate before we can act on it, plus we've cut off the way we normally communicate with users.

My thought is that staff/SM won't look at the mail log directly, and instead there is some script that runs once a day or something and flags suspected spammers (if any), then looks only at those accounts. Obviously this requires trusting root staffers, but they already have the ability to read user emails, so we have to trust them anyway.

The logs are rotated and shredded after 7 days, so even if we get compromised, only the last 7 days of subjects are revealed (vs. the last ~7? days of full messages before).

chriskuehl:
(the `nomail` list was originally intended for "VIP" groups that we don't want to sorry but are having trouble contacting, if I recall correctly)

matthew-mcallister:
>> My thought is that staff/SM won't look at the mail log directly, and instead there is some script that runs once a day or something and flags suspected spammers (if any), then looks only at those accounts. Obviously this requires trusting root staffers, but they already have the ability to read user emails, so we have to trust them anyway.

My original thought was that accounts sending too much email would be flagged and then we could monitor their mail, but that sounds at least as good.

chriskuehl:
I'm going to go ahead and merge this, but comments are still welcome.

PR #8, opened 2015-03-10
Death webhooks

kpengboy:
This branch adds webhooks for the main website and the hello website. The webhooks update the git repositories, and in the case of hello, the webhook also compiles the stylesheets afterward.

Summary of changes in their original form:
   - Add two webhooks in new directory `/srv/sites/control`
   - Add two deploy scripts in new directory `/opt/puppet`
   - Add two webhook secrets in new directory `/opt/puppet`
   - Install `compass` and `bootstrap-sass` via gem onto the web server (to compile hello) (because the normal packages don't exist/result in missing dependencies)
   - Change Apache configuration accordingly
      - Inline `sites-common.conf`, because now it's only included once
      - Block access to `.git` and `.sass-cache` on hello

So far, the webhooks have seemed to work fine with pushes and force pushes on dev-death. Please give this a look further. (As of this time, Chris has already commented twice.)


kpengboy:
For the main site then, do we want the files to be owned by root or by www-data?

chriskuehl:
Didn't think about this yesterday, but my only remaining qualm is that it seems like declaring `/opt/puppet` in a random manifest is a bad idea. If another manifest also tries to create it to put files there, it will error (even if the definitions are identical). It might be preferable to put the files in `/opt/share`, which is created on all servers by puppet already for basically this reason. You could even put it in `/opt/share/webhook` if you wanted (perhaps we want to add a directory in there for storing the secret files? not sure).

Otherwise LGTM

chriskuehl:
I would also probably squash a few of these commits together, but that's personal preference... up to you.

kpengboy:
I based the layout loosely off of `ocf_puppet`. Perhaps `/opt/share/webhook` is a better idea (on that matter, should `github.cgi` be called `github.py`, given that it's not a front-facing CGI script?)

chriskuehl:
Yeah, it should be renamed, good catch. I'll do that quick. Probably overlooked it since it originally was front-facing cgi.

chriskuehl:
Actually the name is probably ok and helps to remind that  for most purposes it's practically a CGI script since it accepts input and prints output (headers and such) like one.

kpengboy:
Up to the distinction that it requires command line arguments, and no one should in the future mistakenly serve from that directory especially if we put our secrets there too.

chriskuehl:
Good point. I'll rename it.

PR #10, opened 2015-06-30
Version which dynamically updates the time left before logout

kpengboy:
I wrote another locker. This one has a dialog which updates the time remaining before lock/logout. I also cut the warning time from 5 minutes to 1 minute and raised the lock time to 9 minutes. One funky thing is that the dialog has an OK button which is permanently grayed out.
I haven't really tested this thing, mostly because I can't physically go to the lab right now and I don't want to bother with VNC in the next few days.

### DIFF
--- a/modules/ocf_desktop/files/xsession/Xsession
+++ b/modules/ocf_desktop/files/xsession/Xsession
@@ -114,4 +114,4 @@ xset m 0 0
 /usr/local/bin/fix-audio &
 
 # Auto logout inactive users
-/usr/bin/xautolock -locker /usr/local/bin/auto-lock -notify 300 -notifier /usr/local/bin/auto-lock-notify &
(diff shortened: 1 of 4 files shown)


kpengboy:
As of now, avalanche is on my puppet environment if anyone feels like testing it during the week

chriskuehl:
@matthew-mcallister might want to take a look

matthew-mcallister:
I should say I've tested these changes and they work and can be merged.

But, as far as personal preference goes, I think this looks slightly too tacky. Now that I've thought about it, I would rather see more subdued notify-send bubbles instead.

Further, in either case, any popup should say "xautolock" in the title so users know which process to kill to get rid of them, in case they are, e.g., watching a movie.

kpengboy:
About the timeout: we could probably raise it, yes. I'm concerned that 5 minutes is too long, though. Maybe 3 minutes?

chriskuehl:
closing this, was merged

PR #11, opened 2015-07-01
ocf_desktop/xsessionn: Use notify-send for logout warning

matthew-mcallister:
Sorry if I'm getting picky, but I was thinking something like this. Less overt and you don't have to click any buttons to keep your session alive.


kpengboy:
Yeah, I didn't like the buttons too much either. I'm thinking that ideally we would have a script (like a PyGObject script) which creates an always-on-top window saying "You will be logged out in _x_ seconds", where _x_ is dynamically updated, and which disappears when the user moves the mouse or presses a key.

We don't have to be afraid of making 10 pull requests, the whole point is to have a discussion.

PR #13, opened 2015-11-16
Allow logut via XScreenSaver new login button

matthew-mcallister:
If anyone strongly thinks this is a bad idea, the button can be disabled instead.

### DIFF
--- a/modules/ocf_desktop/manifests/xsession.pp
+++ b/modules/ocf_desktop/manifests/xsession.pp
@@ -40,7 +40,10 @@ class ocf_desktop::xsession ($staff = false) {
       purge   => true,
       force   => true,
       backup  => false,
-      source  => 'puppet:///modules/ocf_desktop/xsession/xsessions'
+      source  => 'puppet:///modules/ocf_desktop/xsession/xsessions';
+    # application default settings
+    '/etc/X11/Xresources/x11-ocf':
+      source => 'puppet:///modules/ocf_desktop/xsession/x11-ocf';
   }
 
   # wallpaper symlink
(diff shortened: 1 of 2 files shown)


modules/ocf_desktop/manifests/xsession.pp
```diff
@@ -40,7 +40,10 @@
       purge   => true,
       force   => true,
       backup  => false,
-      source  => 'puppet:///modules/ocf_desktop/xsession/xsessions'
+      source  => 'puppet:///modules/ocf_desktop/xsession/xsessions';
+    # application default settings
+    '/etc/X11/Xresources/x11-ocf':
+      source => 'puppet:///modules/ocf_desktop/xsession/x11-ocf';
```
chriskuehl: if there's only one line maybe just use `content => "XScreenSaver*newLoginCommand:	xfce4-session-logout\n"` rather than adding an entire file. your call.

chriskuehl:
I'm not totally crazy about this, I don't really want some non-staff clicking the button and logging me out from the staff computer (since they won't necessarily realize it's the staff computer, they just want to print).

The code looks fine to me though if others are in favor of it.

modules/ocf_desktop/manifests/xsession.pp
matthew-mcallister: Do you think anything else will ever go in this file?

chriskuehl: maybe. I'm okay with leaving it as-is.

you should probably remove the extra newline at the end though

matthew-mcallister:
I'll close this and make a new commit to disable the button instead if no one wants to argue in favor.

matthew-mcallister:
Disabled it in https://github.com/ocf/puppet/commit/5062b1318ebd08d9d5d369945ad311c442bcb2fc.

PR #14, opened 2015-11-17
Include Chrome OS fonts

daradib:
This should fix many LibreOffice layout issues by providing
metric-compatible replacements for Calibri and Cambria.

Later we can consider removing mscorefonts. Their replacements
have better rendering instructions, but don't cover Comic Sans
etc.

 - Croscore: includes Arimo, Cousine, Tinos
             replaces Arial, Courier New, Times New Roman
             roughly supercedes Red Hat Liberation fonts
 - Caladea:  replaces Cambria
 - Carlito:  replaces Calibri

### DIFF
--- a/modules/ocf_desktop/manifests/packages.pp
+++ b/modules/ocf_desktop/manifests/packages.pp
@@ -15,7 +15,7 @@ class ocf_desktop::packages {
     # desktop
     ['desktop-base', 'desktop-file-utils', 'gpicview', 'xarchiver', 'xterm', 'lightdm', 'accountsservice']:;
     # fonts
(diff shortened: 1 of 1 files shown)


chriskuehl:
sheep it


Does this actually replace any fonts, or just provide alternatives?

daradib:
Debian ships [fonts bindings](https://wiki.archlinux.org/index.php/Font_configuration#Replace_or_set_default_fonts) for fontconfig that offer those fonts as replacements. So, e.g., LibreOffice will fall back to rendering Calibri text with Carlito, since we don't have Calibri. It doesn't otherwise change the doc. You can verify by inspecting the exported PDF.

PR #15, opened 2015-11-19
Use custom lightdm greeter strings

matthew-mcallister:
Change the "incorrect password" string as per [`rt#4130`](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=4130).

### DIFF
--- a/modules/ocf_desktop/files/xsession/lightdm/lightdm-gtk-greeter.po
+++ b/modules/ocf_desktop/files/xsession/lightdm/lightdm-gtk-greeter.po
@@ -1,0 +1,3 @@
+#: ../src/lightdm-gtk-greeter.c:1479
+msgid "Incorrect password, please try again"
+msgstr "Incorrect username or password.\n\nUse your OCF username, not CalNet ID, and\ncheck for spaces around your username.\n\nLost your password? Visit ocf.io/password\nNeed an account? Visit ocf.io/join"
--- a/modules/ocf_desktop/manifests/xsession.pp
+++ b/modules/ocf_desktop/manifests/xsession.pp
@@ -62,7 +62,7 @@ class ocf_desktop::xsession ($staff = false) {
     '/etc/lightdm/lightdm.conf':
       source  => 'puppet:///modules/ocf_desktop/xsession/lightdm/lightdm.conf';
     '/etc/lightdm/lightdm-gtk-greeter.conf':
-        source  => 'puppet:///modules/ocf_desktop/xsession/lightdm/lightdm-gtk-greeter.conf';
+      source  => 'puppet:///modules/ocf_desktop/xsession/lightdm/lightdm-gtk-greeter.conf';
     '/etc/X11/default-display-manager':
       source  => 'puppet:///modules/ocf_desktop/xsession/default-display-manager';
     '/etc/lightdm/session-setup':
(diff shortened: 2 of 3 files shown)


modules/ocf_desktop/manifests/xsession.pp
```diff
@@ -72,6 +72,15 @@
...
       source  => 'puppet:///modules/ocf_desktop/xsession/lightdm/session-cleanup';
+    '/opt/share/xsession/lightdm-gtk-greeter.po':
+      source  => 'puppet:///modules/ocf_desktop/xsession/lightdm/lightdm-gtk-greeter.po';
+  }
+
+  # overwrite greeter strings with OCF ones
+  exec { 'lightdm-greeter-compile-po':
+      command => 'msgcat /opt/share/xsession/lightdm-gtk-greeter.po | msgfmt -o \
+                    /usr/share/locale/en_US/LC_MESSAGES/lightdm-gtk-greeter.mo -',
+      require => File['/opt/share/xsession/lightdm-gtk-greeter.po'];
```
chriskuehl: you should put some kind of restriction on when this exec runs or it will run every time:
https://docs.puppetlabs.com/references/latest/type.html#exec-description

adding `refreshonly => true` and `subscribe => File['/opt/share/xsession/lightdm-gtk-greeter.po']` would work I think

chriskuehl: You may wish to `package {'gettext':;}` and add `depends => Package['gettext']` to guarantee msgcat and friends are installed when this exec runs. not sure how early those get installed when we bootstrap new desktops.

chriskuehl: oh, and probably depend on the greeter package as well

modules/ocf_desktop/files/xsession/lightdm/lightdm-gtk-greeter.po
```diff
@@ -0,0 +1,3 @@
+#: ../src/lightdm-gtk-greeter.c:1479
+msgid "Incorrect password, please try again"
+msgstr "Incorrect username or password."
```
chriskuehl: how much space do we have in the UI? Wondering if we have enough to suggest creating an account or the password reset link. Maybe newlines work?

modules/ocf_desktop/manifests/xsession.pp
matthew-mcallister: Good comments, will amend to address these all.

modules/ocf_desktop/files/xsession/lightdm/lightdm-gtk-greeter.po
matthew-mcallister: It works, looks pretty good. (No screenshot)

matthew-mcallister:
Amended with the requested changes.

modules/ocf_desktop/manifests/xsession.pp
```diff
@@ -72,6 +72,19 @@
...
+
+  # overwrite greeter strings with OCF ones
+  package {'gettext':;}
+
+  exec { 'lightdm-greeter-compile-po':
+      command   => 'msgcat /opt/share/xsession/lightdm-gtk-greeter.po | msgfmt -o \
+                      /usr/share/locale/en_US/LC_MESSAGES/lightdm-gtk-greeter.mo -',
+      subscribe => File['/opt/share/xsession/lightdm-gtk-greeter.po'],
+      refreshonly => true,
+      require   => Package['lightdm-gtk-greeter'];
```
chriskuehl: surprised puppet-lint doesn't complain at the arrow alignment

this should probably require gettext too, right?

`require   => Package['lightdm-gtk-greeter', 'gettext']` should work

matthew-mcallister: Yeah, there are a lot of places where the arrows aren't aligned in this code.

Both corrected anyways.

chriskuehl:
shipit assuming you tested

matthew-mcallister:
Haven't tested on a rebuild from scratch but it seems pretty surefire with the dependencies laid out like that.

modules/ocf_desktop/files/xsession/lightdm/lightdm-gtk-greeter.po
```diff
@@ -0,0 +1,3 @@
+#: ../src/lightdm-gtk-greeter.c:1479
```
matthew-mcallister: I just realized this line will probably break if `lightdm-gtk-greeter` gets updated in the next Debian release.

chriskuehl: I don't think the references are actually significant except to humans?
http://pology.nedohodnik.net/doc/user/en_US/ch-poformat.html

matthew-mcallister: I see now. I found [this page](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html) which says

>  It is possible to have several entries with the same untranslated-string in a PO file, provided that they each have a different context.

but it's referring to a field called `msgctxt`, not the English word.

PR #16, opened 2015-11-21
pre-commit all the things

chriskuehl:
Takes advantage of the pre-commit types fork we're using which lets us target files without extensions.


matthew-mcallister:
pre-commit catches errors and the only other changes are whitespace/quotes, so I say ship it.

PR #17, opened 2015-11-21
When non-staffers try to log into staff computer, explain problem in error message

kpengboy:
The error message tells them they aren't a staffer.

### DIFF
--- a/modules/ocf_desktop/files/xsession/lightdm/lightdm-gtk-greeter-staff.po
+++ b/modules/ocf_desktop/files/xsession/lightdm/lightdm-gtk-greeter-staff.po
@@ -1,0 +1,4 @@
+#: ../src/lightdm-gtk-greeter.c:1479
+msgid "Incorrect password, please try again"
+msgstr "Incorrect username or password,\n"
+"or you're not a staffer."
--- a/modules/ocf_desktop/files/xsession/lightdm/lightdm-gtk-greeter-staff.po
+++ b/modules/ocf_desktop/files/xsession/lightdm/lightdm-gtk-greeter-staff.po
@@ -1,4 +1,6 @@
 #: ../src/lightdm-gtk-greeter.c:1479
 msgid "Incorrect password, please try again"
-msgstr "Incorrect username or password,\n"
-"or you're not a staffer."
+msgstr "Sorry, but only OCF volunteer staffers can\n"
+"log in to this computer. Please use a different\n"
+"computer instead (or if you're actually staff,\n"
+"check your password)."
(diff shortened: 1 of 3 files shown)


kpengboy:
Any comments about the message layout or coding style?

modules/ocf_desktop/files/xsession/lightdm/lightdm-gtk-greeter-staff.po
```diff
@@ -0,0 +1,4 @@
+#: ../src/lightdm-gtk-greeter.c:1479
+msgid "Incorrect password, please try again"
+msgstr "Incorrect username or password,\n"
+"or you're not a staffer."
```
chriskuehl: Can we be more explicit about this? Something like...

"Sorry, but only OCF volunteer staffers can log in to this computer. Please use a different computer instead (or if you actually are staff, check your password)."

Anyone trying to log in to the computer is already clearly not very attentive, so I think making the message slightly longer and more direct would be a plus.

modules/ocf_desktop/manifests/xsession.pp
```diff
@@ -72,17 +72,25 @@
...
 
+  $po = $::staff_only ? {
+    true    => 'lightdm-gtk-greeter-staff.po',
+    default => 'lightdm-gtk-greeter.po',
+  }
+
+  file {
+    "/opt/share/xsession/${po}":
+      source  => "puppet:///modules/ocf_desktop/xsession/lightdm/${po}";
+  }
```
chriskuehl: nit: first lines can be combined like

```puppet
file { "/opt/share/xsession/${po}":
  source  => "puppet:///modules/ocf_desktop/xsession/lightdm/${po}";
}
```
totally optional

modules/ocf_desktop/manifests/xsession.pp
```diff
@@ -72,17 +72,25 @@
...
       mode    => '0755',
       source  => 'puppet:///modules/ocf_desktop/xsession/lightdm/session-cleanup';
-    '/opt/share/xsession/lightdm-gtk-greeter.po':
-      source  => 'puppet:///modules/ocf_desktop/xsession/lightdm/lightdm-gtk-greeter.po';
   }
 
   # overwrite greeter strings with OCF ones
   package {'gettext':;}
 
+  $po = $::staff_only ? {
```
chriskuehl: at some point, we should either rename this to `staffOnly` or rename `skipNetworking` and friends to snake case
(but not related to your PR. i'll ticket)

chriskuehl:
looks good, ship it (optionally make the message more explicit?)

modules/ocf_desktop/manifests/xsession.pp
chriskuehl: rt#4146

chriskuehl:
lg2m

PR #18, opened 2015-11-22
Rename skip variables to use snake case

jvperrin:
Just the simple rename change proposed in [rt #4146](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=4146).


chriskuehl:
thanks! at some point we should also fix ldap-lint (and add the missing vars):
https://github.com/ocf/utils/blob/master/staff/sys/ldap-lint#L23

PR #19, opened 2015-11-29
Add docs shorturl and update stats shorturl

jvperrin:
Add a [ocf.io/docs](https://www.ocf.berkeley.edu/docs/) shorturl for documentation and update the stats shorturl to go to the [new stats page](https://www.ocf.berkeley.edu/stats/) instead of the [old one](http://stats.ocf.berkeley.edu).


chriskuehl:
looks good!

also, good catch remembering to add docs to RESERVED_USERNAMES, maybe we should add a comment warning about that.

PR #20, opened 2015-11-29
Update Travis to use container-based architecture

jvperrin:
See https://docs.travis-ci.com/user/migrating-from-legacy/ for more information on the migration, it sped up the build by about 40 seconds over past ones, and since we don't use sudo at all in the build process, this is a bit cleaner overall. We also won't get that annoying yellow notice now, which is nice. Why do we use Travis for this repository and Jenkins for all the others, or will that be changed?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 ---
 language: python
+sudo: false
 install:
     - pip install ckuehl-pre-commit-types==0.6.2.dev1
 script:


chriskuehl:
yeah, we plan to kill Travis soon and switch to Jenkins on this repo:
https://rt.ocf.berkeley.edu/Ticket/Display.html?id=4145

Travis here pre-dates Jenkins and we just haven't changed it over yet.

The sudo is because we used to `apt-get install puppet` in order to run `puppet parser validate` because we didn't realize at the time we could just install the gem:
https://github.com/ocf/puppet/blob/a9d6d5acff8a3b6d853eb13a81ba7956796cfec3/.travis.yml

...but it's been a while since then anyway. This looks good!

PR #21, opened 2015-12-06
ocf_www: provide short url redirects for /docs/*

matthew-mcallister:


chriskuehl:
we need to reserve any usernames that begin with "docs" as well

chriskuehl:
jk, no we don't
ship it

PR #22, opened 2015-12-12
Remove approve group (rt#4215)

chriskuehl:


matthew-mcallister:
Works on desktops and probably(?) shouldn't break anything, so ship it if ready. Ditto for the other PRs.

PR #25, opened 2016-01-30
Add Korean language support

daradib:
We already support Chinese, Japanese, Cyrillic, Sanskrit,
Persian, Arabic, etc.


chriskuehl:
sheep it

PR #26, opened 2016-01-31
Added redshift to list of packages

Baisang:
use redshift to get rid of nasty blue light from screen by opening a terminal and typing "redshift -O <desired color temperature>" eg. redshift -O 4700

### DIFF
--- a/modules/ocf_desktop/manifests/packages.pp
+++ b/modules/ocf_desktop/manifests/packages.pp
@@ -13,7 +13,7 @@ class ocf_desktop::packages {
     # applications
     ['claws-mail', 'geany', 'filezilla', 'inkscape', 'mssh', 'numlockx', 'remmina', 'simple-scan', 'vlc', 'zenmap', 'gimp', 'gparted', 'evince-gtk', 'galculator', 'xchat', 'hexchat', 'atom']:;
     # desktop
-    ['desktop-base', 'desktop-file-utils', 'gpicview', 'xarchiver', 'xterm', 'lightdm', 'lightdm-gtk-greeter', 'accountsservice']:;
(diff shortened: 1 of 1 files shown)


jvperrin:
Looks good to me, redshift is pretty nice. Might want to add `redshift-gtk` instead (or as well, not quite sure how the dependencies work there) so that it can have a gui to use.

chriskuehl:
@ocfbot add to whitelist

chriskuehl:
add to whitelist

jvperrin:
Cool, I don't think I've met ocfbot before.

chriskuehl:
jenkins, how does that work...

Baisang:
So we tried the gtk version before on staff computer (which has regular redshift installed), but the GUI interface never worked. I think the new package for gtk redshift is gtk-redshift and I haven't tried it

chriskuehl:
ok to test

chriskuehl:
i give up, not sure how to get jenkins to run this

@Baisang it didn't run because you aren't in the ocf group
i just sent you an invite, if you accept it should work in the future

change looks good to me as well, I'll let @jvperrin decide to merge or not

jvperrin:
Ah, I see, eruption has `redshift` and `gtk-redshift` installed already, I didn't realize that. I use `redshift-gtk` on my computer, and it works pretty well and depends on `redshift`, that might work well on the OCF computers too.

jvperrin:
@Baisang Here, I'll install `redshift-gtk` on `eruption` manually and you can test it since you are still there

jvperrin:
Unfortunately, looks like it's only in stretch or sid, so I think it's good to just go with what you have here.

PR #27, opened 2016-02-04
[WIP] ocf_www: rewrite for jessie using Apache module

chriskuehl:
This is still pretty rough but the hard work is done and userdirs seem to be working great.

There are some useful scripts for testing in https://dev-www.ocf.berkeley.edu/~ggroup/

Things that still need to be done (please add more if you notice them):

- [x] Add back shorturl
- [x] Add back accounts.ocf.b.e -> ocfweb redirects
- [x] Add back hello.ocf.b.e -> ocfweb redirects
- [x] Add back wiki.ocf.b.e -> ocfweb redirects
- [x] Add back php.ini custom options (using Augeas ideally to avoid commiting the entire file)
- [x] Add back any APC settings (if we need them? I think we do)
- [x] Add back support for virtual hosts
- [x] Add back NFS host for logs
- [x] Fix log locations to match old ones (or decide not to)
- [x] Probably add back / fix logrotate to ensure proper permissions
- [x] Test that custom php.ini inside homedir is still working 
  - [ ] works for vhosts


modules/ocf_www/templates/vhosts/www.conf.erb
```diff
@@ -1,53 +0,0 @@
...
-
-    SSLEngine on
-	SSLCertificateFile /etc/ssl/private/<%= @fqdn %>.crt
-	SSLCertificateKeyFile /etc/ssl/private/<%= @fqdn %>.key
-	SSLCertificateChainFile /etc/ssl/certs/incommon-intermediate.crt
-	Include ssl-common.conf
-
-    <FilesMatch "\.(cgi|shtml|phtml|php)$">
-        SSLOptions +StdEnvVars
-    </FilesMatch>
```
chriskuehl: i think we need to add this FilesMatch back

modules/ocf_www/templates/vhosts/userdir.conf.erb
```diff
@@ -1,23 +0,0 @@
...
-    <Limit GET POST OPTIONS PROPFIND>
-        Order allow,deny
-        Allow from all
-    </Limit>
-
-    <LimitExcept GET POST OPTIONS PROPFIND>
-        Order deny,allow
-        Deny from all
-    </LimitExcept>
-</Directory>
```
chriskuehl: probably need to add most of the things in this file back?

modules/ocf_www/manifests/init.pp
```diff
@@ -6,303 +18,19 @@
...
   }
 
-  # make sure logratate changes the permissions so that it is viewable when exported
-  file {
-    '/etc/logrotate.d/apache2':
-      ensure => file,
-      source => 'puppet:///modules/ocf_www/logrotate/apache2',
+  class { '::apache':
+    default_vhost => false,
+    mpm_module    => 'worker',
```
chriskuehl: mpm_worker :D

modules/ocf_www/manifests/init.pp
```diff
@@ -6,303 +18,19 @@
...
-  exec { '/usr/sbin/a2enmod headers':
-    unless  => '/bin/readlink -e /etc/apache2/mods-enabled/headers.load',
-    notify  => Service['apache2'],
-    require => Package['apache2'],
-  }
-  exec { '/usr/sbin/a2enmod include':
-    unless  => '/bin/readlink -e /etc/apache2/mods-enabled/include.load',
-    notify  => Service['apache2'],
-    require => Package['apache2'],
-  }
```
chriskuehl: need to add this module back

modules/ocf_www/manifests/init.pp
```diff
@@ -6,303 +18,19 @@
...
-  exec { '/usr/sbin/a2enmod actions':
-    unless  => '/bin/readlink -e /etc/apache2/mods-enabled/actions.load',
-    notify  => Service['apache2'],
-    require => Package['apache2'],
-  }
-  exec { '/usr/sbin/a2enmod expires':
-    unless  => '/bin/readlink -e /etc/apache2/mods-enabled/expires.load',
-    notify  => Service['apache2'],
-    require => Package['apache2'],
-  }
```
chriskuehl: need to add this module back?

matthew-mcallister:
Custom php.ini inside public_html Seems to work.

https://dev-www.ocf.berkeley.edu/~mattmcal/test.php

matthew-mcallister:
Is

> Probably add back / fix logrotate to ensure proper permissions

In reference to 0244de0283d43a42d8e402bfe75b32d7b9df6a6c?

chriskuehl:
Hmm, php.ini does indeed seem to work for some vars (e.g. include_path) but doesn't seem to be working for max_input_vars.

The php.ini here sets `max_input_vars = 1337`.

https://dev-www.ocf.berkeley.edu/~ggroup/phpinfo.php

chriskuehl:
Looks like if you set the configs in a `.user.ini` file instead it will work.

http://php.net/manual/en/configuration.file.per-user.php

This has replaced the per-directory `php.ini` files. Not sure why some settings work in `php.ini` but not all.

We could change the `.user.ini` filename but I think it's better to probably just accept the change since this is the default going forward. We can reach out to users to let them know they may need to rename the file.

chriskuehl:
@matthew-mcallister no, it's not in reference to that commit. Currently we provide a custom logrotate config that sets the new Apache logs to be readable by users. We will probably need to do the same thing lest they be readable only by root.

modules/ocf_www/manifests/init.pp
chriskuehl: done

modules/ocf_www/manifests/init.pp
chriskuehl: done

modules/ocf_www/templates/vhosts/www.conf.erb
chriskuehl: done

modules/ocf_www/templates/vhosts/userdir.conf.erb
chriskuehl: done

chriskuehl:
Let's not fix log locations to match the old ones. I think it's okay to change those and I actually prefer the new ones anyway.

chriskuehl:
retest this please

PR #28, opened 2016-02-09
Add IPv6 addresses to all hosts

chriskuehl:
* Merge the networking modules to reduce a lot of boilerplate (args passing around)
* Make the networking modules more OCF-oriented for the same reason
* No DHCP on the desktops since in practice they have static IPs anyway. This also makes IPv6 easier. We still need DHCP for the initial setup, though, so nothing changes there. This also seems to speed up desktop start since it waits for networking.
* Assign IPv6 addresses to all hosts based on their IPv4 last octet. 169.229.226.x -> 607:f140:8801::1:x

There are three hosts already: raptors, blacksheep, and meteorstorm. The first two have both IPv4 and IPv6 DNS, the last is a desktop and doesn't yet.

Firewall rules are in place to allow outgoing traffic the same way we allow IPv4, so this should be transparent to everyone. We don't have the firewall rules in place yet to have incoming IPv6. The next steps after this are:

1. Add incoming DNS rules for IPv6
2. Add AAAA records to DNS hostnames
3. Add IPv6 reverse DNS


chriskuehl:
will probably take advantage of the downtime tomorrow night / wednesday morning to do a full restart so these networking changes are picked up on

PR #29, opened 2016-02-09
Configure munin email alerts

matthew-mcallister:
The main point of this request is so we get notified if coma/ocfweb starts leaking memory again, and Munin warns for disk overuse as well by default. puppet/facter makes it convenient to configure alerting for all kinds of system variables though, if we should choose to get more alerts.

Munin runs every five minutes and reports changed values on all hosts. The custom notification script filters out messages for anything that is within acceptable bounds, or unknown. Still, this means that, if a value which is out of bounds changes, we'll get emails every time. If this winds up being too spammy, we can implement some kind of message rate limiter in the notification script (or switch to Nagios or something).

### DIFF
--- a/modules/ocf/manifests/munin/node.pp
+++ b/modules/ocf/manifests/munin/node.pp
@@ -18,8 +18,6 @@ class ocf::munin::node {
 
   file { '/etc/munin/plugin-conf.d/ocf-plugin-conf':
     content => template('ocf/munin/ocf-plugin-conf.erb'),
-    owner   => root,
-    group   => root,
     mode    => '0644',
     require => Package['munin-node'];
   }
--- a/modules/ocf/manifests/munin/node.pp
+++ b/modules/ocf/manifests/munin/node.pp
@@ -15,4 +15,12 @@ class ocf::munin::node {
     notify  => Service['munin-node'],
     require => Package['munin-node'];
   }
+
+  file { '/etc/munin/plugin-conf.d/ocf-plugin-conf':
+    content => template('ocf/munin/ocf-plugin-conf.erb'),
+    owner   => root,
+    group   => root,
+    mode    => '0644',
+    require => Package['munin-node'];
+  }
 }
--- a/modules/ocf/templates/munin/ocf-plugin-conf.erb
+++ b/modules/ocf/templates/munin/ocf-plugin-conf.erb
@@ -1,0 +1,3 @@
+[memory]
+env.committed_warning :<%= (0.5 * 1e6 * @memorysize_mb.to_f).to_i %>
+env.committed_critical :<%= (0.8 * 1e6 * @memorysize_mb.to_f).to_i %>
--- a/modules/ocf/templates/munin/ocf-plugin-conf.erb
+++ b/modules/ocf/templates/munin/ocf-plugin-conf.erb
@@ -1,3 +1,4 @@
+# Set warning at 80% and critical at 90%
 [memory]
-env.committed_warning :<%= (0.5 * 1e6 * @memorysize_mb.to_f).to_i %>
-env.committed_critical :<%= (0.8 * 1e6 * @memorysize_mb.to_f).to_i %>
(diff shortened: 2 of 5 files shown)


modules/ocf/templates/munin/ocf-plugin-conf.erb
```diff
@@ -0,0 +1,3 @@
+[memory]
+env.committed_warning :<%= (0.5 * 1e6 * @memorysize_mb.to_f).to_i %>
+env.committed_critical :<%= (0.8 * 1e6 * @memorysize_mb.to_f).to_i %>
```
matthew-mcallister: The choice of limits here is loosely based on graphs from the previous times ocfweb had memory leaks on coma:

http://munin.ocf.berkeley.edu/static/dynazoom.html?plugin_name=ocf.berkeley.edu%2Fcoma.ocf.berkeley.edu%2Fmemory&start_iso8601=2015-12-29T12%3A10%3A54-0800&stop_iso8601=2016-01-10T12%3A10%3A54-0800&start_epoch=1451419854&stop_epoch=1452456654&lower_limit=&upper_limit=&size_x=800&size_y=400&cgiurl_graph=%2Fmunin-cgi%2Fmunin-cgi-graph

These should probably get adjusted to fit our needs more closely.

modules/ocf/manifests/munin/node.pp
```diff
@@ -15,4 +15,12 @@
     notify  => Service['munin-node'],
     require => Package['munin-node'];
   }
+
+  file { '/etc/munin/plugin-conf.d/ocf-plugin-conf':
+    content => template('ocf/munin/ocf-plugin-conf.erb'),
+    owner   => root,
+    group   => root,
```
chriskuehl: can omit owner/group if it's root

modules/ocf/templates/munin/ocf-plugin-conf.erb
chriskuehl: warning at 50% memory seems bad, that means we're planning to waste half of our memory

maybe 80% and 90% is more reasonable?

modules/ocf_stats/files/munin/mail-munin-alert
```diff
@@ -0,0 +1,42 @@
...
+
+
+if __name__ == '__main__':
+    if len(argv) < 2:
+        raise RuntimeError(
+            'Expected command-line arguments; possible misconfiguration')
+    if argv[1] == argv[0]:
+        args = argv[1:]
+    else:
+        args = argv
```
chriskuehl: what is happening here?

modules/ocf_stats/files/munin/munin.conf
```diff
@@ -92,7 +92,7 @@ html_strategy cgi
 #
 # For those with Nagios, the following might come in handy. In addition,
 # the services must be defined in the Nagios server as well.
-#contact.nagios.command /usr/bin/send_nsca nagios.host.comm -c /etc/nsca.conf
+contact.root.command | /usr/local/bin/mail-munin-alert /usr/local/bin/mail-munin-alert ${var:host} ${var:wfields} ${var:cfields}
```
chriskuehl: comment above doesn't match the code below, maybe just remove the comments about nagios?

chriskuehl:
looks good otherwise!

modules/ocf_stats/files/munin/mail-munin-alert
matthew-mcallister: The doc comment reads

> Munin seems to require duplicating the name of the script as the first argument
for some reason; this is stripped out.

I'll add a comment here as well.

modules/ocf_stats/files/munin/munin.conf
matthew-mcallister: FYI I tried removing the superfluous first argument, and Munin stopped sending any arguments at all. The way it appears here is consistent with the [documentation](http://guide.munin-monitoring.org/en/latest/tutorial/alert.html#alerts-to-or-through-external-scripts). I'll look for a fix if one exists.

PR #32, opened 2016-02-21
Add systemd service type

chriskuehl:
I need this for something in my secret Mesos project as well.

Usage looks like this:

```puppet
   ocf::systemd::service { 'docker-registry':                                               
     content => "[Unit]                                                                     
 Description=Mesos Docker registry                                                          
 Requires=network-online.target                                                             
 After=docker.service                                                                       
                                                                                            
 [Service]                                                                                  
 User=root                                                                                  
 ExecStart=/usr/bin/docker run --rm -v /var/lib/registry:/var/lib/registry -p 5000:5000 --name registry registry:2                                                                     
 Restart=always                                                                             
                                                                                            
 [Install]                                                                                  
 WantedBy=multi-user.target                                                                 
 Alias=vnc.service",                                                                        
     require => [                                                                           
       Package['docker.io'],                                                                
       File['/var/lib/registry'],                                                           
     ],                                                                                     
   }                                                                                        

```

### DIFF
--- a/modules/ocf/manifests/packages.pp
+++ b/modules/ocf/manifests/packages.pp
@@ -118,8 +118,10 @@ class ocf::packages {
     }
   }
 
-  if $::lsbdistcodename == 'jessie' {
+  if $::lsbdistcodename != 'wheezy' {
     package {
+      'systemd-sysv':;
+
       'python-paramiko':;
       'python3-paramiko':;
       'python3-ocflib':;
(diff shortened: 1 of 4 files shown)


chriskuehl:
@jvperrin

matthew-mcallister:
Are you sure you don't want a large and complicated class that generates the entire .service file from attributes?

jvperrin:
@chriskuehl Looks good to me, this was exactly what I was thinking of.

PR #34, opened 2016-02-26
LDAP changes for DNS

chriskuehl:
I suspect there are people who want to bikeshed about attribute names, so here you go.

I'll probably merge this later tonight, so you have until then to figure out what to call these new attributes.

### DIFF
--- a/modules/ocf_ldap/files/ocf.schema
+++ b/modules/ocf_ldap/files/ocf.schema
@@ -54,9 +54,23 @@ attributetype ( 1.3.6.1.4.1.41759.2.2.1
         SYNTAX 1.3.6.1.4.1.1466.115.121.1.26
         SINGLE-VALUE )
 
+attributetype ( 1.3.6.1.4.1.41759.2.2.2
+        NAME 'cname'
+        DESC 'DNS CNAME record'
+        EQUALITY caseIgnoreMatch
+        SUBSTR caseIgnoreSubstringsMatch
+        SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+
+attributetype ( 1.3.6.1.4.1.41759.2.2.3
+        NAME 'supportsIPv6'
+        DESC 'Whether the host supports IPv6'
+        EQUALITY booleanMatch
+        SYNTAX 1.3.6.1.4.1.1466.115.121.1.7 )
+
 objectclass ( 1.3.6.1.4.1.41759.2.1.1
         NAME 'ocfDevice'
         DESC 'Attributes for OCF hosts'
         SUP ( ipHost $ ieee802Device $ puppetClient )
         AUXILIARY
-        MUST ( cn $ ipHostNumber $ environment $ type ) )
+        MUST ( cn $ ipHostNumber $ type )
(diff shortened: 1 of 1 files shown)


chriskuehl:
untested and please don't test these yourself; we can't roll-back LDAP changes by just switching Puppet environments. rolling out these changes has to be done carefully.

chriskuehl:
* some hosts don't need a puppet environment (printers, switch, wifi AP)
* some hosts won't support ipv6 ever
* want to put CNAMEs in ldap

modules/ocf_ldap/files/ocf.schema
```diff
@@ -54,9 +54,23 @@ attributetype ( 1.3.6.1.4.1.41759.2.2.1
...
+        SYNTAX 1.3.6.1.4.1.1466.115.121.1.7 )
+
 objectclass ( 1.3.6.1.4.1.41759.2.1.1
         NAME 'ocfDevice'
         DESC 'Attributes for OCF hosts'
         SUP ( ipHost $ ieee802Device $ puppetClient )
         AUXILIARY
-        MUST ( cn $ ipHostNumber $ environment $ type ) )
+        MUST ( cn $ ipHostNumber $ type )
+        MAY ( environment $ cname $ supportsIPv6 ) )
```
chriskuehl: eventually supportsIPv6 would be made mandatory (`MUST`) but we need to do this during the transition period.

chriskuehl: alternatively, we could have no `supportsIPv6` and just know that certain types (probably `printer` and `ipmi`) don't support IPv6.

i think i prefer that solution. and we can always revisit if we end up needing more complexity.

jvperrin: Yeah, I'd leave off `supportsIPv6`, it doesn't really seem like we need it currently unless there's something in particular you would use it for.

jvperrin:
Looks good to me, `cname` seems like a fine name for the attribute. `supportsIPv6` in my opinion isn't necessary yet (unless there's some use case I don't know yet) but could still be added.

modules/ocf_ldap/files/ocf.schema
```diff
@@ -54,9 +54,24 @@ attributetype ( 1.3.6.1.4.1.41759.2.2.1
...
+        SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+
 objectclass ( 1.3.6.1.4.1.41759.2.1.1
         NAME 'ocfDevice'
         DESC 'Attributes for OCF hosts'
         SUP ( ipHost $ ieee802Device $ puppetClient )
         AUXILIARY
-        MUST ( cn $ ipHostNumber $ environment $ type ) )
+        MUST ( cn $ ipHostNumber $ type )
+        MAY ( environment $ dnsCname $ dnsA ) )
```
jvperrin: Using dollar signs as separators is a really strange way for LDAP to separate fields, but I guess that's just how it works.

jvperrin:
Using `dnsCname` and `dnsA` looks good to me. :+1:

PR #35, opened 2016-02-29
ocf_ssh: install system-wide shell environments

matthew-mcallister:
Once this is agreed upon, the next PR will unmount `/opt/ocf` and remove references to it.


modules/ocf_ssh/manifests/environment.pp
```diff
@@ -0,0 +1,12 @@
...
+  file {
+    '/etc/bash.bashrc':
+      source => 'puppet:///modules/ocf_ssh/environment/bash.bashrc';
+    '/etc/csh.cshrc':
+      source => 'puppet:///modules/ocf_ssh/environment/csh.cshrc';
+  }
+
+  exec { 'cat /home/s/st/staff/motd/motd > /etc/motd':
+    onlyif => 'test -f /home/s/st/staff/motd/motd'
+  }
```
chriskuehl: can you just symlink?

chriskuehl:
could we do this on all servers and desktops (so that any server you log in to uses a sane bashrc)?

i'd suggest making `ocf::packages::shells` which does `package { ['zsh', 'csh', 'tcsh', 'bash']:; }` and then adds the proper rc files in the right spot

if not, we should at least do it on apphost as well

modules/ocf_ssh/manifests/environment.pp
matthew-mcallister: Oh duh, NFS wasn't mounted on dev-tsunami at the time of writing.

matthew-mcallister:
I agree and was going to suggest this, since I removed the references to /opt/ocf in my .bashrc and I'm not getting path set on supernova now.

matthew-mcallister:
Replaced old diff.

modules/ocf/manifests/packages/shell.pp
```diff
@@ -0,0 +1,16 @@
+class ocf::packages::shell {
+  package { ['bash', 'csh', 'tcsh', 'zsh']: }
```
chriskuehl: oops, probably don't actually want csh, my bad. tcsh will provide /bin/csh

matthew-mcallister: I was wondering why csh was missing from the list of packages.

modules/ocf/manifests/packages/shell.pp
```diff
@@ -0,0 +1,16 @@
+class ocf::packages::shell {
+  package { ['bash', 'tcsh', 'zsh']: }
+
+  file {
+    '/etc/bash.bashrc':
+      source => 'puppet:///modules/ocf/shell/bash.bashrc';
```
chriskuehl: require => Package['bash']

modules/ocf/manifests/packages/shell.pp
```diff
@@ -0,0 +1,16 @@
...
+    '/etc/csh.cshrc':
+      source => 'puppet:///modules/ocf/shell/csh.cshrc';
```
chriskuehl: require => Package['tcsh']

modules/ocf_ssh/manifests/environment.pp
```diff
@@ -0,0 +1,6 @@
+class ocf_ssh::environment {
+  file { '/etc/motd':
+    ensure => link,
+    target => '/home/s/st/staff/motd',
+  }
+}
```
chriskuehl: suggestion: move to `ocf::motd` and `include ocf::motd` inside `ocf_ssh` and `ocf_apphost`

chriskuehl:
otherwise shipit

modules/ocf/manifests/packages/shell.pp
matthew-mcallister: I suppose, but is there a reason why?

modules/ocf_ssh/manifests/environment.pp
matthew-mcallister: Good idea.

modules/ocf/manifests/packages/shell.pp
chriskuehl: during initial install, your file might be put in place before bash installs, then bash will install and want to replace that file (i think it probably won't for noninteractive installs though, so it's probably nbd)

modules/ocf_ssh/manifests/environment.pp
chriskuehl: maybe also on ocf_admin if you're bored

modules/ocf/manifests/packages/shell.pp
matthew-mcallister: Oh, I didn't realize bash places something there by default. (Maybe not now, but it could in the future)

chriskuehl: it does:

```bash
ckuehl@supernova:~/ocf-proj$ dpkg -L bash | grep '^/etc'
/etc
/etc/bash.bashrc
/etc/skel
/etc/skel/.bash_logout
/etc/skel/.profile
/etc/skel/.bashrc
```

matthew-mcallister:
I also removed remaining references to `/opt/ocf` since they were only in like 2 places.

modules/ocf_ldap/files/slapd.conf
```diff
@@ -57,7 +57,7 @@ access to dn.subtree="ou=People,dc=OCF,dc=Berkeley,dc=EDU" attrs=userPassword
     by * none
 
 # prevent changing own shell when sorried
-access to dn.subtree="ou=People,dc=OCF,dc=Berkeley,dc=EDU" filter=(loginShell=/opt/ocf/bin/sorry) attrs=loginShell
+access to dn.subtree="ou=People,dc=OCF,dc=Berkeley,dc=EDU" filter=(loginShell=/opt/share/utils/sbin/sorry) attrs=loginShell
```
chriskuehl: utils/bin

matthew-mcallister: Actually should be utils/bin/sorried. Fixed in new diff.

PR #37, opened 2016-03-16
Add desktop homepage variable for easy future modification

jvperrin:
I'm thinking that the constant defined in `manifests/site.pp` should be moved to something like `modules/ocf/manifests/homepage.pp` maybe? I don't really like the idea of having a module with just one constant in it though, so I'm not sure if that would be the best idea. Thoughts on organization?

Closes [rt#4431](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=4431).


chriskuehl:
This seems fine to me. It would be tricky to organize any other way especially now that Chrome is used outside of desktops.

jvperrin:
Ok, sounds good, I'll leave it where it is then.

PR #38, opened 2016-03-17
[WIP] Jessie pollution

chriskuehl:
starting a PR with @matthew-mcallister's changes so we can comment on it better


modules/ocf_ssh/manifests/makeservices.pp
```diff
@@ -7,7 +7,7 @@
   file {
     '/opt/share/makeservices':
       ensure => directory,
-      mode   => '0700',
+      mode   => '0500',
```
chriskuehl: is this change meant to be here?

modules/ocf_printhost/manifests/cups.pp
```diff
@@ -0,0 +1,48 @@
+class ocf_printhost::cups {
+  package { [ 'cups', 'cups-bsd' ]: }
+  ocf::repackage { 'hplip':
+    recommends => false
+  }
+  service { 'cups': }
```
chriskuehl: `require => Package['cups', 'cups-bsd']`

modules/ocf_printhost/manifests/cups.pp
```diff
@@ -0,0 +1,48 @@
...
+      content => '# deny printing raw jobs';
+    '/etc/cups/ppd':
+      ensure  => directory,
+      group   => 'lp';
+    ['/etc/cups/ppd/deforestation-double.ppd', '/etc/cups/ppd/logjam-double.ppd']:
+      source  => 'puppet:///modules/ocf_printhost/cups/double.ppd',
+      require => File['/etc/cups/ppd'];
+    ['/etc/cups/ppd/deforestation-single.ppd', '/etc/cups/ppd/logjam-single.ppd']:
+      source  => 'puppet:///modules/ocf_printhost/cups/single.ppd',
+      require => File['/etc/cups/ppd'];
```
chriskuehl: don't need to require parent directories, puppet automatically sets up the dependency (as long as you actually do declare the parent)

(same comment above)

modules/ocf_printhost/manifests/cups.pp
```diff
@@ -0,0 +1,48 @@
...
+    notify  => Service['cups'],
+  }
+
+  file {
+    '/etc/cups/cupsd.conf':
+      content => template('ocf_printhost/cups/cupsd.conf.erb');
+    '/etc/cups/cups-files.conf':
+      content => template('ocf_printhost/cups/cups-files.conf.erb');
+    '/etc/cups/lpoptions':
+      content => 'Default double';
```
chriskuehl: "Default double\n"

modules/ocf_printhost/manifests/cups.pp
```diff
@@ -0,0 +1,48 @@
...
+    ['/etc/cups/raw.convs', '/etc/cups/raw.types']:
+      content => '# deny printing raw jobs';
```
chriskuehl: `"# deny printing raw jobs\n"`

modules/ocf_printhost/manifests/cups.pp
```diff
@@ -0,0 +1,48 @@
...
+    '/etc/cups/ppd':
+      ensure  => directory,
+      group   => 'lp';
```
chriskuehl: no quotes around group names has been our convention

modules/ocf_printhost/files/enforcer-pc
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+gs -o /dev/null -sDEVICE=bbox $1 2>&1 | grep HiResBoundingBox | wc -l
```
chriskuehl: * Can you make the shebang `#!/bin/bash -eu` and add `set -o pipefail` as the second line of the file?
* Double quotes around `$1` just in case it has spaces

chriskuehl: It'd be cool to have tests for this file (e.g. a bunch of postscript files we cobbled together along with expected page count).

That would give us a lot more confidence when changing things in the future.

(but maybe a long-term goal)

modules/ocf_printhost/files/enforcer
```diff
@@ -0,0 +1,182 @@
...
+        parser.add_argument('command', choices={'prehook', 'posthook'})
+        args = parser.parse_args(argv[1:])
+        job = create_job(os.environ)
+        with quota.get_connection(*read_config()) as c:
+            if args.command == 'prehook':
+                prehook(c, job)
+            else:
+                success = os.environ['TEASTATUS'] == '0'
+                posthook(c, job, success)
+    except Exception:
```
chriskuehl: i think `Exception` can be dropped here (just `except:`)

modules/ocf_printhost/files/enforcer
```diff
@@ -0,0 +1,182 @@
...
+    conf.readfp(open(CONF_FILE))
+    mysql_host = conf.get('enforcer', 'user')
+    mysql_passwd = conf.get('enforcer', 'passwd')
+    return mysql_host, mysql_passwd
+
+
+def page_count(path):
+    p = Popen(['/usr/local/bin/enforcer-pc', path], stdout=PIPE)
+    p.wait(30)
+    return int(p.stdout.read())
```
chriskuehl: I think you want

```python
def page_count(path):
    return int(subprocess.check_output(('/usr/local/bin/enforcer-pc', path), timeout=30))
```

modules/ocf_ssh/manifests/makeservices.pp
matthew-mcallister: It shouldn't be. I made `/opt/share/enforcer` write-only (comments on that?) and I suppose committed it to this branch instead of another by mistake.

modules/ocf_printhost/files/enforcer
```diff
@@ -0,0 +1,182 @@
...
+
+
+        """) + MAIL_SIGNATURE
+)
+
+
+def read_config():
+    conf = ConfigParser()
+    conf.readfp(open(CONF_FILE))
+    mysql_host = conf.get('enforcer', 'user')
```
chriskuehl: `mysql_host` comes from `enforcer->user`?

(noticed this last night too, the pollution file is mislabeled as well)

modules/ocf_printhost/files/enforcer
```diff
@@ -0,0 +1,182 @@
...
+            help@ocf.berkeley.edu
+
+
+        """) + MAIL_SIGNATURE
+)
+
+
+def read_config():
+    conf = ConfigParser()
+    conf.readfp(open(CONF_FILE))
```
chriskuehl: `conf.read(CONF_FILE)` probably?

modules/ocf_printhost/manifests/cups.pp
matthew-mcallister: I don't object, but what's the specific reason?

modules/ocf_printhost/files/enforcer
```diff
@@ -0,0 +1,182 @@
...
+    if success:
+        quota.add_job(c, job)
+    else:
+        quo = quota.get_quota(c, job.user)
+        send_printer_mail(PRINTER_ERROR_MESSAGE, job, quo)
+
+
+def main(argv):
+    try:
+        parser = ArgumentParser(description='Pre- and post-hook for Tea4CUPS')
```
chriskuehl: Since you wrote a nice docstring, I'm a big fan of doing...

```python
parser = argparse.ArgumentParser(
    description=__doc__,
    formatter_class=argparse.RawDescriptionHelpFormatter,
)
```

then your nice module docstring will be the program's description

modules/ocf_printhost/manifests/cups.pp
chriskuehl: text files should end with a single newline:
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

(one of my personal pet-peeves. not a huge deal here as long as the tool reads it properly, but especially annoying when people commit actual source code without a newline)

modules/ocf_printhost/files/enforcer-pc
matthew-mcallister: OK, my bad on the bash script, I didn't really think about it because I meant it as a workaround since using pipes in a Python subprocess requires some effort.

I think it makes sense to leave it as a separate script, though. As you suggested, testing would be a valuable benefit, for one thing. It would also be nice to find out a faster way to do the page count, but it probably won't matter since we cap documents to 30 pages. Still, it's way faster to page count PDFs with GhostScript (e.g. can count a 398 page PDF in < 1s while it takes 3s to count a 5 page document with this script. Blame grep).

modules/ocf_printhost/manifests/cups.pp
matthew-mcallister: I figured as much. I'm okay with that.

vim actually puts a trailing newline in every single file you edit, so it's unlikely for us to ever commit an actual file like that.

modules/ocf_printhost/files/enforcer
matthew-mcallister: No, that will catch `SystemExit` exceptions which shouldn't be reported.

modules/ocf_printhost/files/enforcer
matthew-mcallister: That's a pretty neat convention!

modules/ocf_printhost/files/enforcer
matthew-mcallister: copypasta from makemysql

modules/ocf_printhost/files/enforcer
matthew-mcallister: That's what I get for copying from makemysql

modules/ocf_printhost/files/enforcer
```diff
@@ -0,0 +1,243 @@
...
+            'gs',
+            '-q',
+            '-dBATCH',
+            '-dNOPAUSE',
+            '-sDEVICE=ps2write',
+            '-sOutputFile=-',
+            '-sPAPERSIZE=letter',
+            '-dFIXEDMEDIA',
+            '-dPSFitPage',
+            '-dCompatabilityLevel=1.4',
```
matthew-mcallister: This param is unnecessary, can be deleted.

modules/ocf_printhost/files/cups/ppd/logjam.ppd
```diff
@@ -0,0 +1,237 @@
...
+*LanguageLevel: "3"
+*ColorDevice: True
+*DefaultColorSpace: RGB
+*FileSystem: False
+*Throughput: "1"
+*LandscapeOrientation: Plus90
+*TTRasterizer: Type42
+*% Driver-defined attributes...
+*RequiresPageRegion All: True
+*cupsEvenDuplex: True
```
matthew-mcallister: This should be false or else single jobs will get doubly charged

modules/ocf_printhost/manifests/cups.pp
```diff
@@ -0,0 +1,46 @@
+class ocf_printhost::cups {
+  package { ['cups', 'cups-bsd']: }
+  ocf::repackage { 'hplip':
```
matthew-mcallister: We can probably uninstall this now

PR #40, opened 2016-03-19
ocf_desktop: replace ocf-suspend using ocflib

chriskuehl:
Benefits of the change (in decending order of importance):

  * Uses ocflib hours, so we no longer will have desktops running over breaks when we forget to disable wakeups.

  * Adds some output that can be useful for determining why a desktop isn't suspending regularly.

  * Python

There are probably a few bugs in this that we'll have to fix, but hopefully this is an improvement.

### DIFF
--- a/modules/ocf_desktop/files/suspend/ocf-suspend
+++ b/modules/ocf_desktop/files/suspend/ocf-suspend
@@ -1,28 +1,81 @@
-#!/bin/bash
-# Suspend the machine and wake on LAN.
+#!/usr/bin/env python3
+"""Suspend this desktop if not in use and lab is closed.
 
-# Don't sleep if...
-# installing packages
-! /usr/bin/lsof -- /var/lib/dpkg/lock > /dev/null 2>&1 || exit 1
-# puppet is running
-[ ! -f /var/lib/puppet/state/agent_catalog_run.lock ] || exit 2
-# user present
-[ $(w --no-header | wc -l) -eq 0 ] || exit 3
-# is a virtual machine
-imvirt | grep -qi Physical || exit 4
+Contains a few checks (is anyone logged in, is Puppet running, etc.) before
+shutting off.
 
-# Try to unmount /home and /tmp (both tmpfs)
-if grep $'tmpfs\t/home' /etc/fstab > /dev/null; then
-	umount /home 2> /dev/null && mount /home || exit 255
-fi
-umount /tmp 2> /dev/null && mount /tmp
+If a kernel update is needed, the desktop will restart instead of suspending.
+"""
+import argparse
+import os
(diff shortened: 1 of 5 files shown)


modules/ocf_desktop/files/suspend/ocf-suspend
```diff
@@ -1,28 +1,85 @@
...
+
+from ocflib.lab.hours import Day
+
+
+devnull = open('/dev/null', 'w')
+
+
+def file_is_open(path):
+    """Return whether a file is open by any process on the system."""
+    return subprocess.call(('lsof', '--', path), stdout=devnull, stderr=devnull) == 0
```
kpengboy: You can use `subprocess.DEVNULL` instead

chriskuehl: good catch, forgot we're on python3.4 now
added in a new commit

modules/ocf_desktop/files/suspend/ocf-suspend
```diff
@@ -1,28 +1,83 @@
...
+
+def main(argv=None):
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument('--force', '-f', action='store_true', help='force suspend, even if the lab is open')
+    parser.add_argument('--quiet', '-q', action='store_true', help='stay quiet except for unexpected errors')
+    args = parser.parse_args(argv)
+
+    def report(line):
+        if not args.quiet:
+            print('[ocf-suspend] {}'.format(line), file=sys.stderr)
```
kpengboy: ...wait, do we have other commands doing it this way? Most programs I've seen use the `programname: errormessage` format

chriskuehl: shrug

kpengboy:
lg2m, but maybe I missed something

chriskuehl:
thanks for quick review

i'm guessing there are still some bugs (or at least some commands that can fail and cause cron spam that we don't want), so we'll just iterate on this until it's working...

modules/ocf_dhcp/manifests/init.pp
```diff
@@ -30,27 +30,18 @@
...
-      minute  => 0,
-      weekday => 6,
-      require => File['/usr/local/bin/lab-wakeup'];
-    'lab-wakeup-sunday':
-      command => '/usr/local/bin/lab-wakeup > /dev/null',
-      hour    => 12,
-      minute  => 0,
-      weekday => 0,
+    'lab-wakeup':
+      command => '/usr/local/bin/lab-wakeup',
```
kpengboy: You meant `lab-wakeup -q`.

PR #41, opened 2016-03-19
Switch from Percona to MariaDB (rt#4527)

chriskuehl:


modules/ocf_mysql/manifests/init.pp
```diff
@@ -1,31 +1,60 @@
...
+
+      # http://haydenjames.io/mysql-query-cache-size-performance/
+      "set target[.='mysqld']/query_cache_size 32M",
+
+      # https://mariadb.com/kb/en/mariadb/xtradbinnodb-server-system-variables/
+      # https://dev.mysql.com/doc/refman/5.6/en/innodb-multiple-buffer-pools.html
+      "set target[.='mysqld']/innodb_file_per_table 1",
+      "set target[.='mysqld']/innodb_buffer_pool_size 8G",
+      "set target[.='mysqld']/innodb_buffer_pool_instances 8",
+      "set target[.='mysqld']/innodb_flush_log_at_trx_commit 1",
```
nattofriends: This is default. I'm assuming you set this value here to overwrite because 1) Puppet will not delete the file 2) Dpkg will not delete the file?

chriskuehl: ah, I'll remove it, didn't realize it was the default.

the new mariadb will be on a new server (not the old one) so the old values aren't a big deal

PR #42, opened 2016-03-21
Replace Iceweasel with Firefox

daradib:
After almost 10 years, Firefox is back in Debian unstable.
Replace iceweasel package with firefox from mozilla.debian.net.
Also use the default Firefox icon for desktop launcher.

We are now again tracking the regular Firefox release version
rather than ESR as Iceweasel does.

 - On desktops and ocf_tv, iceweasel is removed and replaced with firefox.
 - On user-facing servers, iceweasel has been removed from Puppet
   but won't be purged yet, pending discussion.


modules/ocf/manifests/apt.pp
```diff
@@ -92,6 +92,19 @@
       notify => Exec['apt_update'];
     }
 
+    package { 'pkg-mozilla-archive-keyring':; }
+
+    apt::source {
+      'mozilla':
+        location => 'http://mozilla.debian.net/',
+        release  => "${::lsbdistcodename}-backports",
+        repos    => 'firefox-release',
```
matthew-mcallister: Getting off ESR and on release will be really nice.

modules/ocf_desktop/manifests/firefox.pp
```diff
@@ -0,0 +1,19 @@
...
+    # disable caching, history, blacklisting, and set homepage
+    '/etc/firefox/profile/prefs.js':
+      content => template('ocf_desktop/firefox/prefs.js.erb'),
+      require => Package['firefox'];
+    # start maximized by default
+    '/etc/firefox/profile/localstore.rdf':
+      source  => 'puppet:///modules/ocf_desktop/firefox/localstore.rdf',
+      require => Package['firefox']
+  }
+}
```
chriskuehl: I'd suggest moving this to `ocf::packages::firefox` since we include it outside of desktops (on the TV).

Then in extrapackages we can just `include ocf::packages::firefox` like we do for Chrome (though I think we don't include it in extrapackages?)

modules/ocf/manifests/apt.pp
daradib: Agreed. We're currently using the old ESR. The new ESR was [released a few days ago](https://www.mozilla.org/en-US/firefox/45.0esr/releasenotes/) and is based on Firefox 45, the latest release.

modules/ocf_desktop/manifests/firefox.pp
daradib: @chriskuehl I agree. Will move to `ocf::packages::firefox` for consistency with Chrome. However, ocf_tv and servers don't include mozilla.debian.net and google-chrome sources.list, so it won't work in extrapackages.

modules/ocf/manifests/packages/firefox.pp
```diff
@@ -0,0 +1,19 @@
...
+  package {
+    'firefox':
+      ensure => present;
+    'iceweasel':
+      ensure => absent;
+  }
+
+  file {
+    # disable caching, history, blacklisting, and set homepage
+    '/etc/firefox/profile/prefs.js':
```
matthew-mcallister: Does `/etc/firefox/profile` still exist?

daradib: From [Debian Bug #815006](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=815006#110):

> Speaking of conffiles, version 46 is removing /etc/$package/profile.
> I'm very tempted to remove it in 45 because that would only leave
> /etc/iceweasel/pref/*.js to migrate, and would keep upgrade code free of
> conffiles removal entirely. This would allow to simply use
> /etc/firefox{,-esr}/*.js files.

Looks like it was changed with Firefox 45. No idea about where to put `localstore.rdf` or if we have an alternative way to always open maximized.

matthew-mcallister: Well, for one, [localstore.rdf seems to be deprecated](https://support.mozilla.org/en-US/kb/firefox-support-troubleshooting-guide#w_reset-xulstore-or-localstore-file). I suppose we should move be using xulstore.json instead.

Most likely, we can just put one of the two in `/etc/firefox`, which is where prefs.js goes now. Test it on a lab computer.

daradib: Sounds good. Will do next time in lab.

daradib:
Rebased to incorporate #43, added firefox to ocf_tv, updated prefs for newer versions, and tested on acid. I think we should be good to merge.

chriskuehl:
cool, i vote to merge

modules/ocf/templates/firefox/prefs.js.erb
```diff
@@ -0,0 +1,12 @@
+lockPref("browser.startup.homepage", "<%= @desktop_homepage %>");
+pref("browser.cache.disk.capacity", 0);
+pref("browser.download.useDownloadDir", false);
```
chriskuehl: don't we want this true?

> True (default): Without asking a location, download files to the directory specified in browser.download.folderList

I think the Downloads folder is pretty standard at this point, people are going to get confused when they accidentally end up in `/` or something.

chriskuehl: (i realize this is just a renamed file in this PR, though. don't let this block merging.)

PR #43, opened 2016-03-21
Stop treating desktops as special case for apt (rt#4639)

chriskuehl:
This way any host can `include ocf::packages::chrome` and the proper sources get added, without adding more hackery into `ocf::apt`'s optional arguments.

Will do something similar for Firefox (ref #42).

paging @daradib

### DIFF
--- a/modules/ocf/manifests/packages/chrome.pp
+++ b/modules/ocf/manifests/packages/chrome.pp
@@ -1,4 +1,8 @@
 class ocf::packages::chrome {
+  class { 'ocf::packages::chrome::apt':
+    stage => first,
+  }
+
   package { 'google-chrome-stable':; }
 
   file {
@@ -14,6 +18,8 @@ class ocf::packages::chrome {
       source  => 'puppet:///modules/ocf/chrome/master_preferences',
       require => Package['google-chrome-stable'];
 
+    # Disable Google's broken cronjob which sets up the wrong apt sources
+    # (rt#4589).
     '/opt/google/chrome/cron/google-chrome':
       mode    => '0755',
       content => "#!/bin/sh\n# Disabled by OCF (rt#4589)\n",
--- a/modules/ocf_desktop/manifests/drivers.pp
+++ b/modules/ocf_desktop/manifests/drivers.pp
@@ -1,8 +1,4 @@
 class ocf_desktop::drivers {
-  class { 'ocf::apt::i386':
-    stage => first,
-  }
-
   # install proprietary nvidia drivers
   if $::gfx_brand == 'nvidia' {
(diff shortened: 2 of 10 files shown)


modules/ocf/manifests/packages/chrome.pp
```diff
@@ -1,4 +1,8 @@
 class ocf::packages::chrome {
+  class { 'ocf::packages::chrome::apt':
```
daradib: ??

chriskuehl: the apt changes needs to happen in `stage => first` and puppet kinda sucks

(we used to follow this same pattern for percona until last night)

daradib: Yeah, I understand now. Thanks for clarification. I set up that `stage => first` hack 4 years ago and it's ugly.

modules/ocf_desktop/manifests/drivers.pp
```diff
@@ -1,4 +1,8 @@
 class ocf_desktop::drivers {
+  class { 'ocf::apt::i386':
```
daradib: Should this be moved to `init.pp`? We can only define it once even though it's needed for other desktop things like steam.

chriskuehl: yeah, i think that makes sense. i moved it from steam to drivers because it seemed better, but init is probably better than both

puppet sucks sometimes

daradib:
I think this is the right direction. The main problem is that we can only include these extra manifests like `ocf::apt::i386` once per server. Can we `stage => first` individual resources like `exec` so that we can simply `include ocf::apt::i386` instead of define it with `stage => first` parameter?

chriskuehl:
@daradib I would love to but unfortunately it doesn't work:

> Mar 20 18:08:44 destruction puppet-agent[24299]: Could not retrieve catalog from remote server: Error 400 on SERVER: Puppet::Parser::AST::Resource failed with error ArgumentError: Only classes can set 'stage'; normal resources like Apt::Key[google] cannot change run stage at /opt/puppet/env/ckuehl/modules/ocf/manifests/packages/chrome.pp:6 on node destruction.ocf.berkeley.edu

chriskuehl:
I imagine you could use ensure_resource as a workaround from stdlib :\

daradib:
Can you include a wrapper class that only calls `stage => first` on another class as a workaround?

chriskuehl:
@daradib yeah, but that kinda sucks too :\

chriskuehl:
on second thought, we basically have that already for chrome (`ocf::packages::chrome` includes `ocf::packages::chrome::apt` with `stage => first`)

i'll just add one for i386 to be done with it

daradib:
+1

PR #46, opened 2016-04-01
[WIP] Enable per-user systemd for apphosting

kpengboy:


modules/ocf_apphost/manifests/systemd_linger.pp
```diff
@@ -0,0 +1,6 @@
+# Declares that the following user is to have a persistent per-user systemd
+define ocf_apphost::systemd_linger($user = $title) {
+  file { "/var/lib/systemd/linger/${user}":
```
kpengboy: The way this is enabling lingering is by manually sticking files in the `/var/lib/systemd/linger/` directory. This makes it straightforward for Puppet to disable lingering for users who are removed from ocfdev. But it's also some of an abstraction barrier violation, and I can't be certain it will work forever.

The other way to do it is by exec'ing `loginctl enable-linger ${user}` as the manual says. This is maybe more robust, but then it's not easy to disable lingering for users who get removed from ocfdev.

What do you think?

chriskuehl: I think this is acceptable. It's not going to break between Debian releases, and we'll do careful testing on release upgrades anyway.

Agree with your concerns about not using the proper interface, but I think this is better than trying to script that via Puppet.

chriskuehl:
Just documenting some concerns:

* need a way to make sure the services are only started on one server
* need to update documentation when we push this (or shortly after)
* need to contact groups afterwards

Does journalctl work on these services? That would be really handy for groups to see logs.

matthew-mcallister:
The man page shows a parameter `--user-unit=` which seems to do just that.

chriskuehl:
Also, I'm assuming you tested this, but are there race conditions where e.g. NFS hasn't mounted yet during boot and so user services aren't started?

kpengboy:
Probably no NFS race condition - per-user systemd instances seem to be started after `systemd-user-sessions.service` is started, which seems to be after remote FSes are mounted.

How do we make NFS slower?

modules/ocf_apphost/manifests/init.pp
```diff
@@ -9,8 +9,17 @@
     cron => true;
   }
 
-  # useful for supervising user processes
-  package { 'daemontools':; }
+  # enable a persistent per-user systemd for all ocfdev users
+  file { '/var/lib/systemd/linger':
+    ensure  => directory,
+    # This makes sure only ocfdev users get per-user systemd. TODO: is this
```
kpengboy: I think that only allowing ocfdev users to get per-user systemd on the apphost is reasonable. What do others think?

chriskuehl: i think this sounds good.

only concern is that everyone still gets a user systemd instance when they log in (e.g. look at tsunami, there are tons of them; they just don't linger) so we need to be careful about spawning duplicate services. sounds like you have a solution for that already though.

PR #47, opened 2016-04-20
Add LICENSE, as approved by BoD

chriskuehl:
This is my attempt to format the LICENSE file as approved by BoD tonight and last week.

I will merge this (and add the same file to other repos) if nobody objects (and ideally if some others confirm this looks correct to them).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,1170 @@
+The Open Computing Facility's directors and members believe strongly in
+the importance of free and open-source software, and we hope for
+software and other works developed by our organization to be used to
+advance the free software movement.
+
+We have spent a considerable amount of time discussing which licenses
+best reflect our organization's mission, with a roughly even split among
+members who believe that a copyleft license like the GNU GPL best
+promotes the movement's interests, and those who believe that a
+permissive license like the Apache license promotes them better.
+
+The Board of Directors has agreed to dual-license our code under both
+the GNU GPLv3 and the Apache 2.0 licenses. While we recognize that
+dual-licensing our code negates much of the power of copyleft licenses
+like the GNU GPL, we chose to dual-license in an attempt to respect the
+legitimate divide in our organization regarding software licensing.
+
+You may choose to use the code assets in this repository under either
+the GNU GPLv3 or Apache 2.0 license (both printed in full below). The
+non-code assets (text, images, etc.) are licensed under the Creative
+Commons Attribution 4.0 International ("CC BY 4.0") license. A
+human-readable summary of the CC BY 4.0 is available at:
+https://creativecommons.org/licenses/by/4.0/
+
+All copyrights are owned by the original contributors. Contributors are
+required to sign the Open Computing Facility's CLA, which allows the Board
+to license their contributions.
+
+Note that while we have attempted to identify which files in this
+repository do not contain exclusively OCF contributions (and thus are
+not necessarily licensed under the terms above), it is your
+responsibility to confirm this.
+
+
+====================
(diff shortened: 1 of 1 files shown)


jvperrin:
Yup, looks good to me!

LICENSE
```diff
@@ -0,0 +1,1170 @@
...
+https://creativecommons.org/licenses/by/4.0/
+
+All copyrights are owned by the original contributors. Contributors are
+asked to sign the Open Computing Facility's CLA, which allows the Board
+to license their contributions.
+
+Note that while we have attempted to identify which files in this
+repository do not contain exclusively OCF contributions (and thus are
+not necessarily licensed under the terms above), it is your
+responsibility to confirm this.
```
matthew-mcallister: This sounds like a legally grey area to me. I don't know of a better solution off the top of my head, unfortunately, but it might be a good idea to rephrase this to sound less dubious. Quoth dwc:

   > \<dwc> how does that work? "we have stuff in here that doesn't match the license that we've taken from somewhere and reproduced without its accompanying licensing terms?"
   > \<dwc> "... it is left as an exercise to the reader to guess!"

LICENSE
```diff
@@ -0,0 +1,1170 @@
...
+You may choose to use the code assets in this repository under either
+the GPLv3 or Apache 2.0 license (both printed in full below). The
+non-code assets (text, images, etc.) are licensed under the Creative
+Commons Attribution 4.0 International ("CC BY 4.0") license. A
+human-readable summary of the CC BY 4.0 is available at:
+https://creativecommons.org/licenses/by/4.0/
+
+All copyrights are owned by the original contributors. Contributors are
+asked to sign the Open Computing Facility's CLA, which allows the Board
+to license their contributions.
```
matthew-mcallister: Are we _asked_ to sign the CLA or _required_? It sounds as if there may be contributions in here to which we do not have explicit approval to apply a license.

LICENSE
```diff
@@ -0,0 +1,1170 @@
+The Open Computing Facility's directors and members believe strongly in
+the importance of free and open-source software, and we hope for
+software and other works developed by our organization to be used to
+advance the free software movement.
+
+We have spent a considerable amount of time discussing which licenses
+best reflect our organization's mission, with a roughly even split among
+members who believe that a copyleft license like the GPL best promotes
```
matthew-mcallister: There are a few instances of the bare term "GPL" which should be replaced with "GNU GPL" or "GPLv3". "GPL" stands for "General Public License" and, without any qualifier, doesn't technically specify any license. This isn't an issue in informal conversation but we should be as clear as possible in our license.

LICENSE
chriskuehl: This is the case with all license grants. Most projects are including code which is not properly licensed/attributed, unfortunately. We can't grant rights to other peoples' work; thus we've done our best to identify it, but if we've messed up, we can't be held responsible for you using it.

The fix is to be extra careful with including others' code and make sure attribution is properly done (as we should/must be doing already). I think our effort is better spent here.

LICENSE
chriskuehl: We've included the entire text of the GPL v3, I don't think there's any ambiguity here...

LICENSE
chriskuehl: updated to "required"

LICENSE
matthew-mcallister: I don't think the language is ambiguous, it is just noticeably informal.

chriskuehl:
okay, it says GNU GPL now

PR #48, opened 2016-04-21
Install debian-security-support

daradib:
Package provides check-support-status which lists official
packages that are missing security support.

Unfortunately this includes V8, used by Node.js.
https://security-tracker.debian.org/tracker/source-package/libv8


chriskuehl:
nobody expected security from nodejs anyway :P

i vote ship

daradib:
In fairness, Node.js is used to run "trusted" code, not some arbitrary webpage. Ok merging.

PR #50, opened 2016-05-12
ocf_mirrors: add kali, parrot, tails, trisquel

chriskuehl:
@kpengboy @tmochida please verify if kali and trisquel look correct to you

most are still doing the initial sync, but you should be able to tell if it's syncing from/to the correct place by looking at https://mirrors.ocf.berkeley.edu/


tmochida:
Kali looks good, thanks for the changes.

matthew-mcallister:
So far trisquel, kali, parrot, and tails all seem to be successfully syncing and look correct according to the instructions on their websites.

PR #51, opened 2016-05-20
Rewrite clean-temp-files to be more reliable

jvperrin:
I've rewritten `clean-temp-files` in Python, it now reads last month's login records too to give the correct last login date for users, even when it's close to the start of the month. It also makes sure that logged in users never have their files deleted out from under them, which could have been a problem before at the start of each month.

Resolves rt#4805.

TODO: It might be good to potentially delete directories based on space taken up? This might be nice for steam directories, which take up a lot of space. They could be deleted after a week, for instance, instead of the current 2 weeks of delay if we find it is needed (like we keep getting Munin alerts about disk usage).

Any suggestions for making the script any better/cleaner?

### DIFF
--- a/modules/ocf_desktop/files/clean-temp-files
+++ b/modules/ocf_desktop/files/clean-temp-files
@@ -1,19 +1,71 @@
-#!/bin/bash -eu
-# Find and remove temporary files (/var/local/tmp/*) from users who haven't
-# logged in recently.
-#
-# For simplicity, /var/local/tmp/ is owned (and only writable) by root. On
-# login, root creates a new directory for the current user to use. We remove
-# these directories periodically via this cron script.
-set -o pipefail
-shopt -s nullglob
+#!/usr/bin/env python3
+"""Clean temporary files on desktops if users owning the files have not logged
+in for 2 weeks.
 
-cd /var/local/tmp
-for u in *; do
-	# TODO: this check for recent logins is not great; wtmp is rotated monthly
-	# (rather than being a rolling log), so this basically is equivalent to
-	# clearing everyone at the start of the month
-	if ! last -Rw | cut -d' ' -f1 | head -n-2 | grep -q "^$u$"; then
-		rm -rf --one-file-system -- "/var/local/tmp/$u"
-	fi
-done
+Also makes sure that files are not deleted under a user if they are currently
+logged in to the desktop.
+"""
+import os
+import subprocess
+from datetime import datetime
+from datetime import timedelta
+
+from dateutil import parser
+
+
+def get_command_by_lines(command):
+    """Get the output of a command split by lines"""
+    return subprocess.check_output(command).decode('utf-8').splitlines()
+
+
+def get_current_users():
+    """Get a set of the current users logged in"""
+    who = get_command_by_lines('who')
+
+    return {user.split()[0] for user in who}
+
+
+def get_last_logins():
+    """Get a dictionary of the last time each user logged in to this desktop by
+    parsing the "last" command's output. last uses /var/log/wtmp, which is
+    rotated at the start of each month, so we have to make sure to also check
+    /var/log/wtmp.1 (last month's records) in case it is near the start of the
+    month.
+    """
+    curr_users = get_current_users()
+
+    this_month = get_command_by_lines(('last', '-wF'))
+    last_month = get_command_by_lines(('last', '-wF', '-f', '/var/log/wtmp.1'))
+
+    users = {}
+    ignored_names = ['reboot', 'apt-dater', 'wtmp', 'wtmp.1']
+
+    for line in reversed(this_month + last_month):
+        session = line.split()
+
+        if session:
+            username = session[0]
+
+            if username in curr_users:
+                users[username] = datetime.now()
+            elif username in ignored_names:
+                continue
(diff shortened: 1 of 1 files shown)


modules/ocf_desktop/files/clean-temp-files
```diff
@@ -1,19 +1,71 @@
...
+    """Get a dictionary of the last time each user logged in to this desktop by
+    parsing the "last" command's output. last uses /var/log/wtmp, which is
+    rotated at the start of each month, so we have to make sure to also check
+    /var/log/wtmp.1 (last month's records) in case it is near the start of the
+    month.
+    """
+    curr_users = get_current_users()
+
+    this_month = get_command_by_lines(('last', '-wF'))
+    last_month = get_command_by_lines(('last', '-wF', '-f', '/var/log/wtmp.1'))
```
chriskuehl: will this fail on a new desktop which hasn't been running >= one month?

modules/ocf_desktop/files/clean-temp-files
```diff
@@ -1,19 +1,71 @@
...
+
+    users = {}
+    ignored_names = ['reboot', 'apt-dater', 'wtmp', 'wtmp.1']
```
chriskuehl: set?

modules/ocf_desktop/files/clean-temp-files
```diff
@@ -1,19 +1,71 @@
...
+        session = line.split()
+
+        if session:
+            username = session[0]
+
+            if username in curr_users:
+                users[username] = datetime.now()
+            elif username in ignored_names:
+                continue
+            else:
```
chriskuehl: seems redundant to continue in an if-tree where everything else is under an else :)

maybe just replace with `elif username not in ignored_names:` ?

modules/ocf_desktop/files/clean-temp-files
```diff
@@ -1,19 +1,71 @@
...
+                session_start = ' '.join(session[3:8])
+                users[username] = parser.parse(session_start)
+
+    return users
+
+
+def main(argv=None):
+    users = get_last_logins()
+    directories = os.listdir('/var/local/tmp')
+    cutoff = datetime.now() + timedelta(weeks=-2)
```
chriskuehl: `datetime.now() - timedelta(weeks=2)` would probably look less derpy :P

modules/ocf_desktop/files/clean-temp-files
jvperrin: Hmm, that's a good point, I hadn't considered that, it probably would have some problems, I should check if the file exists first.

modules/ocf_desktop/files/clean-temp-files
jvperrin: Yeah, that would be better, sounds good.

modules/ocf_desktop/files/clean-temp-files
```diff
@@ -1,19 +1,71 @@
...
+
+    for user in directories:
+        if user not in users.keys() or users[user] < cutoff:
+            subprocess.call(('rm', '-rf', '--one-file-system', '--', '/var/local/tmp/' + user))
```
chriskuehl: could you add `assert all(c in string.ascii_lowercase for c in user), user`? I'd feel safer if we were sure this was only lowercase letters (and that somehow other stuff didn't creep in, e.g. due to parsing errors)

chriskuehl:
overall looks solid to me!

modules/ocf_desktop/files/clean-temp-files
jvperrin: Yeah, that's a good idea, I was thinking this could be a problem, since using `rm -rf` is pretty dangerous, so adding some extra checks would be a great plan.

modules/ocf_desktop/files/clean-temp-files
jvperrin: Haha yeah, good point, I'll do that

modules/ocf_desktop/files/clean-temp-files
chriskuehl: you might also `check_call` so it raises an exception if the `rm` fails for some reason

jvperrin:
@chriskuehl Ok, thanks for all the suggestions, I used them all to clean up the script some more and tested it on a couple desktops (acid and blizzard), it seems to be working well.

chriskuehl:
lgtm!


side note: jenkins broken somehow? https://jenkins.ocf.berkeley.edu/job/puppet-test-pr/61/console

jvperrin:
Actually looks like it failed due to a change in the Jenkins pull request builder plugin:
https://wiki.jenkins-ci.org/display/JENKINS/Plugins+affected+by+fix+for+SECURITY-170
https://wiki.jenkins-ci.org/display/SECURITY/Jenkins+Security+Advisory+2016-05-11

I upgraded jenkins with apt-dater from 2.1 to 2.3 on May 13th, so that's probably when things started breaking for pull request builds.

Looks like this can be fixed by editing `/etc/default/jenkins` and adding some parameters there, at least until the plugin is fixed: janinko/ghprb#408

jvperrin:
Retest this please

PR #52, opened 2016-05-27
Custom puppet reporting

matthew-mcallister:
The idea is to process reports from production hosts every half hour to reduce the number of puppet emails sent out. For immediate feedback, and for development environments, failed builds will be reported immediately via IRC.

Until the latter feature is ready, this doesn't have to be merged, but it's here for looking.

### DIFF
--- a/modules/ocf_puppet/files/puppet-report
+++ b/modules/ocf_puppet/files/puppet-report
@@ -1,0 +1,181 @@
+#!/usr/bin/env python3
+"""A custom batch report processor for Puppet.
+
+Features of the processor
+* Doesn't email reports for nodes not on the production environment
+* Can send reports for multiple hosts in a single digest-like report
+  with the '--digest' option
+* Doesn't email reports which only contain the "switching agent
+  environment" message
+* Sends puppet reports for everything else like tagmail does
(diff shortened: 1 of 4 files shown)


modules/ocf_puppet/files/puppet-report
```diff
@@ -0,0 +1,181 @@
...
+)
+yaml.add_constructor(
+    u'!ruby/object:Puppet::Util::Log',
+    NamedtupleConstructor(PuppetLog),
+    Loader=yaml.SafeLoader
+)
+
+# Treat all other tagged objects like plain YAML
+# Do this by removing the default constructor, which throws an error
+del yaml.SafeLoader.yaml_constructors[None]
```
chriskuehl: this (and the stuff above) is kinda sad, this has global impact within the python process (i.e. `yaml` in another module is impacted). is there any way to avoid that?

matthew-mcallister: Sure, there are multiple ways. This was just the absolute most
straightforward.

matthew-mcallister:
Closing. I would try PuppetDB first if we really want this.

PR #53, opened 2016-05-29
Productionize mesos

chriskuehl:
This sets up a Mesos master on each of the three physical servers, as well as a Mesos slave and Docker.

Zookeeper is used to provide high-availability for Mesos and for Marathon, which is a Mesos framework that we can use for long-running services in Docker containers.

This also adds a load balancer (haproxy) which runs using [marathon-lb](https://github.com/mesosphere/marathon-lb) inside a Docker container on each physical server. This means that we can assign each service a unique port, and will then always be able to reach it by connecting to that same port on any of the physical servers. (The load balancer will round-robin between any of the several Docker containers hosting that service.)

An example usage would be making a Docker container for ocfweb. death would be configured to proxy to the well-known ocfweb port on any physical server, rather than always proxying to coma like it does now.

With both Mesos and Marathon, only one server will be the leader at any time. In order for either to operate, a majority is needed (more than half of all servers). So in practice, this means we could turn off one server at a time and still have our services working.

You can connect to either web UI, and you'll just get redirected to the current leader if you picked the wrong one.

Some things we'd need to do before we start using this in production:

* Set up authentication (Mesos supports this for the web UI and for frameworks)
* Probably set up on-host firewall rules
* Probably set up an on-prem Docker registry that we can push images to via Jenkins
* Test some more

Some other nice-to-haves:
* Clean up the Zookeeper config a bunch
* Make a zookeeper service file, currently it forks and systemd loses track of it, so you can't restart it properly

Mesos right now (connect to `jaws:5050` or `pandemic:5050`)

Marathon right now: (connect to `jaws:8080` or `pandemic:8080`)

You can connect to `jaws:10001` or `pandemic:10001` if you want to see the service running.


chriskuehl:
I think the best approach here is to keep the physical servers (hal, jaws, pandemic) relatively clean. We still want to run Mesos slaves on them (since that's where the containers actually run), but I think I'll set up a new server on each to host Mesos masters. The master needs minimal resources, so that should be fine.

We'll then eventually probably set up load balancers on the mesos VMs as well. I'm going to start three new servers:
* `whirlwhind`, 169.229.226.50 mesos0 on jaws
* `pileup`, 169.229.226.51 mesos1 on pandemic
* `monsoon`, 169.229.226.52 mesos2 on hal

chriskuehl:
Here's an updated TODO list.

These are the things we should do before merging:

* [ ] Set up authentication (Mesos supports this for the web UI and for frameworks)
* [ ] Evaluate if we want to set up on-host firewall rules (hopefully we can just use authentication instead)
* [x] Probably set up an on-prem Docker registry that we can push images to via Jenkins
* [ ] Clean up the Zookeeper config a bunch
* [ ] Make a zookeeper service file, currently it forks and systemd loses track of it, so you can't restart it properly
* [x] Add a web UI proxy with HTTPS for mesos and marathon
* [ ] Expose the web UI publicly for staff (blocked by adding authentication)

These are things we'll need/want to do afterwards:

* [ ] Figure out how to to do load balancing. marathon-lb is good but not suitable for use out-of-the-box. My current thought is to have the mesos masters (which are VMs on each of the physical servers) act as load balancers. We'd create an "internal" load balancer host (e.g. `intlb.ocf.berkeley.edu`), and certain ports on this host would redirect to specific services (this can just be marathon-lb). We'd also want an "external" load balancer that we expose publicly and that can route traffic to the correct host. But this is harder because it needs to be a L7 proxy (or have tons of IP addresses, one per service). Don't know the best approach yet here, we need to figure out what the actual requirements are.

chriskuehl:
Regarding authentication: Mesos in principle supports pluggable authentication, and it uses SASL which means we could even do GSSAPI with the right auth backend in Mesos. However, sadly the only currently-supported method is CRAM-MD5 which requires a config file of users/passwords.

I wrote a simple Mesos module that uses PAM for authentication, but it's going to be too painful to use in practice as we'd potentially need to maintain it and recompile frequently.

Instead the current plan is to create a user for use by other masters (e.g. `master`) and another user for staff (e.g. `staff`). The staff password will be secret and never used by anyone; instead, staff will access it via nginx which will use PAM auth, and will proxy using the hidden staff password if you succeed to auth against the nginx vhost. This will be a lot more usable and secure than requiring staff to add a password or just sharing a staff password.

chriskuehl:
* [x] Set up authentication for frameworks to the master
* [ ] Set up authentication to talk to Marathon
* [ ] Set up authentication for HTTP **This should be working but isn't. Trying to figure it out still.**
* [x] Set up authentication for slaves
* [x] Set up authentication to access HTTPS web UIs behind nginx
* [ ] Evaluate if we want to set up on-host firewall rules (hopefully we can just use authentication instead)
* [x] Probably set up an on-prem Docker registry that we can push images to via Jenkins
* [ ] Clean up the Zookeeper config a bunch
* [ ] Make a zookeeper service file, currently it forks and systemd loses track of it, so you can't restart it properly
* [x] Add a web UI proxy with HTTPS for mesos and marathon
* [x] Expose the web UI publicly for staff (blocked by adding authentication)

More nice-to-haves:
* [ ] Mesos SSL
* [ ] Marathon SSL

Sadly Zookeeper does not support SSL between masters.

chriskuehl:
This branch is mostly stable and has been open too long, I'm going to merge and file tickets for the remaining issues for better tracking.

PR #54, opened 2016-06-07
Add script to make offsite Box.com backups (rt#3903)

jvperrin:
Quite a bit of a longer script than I would have liked, but I just kept adding small features, like progress bars, and auto-updating output. It uses pycurl and uploads up to 10 files at once (although it could do more, but it doesn't appear to help with speed at all).

Here's a screenshot of it in progress. It's mostly done, as only 5 files remain to upload:
upload-to-box-progress

Errors are reported at the bottom, but the file uploads are retried, all of the errors I have seen in uploads are temporary, but they seem to be able to happen at any point in the upload. Most of the errors are curl errors 55 or 56 ([error code list](https://curl.haxx.se/libcurl/c/libcurl-errors.html)).

I've done a bunch of uploads, they seem to average around 280 Mb/s and take around 3 to 3.5 hours with our current 371.6 GB backup archive.

### DIFF
--- a/modules/ocf_backups/files/create-encrypted-backup
+++ b/modules/ocf_backups/files/create-encrypted-backup
@@ -39,7 +39,7 @@ fi
 cleanup() {
     if [ -e "$SNAPSHOT_PATH" ]; then
         echo "Removing logical volume: $SNAPSHOT_PATH"
-        lvremove -f "$SNAPSHOT_PATH"
+        /sbin/lvremove -f "$SNAPSHOT_PATH"
     fi
 }
 trap cleanup EXIT
@@ -47,7 +47,7 @@ trap cleanup EXIT
 # we don't have *quite* enough space to make a snapshot with the full 2TB of
 # disk space, so we make a large enough snapshot that we'll *probably* avoid
 # trouble
-lvcreate -L 500G --snapshot --name "$SNAPSHOT_PATH" "$TARGET_VOLUME"
+/sbin/lvcreate -L 500G --snapshot --name "$SNAPSHOT_PATH" "$TARGET_VOLUME"
 
 umask 077
 rm -rf "$ARCHIVE_DIRECTORY"
--- a/modules/ocf_backups/files/create-encrypted-backup
+++ b/modules/ocf_backups/files/create-encrypted-backup
@@ -5,6 +5,22 @@
 # to Google Nearline, etc.
 set -o pipefail
 
+# Quiet mode is off by default (show pv output)
+QUIET=false
+
+# If -q is provided, turn on quiet mode
+while getopts "q" opt; do
+    case $opt in
+        q)
+            QUIET=true
+            ;;
+        \?)
+            # Invalid option, just exit
+            exit 1
+            ;;
+    esac
+done
+
 ARCHIVE_DIRECTORY="/opt/backups/scratch/archive"
 TARGET_VOLUME="/dev/vg-backups/backups-live"
 SNAPSHOT_PATH="/dev/vg-backups/backups-snapshot"
@@ -37,16 +53,26 @@ umask 077
 rm -rf "$ARCHIVE_DIRECTORY"
 mkdir "$ARCHIVE_DIRECTORY"
 
-dd if="$SNAPSHOT_PATH" bs=32M | pigz |
-    gpg --homedir "$gpgdir" \
-        --quiet \
-        --encrypt \
-        --trust-model always \
-        --compress-algo none \
-        --recipient ckuehl@berkeley.edu \
-        --recipient willh@ocf.berkeley.edu \
-        --recipient jvperrin@ocf.berkeley.edu \
-        --recipient mattmcal@ocf.berkeley.edu |
-    pv |
+compress_and_encrypt() {
+    dd if="$SNAPSHOT_PATH" bs=32M | pigz |
+        gpg --homedir "$gpgdir" \
+            --quiet \
+            --encrypt \
+            --trust-model always \
+            --compress-algo none \
+            --recipient ckuehl@berkeley.edu \
+            --recipient willh@ocf.berkeley.edu \
+            --recipient jvperrin@ocf.berkeley.edu \
+            --recipient mattmcal@ocf.berkeley.edu
+}
+
+split_parts() {
     split -d -a 4 -b 5G - \
         "${ARCHIVE_DIRECTORY}/backup-$(date +%Y-%m-%d).img.gz.gpg.part"
+}
+
+if [ "$QUIET" = true ]; then
+    compress_and_encrypt | split_parts
+else
+    compress_and_encrypt | pv | split_parts
+fi
--- a/modules/ocf_backups/files/create-encrypted-backup
+++ b/modules/ocf_backups/files/create-encrypted-backup
@@ -2,24 +2,8 @@
 # Create an encrypted backup (in 5 GB pieces) inside our scratch space.
 #
 # This can be put on a hard drive and stored under the SM's mattress, uploaded
-# to Google Nearline, etc.
+# to Google Nearline, automatically uploaded to Box.com, etc.
 set -o pipefail
-
-# Quiet mode is off by default (show pv output)
-QUIET=false
-
-# If -q is provided, turn on quiet mode
-while getopts "q" opt; do
-    case $opt in
-        q)
-            QUIET=true
-            ;;
-        \?)
-            # Invalid option, just exit
-            exit 1
-            ;;
-    esac
-done
 
 ARCHIVE_DIRECTORY="/opt/backups/scratch/archive"
 TARGET_VOLUME="/dev/vg-backups/backups-live"
@@ -39,7 +23,7 @@ fi
 cleanup() {
     if [ -e "$SNAPSHOT_PATH" ]; then
         echo "Removing logical volume: $SNAPSHOT_PATH"
-        /sbin/lvremove -f "$SNAPSHOT_PATH"
+        lvremove -f "$SNAPSHOT_PATH"
     fi
 }
 trap cleanup EXIT
@@ -47,7 +31,7 @@ trap cleanup EXIT
 # we don't have *quite* enough space to make a snapshot with the full 2TB of
 # disk space, so we make a large enough snapshot that we'll *probably* avoid
 # trouble
-/sbin/lvcreate -L 500G --snapshot --name "$SNAPSHOT_PATH" "$TARGET_VOLUME"
+lvcreate -L 500G --snapshot --name "$SNAPSHOT_PATH" "$TARGET_VOLUME"
 
 umask 077
 rm -rf "$ARCHIVE_DIRECTORY"
@@ -71,8 +55,9 @@ split_parts() {
         "${ARCHIVE_DIRECTORY}/backup-$(date +%Y-%m-%d).img.gz.gpg.part"
 }
 
-if [ "$QUIET" = true ]; then
-    compress_and_encrypt | split_parts
-else
+# Only show pv output if the terminal is interactive (no cron email pv spam)
+if [ -t 0 ]; then
     compress_and_encrypt | pv | split_parts
+else
+    compress_and_encrypt | split_parts
 fi
--- a/modules/ocf_backups/files/upload-to-box
+++ b/modules/ocf_backups/files/upload-to-box
@@ -14,8 +14,9 @@ import pycurl
 
 ARCHIVE = '/opt/backups/scratch/archive'
 BOX_DAV = 'https://dav.box.com/dav/'
-MAX_CONCURR_UPLOADS = 10
+BOX_EMAILS = ['ckuehl@berkeley.edu', 'kpeng3.4@berkeley.edu']
 FOLDER = 'ocf-backup-' + time.strftime('%Y-%m-%d')
+MAX_CONCURR_UPLOADS = 10
 
 # To get the OAuth keys for an account:
 #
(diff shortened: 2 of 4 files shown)


chriskuehl:
looks pretty solid to me! nice work!

chriskuehl:
we should get some trusted staffer (you?) who doesn't really care about their box account to put their pass in a config file so that we can cronjob this

jvperrin:
Sure, that would work fine, I don't really have anything much on there besides some encrypted laptop backups. Right now the script asks for an email and password, but it would be really easy to fetch from some puppet config file instead.

It would definitely be good to have more regular offsite backups though, so a cronjob would definitely be nice.

chriskuehl:
if you're using yours, i'd just pick someone else. probably only one or two staff are using theirs at all

we don't even need to trust that person since our backups are encrypted, just need to know we'll be able to access them if we need them

jvperrin:
I don't really mind either way, but it would be pretty easy to switch accounts if needed. We could actually upload the backups to one account and then share those uploaded files with all other accounts, and then anyone should be able to access them, not just the person who originally uploaded the files.

We could also use the [shared link API](https://docs.box.com/reference#create-a-shared-link-for-a-folder) to generate a link that could be used to download the backups (even without an account), which then could be emailed to root or something like that. I think the API would need to use OAuth though, but there is a [Python SDK](https://github.com/box/box-python-sdk) if that would make things easier. I think it would have to be authenticated manually the first time to get the `refresh_token`, but after that as long as we make a backup at least once every 60 days the same token should stay alive, [as described on their forums](https://community.box.com/t5/Developer-Forum/Re-How-I-can-connect-to-OAuth2-from-server-side-script-without/m-p/1980#U1980).

jvperrin:
We would need to have somewhere to store the refresh token between backup runs though, maybe a private file somewhere. The token would change after every run, so it would need to be stored in a place that could be updated regularly...

jvperrin:
Ok, I think this is ready to be merged now! It seems to be working as intended and will automatically make backups weekly. (or more frequently, but it takes about 20 hours to make and upload a full backup, so probably weekly is best)

modules/ocf_backups/files/create-encrypted-backup
```diff
@@ -5,6 +5,22 @@
...
+    case $opt in
+        q)
+            QUIET=true
+            ;;
+        \?)
+            # Invalid option, just exit
+            exit 1
+            ;;
+    esac
+done
```
chriskuehl: meh. parsing command line options in bash is the worst.

can you just check if the shell is interactive? or just make QUIET an environment variable

modules/ocf_backups/files/create-encrypted-backup
```diff
@@ -23,30 +39,40 @@ fi
 cleanup() {
     if [ -e "$SNAPSHOT_PATH" ]; then
         echo "Removing logical volume: $SNAPSHOT_PATH"
-        lvremove -f "$SNAPSHOT_PATH"
+        /sbin/lvremove -f "$SNAPSHOT_PATH"
```
chriskuehl: maybe we should change the cron PATH globally via puppet? seems same to always put sbin on the path (and avoids trouble when debian moves things, e.g. we had trouble when kadmin moved)

modules/ocf_backups/files/create-encrypted-backup
```diff
@@ -23,30 +39,40 @@ fi
...
+split_parts() {
     split -d -a 4 -b 5G - \
         "${ARCHIVE_DIRECTORY}/backup-$(date +%Y-%m-%d).img.gz.gpg.part"
+}
+
+if [ "$QUIET" = true ]; then
+    compress_and_encrypt | split_parts
+else
+    compress_and_encrypt | pv | split_parts
+fi
```
chriskuehl: nice

modules/ocf_backups/files/upload-to-box
```diff
@@ -0,0 +1,444 @@
...
+        conn.file_bytes = os.path.getsize(file_path)
+        conn.fp = open(file_path, 'rb')
+
+        conn.setopt(conn.INFILESIZE, conn.file_bytes)
+        conn.setopt(conn.READFUNCTION, conn.fp.read)
+        conn.setopt(conn.UPLOAD, True)
+        conn.setopt(conn.URL, '{}{}/{}'.format(BOX_DAV, FOLDER, self.filename))
+        conn.setopt(conn.USERPWD, self.auth)
+
+    def progress(self, download_t, download_d, upload_t, upload_d):
```
chriskuehl: maybe call `store_progress`? current name implies it might return the current progress

modules/ocf_backups/files/upload-to-box
```diff
@@ -0,0 +1,444 @@
...
+
+    if method:
+        conn.setopt(conn.CUSTOMREQUEST, method)
+
+    conn.perform()
+    http_status = conn.getinfo(pycurl.HTTP_CODE)
+    conn.close()
+
+    # Check if HTTP status is in 2XX (success of some sort)
+    if http_status in range(200, 299):
```
chriskuehl: does this look better as `if 200 <= http_status < 300:`?

that's probably how i'd write it, this seems fine though

modules/ocf_backups/files/upload-to-box
```diff
@@ -0,0 +1,444 @@
...
+            creds['api_client_id'],
+            creds['api_client_secret']
+        )
+        response = box_api_call('https://app.box.com/api/oauth2/token',
+                                data='grant_type=refresh_token' + post_data)
+
+        # Write the new refresh token to the file
+        f.seek(0)
+        f.write(response['refresh_token'])
+        f.truncate()
```
chriskuehl: nice. everyone forgets to truncate :)

modules/ocf_backups/files/upload-to-box
```diff
@@ -0,0 +1,444 @@
...
+    for entry in response['entries']:
+        if entry['name'] == folder_name and entry['type'] == 'folder':
+            return entry['id']
+
+
+def add_collaborators(access_token, folder_id, emails):
+    """Set a list of people who can access the folder with the backups"""
+    for email in emails:
+        data = '{"item": {"id": "' + folder_id + '", "type": "folder"}, "accessible_by":\
+            {"login": "' + email + '", "type": "user"}, "role": "viewer"}'
```
chriskuehl: this seems better than manually constructing json?
```python
data = json.dumps({
    'item': {'id': folder_id, 'type': 'folder'},
    'accessible_by': {'login': email, 'type': 'user'},
    'role': 'viewer',
})
```

modules/ocf_backups/files/upload-to-box
```diff
@@ -0,0 +1,444 @@
...
+        print('Error creating folder, HTTP ' + str(http_status))
+        sys.exit(3)
+
+
+def print_progress(multi, stats, prev_total):
+    """Show an updating progress bar for all ongoing uploads"""
+    width = get_terminal_size((80, 20)).columns  # Default size as a fallback
+
+    # Move cursor to the top left using ANSII magic to overwrite previous lines
+    sys.stdout.write('\033[1;1H')
```
chriskuehl:
```
 (ノﾟοﾟ)ノﾐ★゜・。。・゜゜・。。・゜☆゜・。。・゜゜・。。・゜゜・。。・゜☆゜・。。・゜゜・。。・゜
```
such magic

modules/ocf_backups/files/upload-to-box
```diff
@@ -0,0 +1,444 @@
...
+    conn.file_bytes = 0
+    multi.free.append(conn)
+
+
+def friendly_file_size(num_bytes):
+    """To convert bytes to human-readable units
+
+    Adapted from http://stackoverflow.com/a/1094933/1979001
+    """
+    for unit in ['', 'KB', 'MB', 'GB', 'TB', 'PB']:
```
chriskuehl: i sure hope we never have petabytes of backups :P

modules/ocf_backups/files/upload-to-box
```diff
@@ -0,0 +1,444 @@
...
+    except PermissionError:
+        print('You must run this script as root!')
+        sys.exit(1)
+
+    # Make sure there are actually files to upload
+    assert files, 'No files found to upload, check your path'
+
+    # Get authentication for Box.com from credentials file
+    with open('/opt/share/backups/box-creds.json') as f:
+        creds = json.loads(f.read())
```
chriskuehl: `json.load(f)`

modules/ocf_backups/manifests/offsite.pp
```diff
@@ -19,5 +28,25 @@
...
+
+    # Box.com credentials and API id/secret
+    '/opt/share/backups/box-creds.json':
+      source => 'puppet:///private/box-creds.json',
+      mode   => '0600';
+
+    # This changes each time a backup is made, so it is not stored on the
+    # puppetmaster and instead is just a local file on the backups server.
+    '/opt/share/backups/box-refresh-token':
+      mode => '0600';
```
chriskuehl: what happens here on the first time? puppet error?

modules/ocf_backups/manifests/offsite.pp
```diff
@@ -19,5 +28,25 @@
...
+  }
+
+  # Runs Saturday at noon, makes a backup and then uploads it to Box.com
+  cron {
+    'encrypt-and-backup':
+      command => '/opt/share/backups/create-encrypted-backup -q && /opt/share/backups/upload-to-box -q',
+      user    => 'root',
```
chriskuehl: we usually use barewords for usernames

chriskuehl:
sweet! i won't pretend i understand this crazy curl magic, but looks pretty good to me. automated offsite backups are awesome.

shipit!

modules/ocf_backups/files/create-encrypted-backup
jvperrin: I don't think it's too bad really, `getopts` makes it pretty easy, but checking if the shell is interactive isn't hard, so I'll just do that instead.

modules/ocf_backups/files/create-encrypted-backup
jvperrin: Yeah, seems like a good idea, especially since the default path used is `/usr/bin:/bin`, but files like `/etc/crontab` set the path to `/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin`. It looks like there is [some discussion](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=617898) of changing it but that it probably won't happen soon.

I don't see there being much trouble if this moved though, this script would just fail for one week (at the `lvcreate` stage) and then be fixed soon after that with the new path. It's still less portable though, so I agree that setting the `PATH` globally would be a good idea. Plus it avoids the need for absolute paths for everything not in `/usr/bin`or `/bin`, which is good.

modules/ocf_backups/files/create-encrypted-backup
jvperrin: I was originally trying to just add the `-q` flag to `pv` if the variable was set, but that didn't work so well, so I went with this instead. I should really use bash functions more in scripts, they are pretty nice.

modules/ocf_backups/files/upload-to-box
jvperrin: Good idea :+1:

modules/ocf_backups/files/upload-to-box
jvperrin: That's probably more readable than using a range, so I'll change it to that.

modules/ocf_backups/files/upload-to-box
jvperrin: Pretty sure I just took it from [here](http://stackoverflow.com/questions/2424000/read-and-overwrite-a-file-in-python), but thanks :P

modules/ocf_backups/files/upload-to-box
jvperrin: Ooh, yes it does! I had no idea that was possible, cool!

modules/ocf_backups/files/upload-to-box
jvperrin: And of course I misspelled ANSI lol

modules/ocf_backups/files/upload-to-box
jvperrin: I really hope not as well, but it's there just in case someone tries to back up petabytes of something I suppose. Good luck to them, it's going to take like a year with our transfer rate. (~10 hours for 1.2 TB currently, so 1024 TB ~= 355 days)

Yeah, I really hope we never have petabytes of backups lol

modules/ocf_backups/files/upload-to-box
jvperrin: Nice, that's much cleaner.

modules/ocf_backups/manifests/offsite.pp
jvperrin: Yeah, pretty much. The issue is that it can't be stored in the puppetmaster's private store from what I can tell because it has to change each time a backup is made, but the token should be kept secure, since if someone has it then they could delete all the backups. I feel like there's a much better way to do this, but I don't know what it would be. Maybe storing the token in the mysql database or on NFS somewhere would be better?

modules/ocf_backups/manifests/offsite.pp
jvperrin: Gah, I keep forgetting that, thanks!

modules/ocf_backups/manifests/offsite.pp
chriskuehl: I think it's fine as long as the job will renew it, but I don't think Puppet shouldn't error on the first run? Does that mean a human will have to manually create the file?

jvperrin: I would like it if Puppet didn't error on the first run too, but that would also mean backups wouldn't work once they run on that week (or the uploaded folder wouldn't be shared correctly at least, the encryption and upload would still work). Unless a human had copied the refresh token from the old server (currently hal) and added it on the new server, then it doesn't have access to the Box.com API.

The issue is that as part of OAuth for Box, whenever a new access token is claimed using the old refresh token, a new refresh token is sent back too, and this can happen indefinitely as long as each refresh token is used within 60 days of being issued. However, that means that the new refresh token has to be stored somewhere until the next run. Storing it in the puppetmaster's private store would be best, but I don't see a way to do that without having to ssh in or something.

chriskuehl: Can you separate token refreshing into a separate script and run it in a cronjob on the puppetmaster?

Alternatively:
* MySQL would works but is a little heavy for storing one value
* NFS would work but we don't mount NFS on hal currently, and it's really bad for secrets (it's not encrypted at all)

jvperrin: Yeah, I like the idea of doing a separate cronjob on the puppetmaster the most out of those. I agree that MySQL would be a bit heavy, but might be the easiest to update. The biggest issue I see with the separate cronjob idea is that it would have to run every hour (or slightly more often since every hour would create a small gap between valid tokens) to have a valid access token at all times just for the one time it needs it per week. That's probably fine, if a little wasteful, but better than having puppet runs fail when they shouldn't.

jvperrin:
Well, the pycurl part of it is mostly taken from [here](https://github.com/pycurl/pycurl/blob/master/examples/retriever-multi.py). It's not very pythonic, since it originates from curl, but it does the trick and seems pretty fast. Definitely some magic going on there though, and I wouldn't say I understand that part of it either :stuck_out_tongue: 

Alright, I've replaced the local file on hal with a MySQL database. It's a bit silly as it is, because the database only has one table that holds one row, but it works, and is better than having a local file that could easily be forgotten about when migrating backups.

chriskuehl:
nice, lgtm

jvperrin:
Thanks for reviewing it so thoroughly! First real backup should happen in a couple days.

PR #55, opened 2016-07-09
Puppet irc

jvperrin:
Pretty much done now, it just needs testing! There's a test IRC server at `dev-irc.ocf.berkeley.edu` for testing, so I'd like to see if any features are missing or things are broken over there before merging this. I've already imported data from the old IRCd into mysql, so all NickServ credentials should work the same as before. Opers for the IRCd use LDAP for authentication, so they will use regular OCF credentials, but [only specified users](https://github.com/ocf/puppet/pull/55/files#diff-9cf5ceb409581fa83aab5a486c77e8e5R98) can authenticate to become opers, and IRCd opers aren't especially useful for much besides being able to access more anope commands, as far as I can tell. Maybe they would be more useful if we weren't using services for most things. Auto-op in channels should work fine as usual after authenticating with NickServ, since it is stored in mysql, so hopefully everything already works there.

### DIFF
--- a/modules/ocf_irc/files/anope/chanserv.conf
+++ b/modules/ocf_irc/files/anope/chanserv.conf
@@ -3,337 +3,337 @@
 # https://wiki.anope.org/index.php/2.0/Configuration#ChanServ
 
 service {
-	nick = "ChanServ"
-	user = "services"
-	host = "services.irc.ocf.berkeley.edu"
-	gecos = "Channel Registration Service"
-	channels = "#services"
+  nick = "ChanServ"
+  user = "services"
+  host = "services.irc.ocf.berkeley.edu"
+  gecos = "Channel Registration Service"
+  channels = "#services"
 }
 
 module {
-	name = "chanserv"
-	client = "ChanServ"
+  name = "chanserv"
+  client = "ChanServ"
 
-	defaults = "keeptopic peace cs_secure securefounder signkick"
-	expire = 30d
+  defaults = "keeptopic peace cs_secure securefounder signkick"
+  expire = 120d
 
-	accessmax = 1024
-	inhabit = 15s
+  accessmax = 1024
+  inhabit = 15s
 
-	reasonmax = 200
+  reasonmax = 200
 
-        # Signed kick messages:
-        # %n is the nick of the kicker
-        # %m is the message specified
-	signkickformat = "%m (%n)"
+  # Signed kick messages:
+  # %n is the nick of the kicker
+  # %m is the message specified
+  signkickformat = "%m (%n)"
 
-	disallow_hostmask_access = false
-	disallow_channel_access = false
+  disallow_hostmask_access = false
+  disallow_channel_access = false
 
-	always_lower_ts = false
+  always_lower_ts = false
 }
 
 # Privileges
 privilege {
-	name = "ACCESS_CHANGE"
-	rank = 0
-	level = 10
-	flag = "f"
-	xop = "SOP"
+  name = "ACCESS_CHANGE"
+  rank = 0
+  level = 10
+  flag = "f"
+  xop = "SOP"
 }
 
 privilege {
-	name = "ACCESS_LIST"
-	rank = 10
-	level = 3
-	flag = "f"
-	xop = "VOP"
+  name = "ACCESS_LIST"
+  rank = 10
+  level = 3
+  flag = "f"
+  xop = "VOP"
 }
 
 privilege {
-	name = "AKICK"
-	rank = 250
-	level = 10
-	flag = "K"
-	xop = "SOP"
+  name = "AKICK"
+  rank = 250
+  level = 10
+  flag = "K"
+  xop = "SOP"
 }
 
 privilege {
-	name = "ASSIGN"
-	rank = 270
-	level = "founder"
-	flag = "s"
-	xop = "QOP"
+  name = "ASSIGN"
+  rank = 270
+  level = "founder"
+  flag = "s"
+  xop = "QOP"
 }
 
 privilege {
-	name = "AUTOHALFOP"
-	rank = 100
-	level = 4
-	flag = "H"
-	xop = "HOP"
+  name = "AUTOHALFOP"
+  rank = 100
+  level = 4
+  flag = "H"
+  xop = "HOP"
 }
 
 privilege {
-	name = "AUTOOP"
-	rank = 210
-	level = 5
-	flag = "O"
-	xop = "AOP"
+  name = "AUTOOP"
+  rank = 210
+  level = 5
+  flag = "O"
+  xop = "AOP"
 }
 
 privilege {
-	name = "AUTOOWNER"
-	rank = 330
-	level = 9999
-	flag = "Q"
-	xop = "QOP"
+  name = "AUTOOWNER"
+  rank = 330
+  level = 9999
+  flag = "Q"
+  xop = "QOP"
 }
 
 privilege {
-	name = "AUTOPROTECT"
-	rank = 240
-	level = 10
-	flag = "A"
-	xop = "SOP"
+  name = "AUTOPROTECT"
+  rank = 240
+  level = 10
+  flag = "A"
+  xop = "SOP"
 }
 
 privilege {
-	name = "AUTOVOICE"
-	rank = 50
-	level = 3
-	flag = "V"
-	xop = "VOP"
+  name = "AUTOVOICE"
+  rank = 50
+  level = 3
+  flag = "V"
+  xop = "VOP"
 }
 
 privilege {
-	name = "BADWORDS"
-	rank = 260
-	level = 10
-	flag = "K"
-	xop = "SOP"
+  name = "BADWORDS"
+  rank = 260
+  level = 10
+  flag = "K"
+  xop = "SOP"
 }
 
 privilege {
-	name = "BAN"
-	rank = 150
-	level = 4
-	flag = "b"
-	xop = "HOP"
+  name = "BAN"
+  rank = 150
+  level = 4
+  flag = "b"
+  xop = "HOP"
 }
 
 privilege {
-	name = "FANTASIA"
-	rank = 30
-	level = 3
-	flag = "c"
-	xop = "VOP"
+  name = "FANTASIA"
+  rank = 30
+  level = 3
+  flag = "c"
+  xop = "VOP"
 }
 
 privilege {
-	name = "FOUNDER"
-	rank = 360
-	level = 10000
-	flag = "F"
-	xop = "QOP"
+  name = "FOUNDER"
+  rank = 360
+  level = 10000
+  flag = "F"
+  xop = "QOP"
 }
 
 privilege {
-	name = "GETKEY"
-	rank = 180
-	level = 5
-	flag = "G"
-	xop = "AOP"
+  name = "GETKEY"
+  rank = 180
+  level = 5
+  flag = "G"
+  xop = "AOP"
 }
 
 privilege {
-	name = "HALFOP"
-	rank = 120
-	level = 5
-	flag = "h"
-	xop = "AOP"
+  name = "HALFOP"
+  rank = 120
+  level = 5
+  flag = "h"
+  xop = "AOP"
 }
 
 privilege {
-	name = "HALFOPME"
-	rank = 110
-	level = 4
-	flag = "h"
-	xop = "HOP"
+  name = "HALFOPME"
+  rank = 110
+  level = 4
+  flag = "h"
+  xop = "HOP"
 }
 
 privilege {
-	name = "INFO"
-	rank = 80
-	level = 9999
-	flag = "I"
-	xop = "QOP"
+  name = "INFO"
+  rank = 80
+  level = 9999
+  flag = "I"
+  xop = "QOP"
 }
 
 privilege {
-	name = "INVITE"
-	rank = 190
-	level = 5
-	flag = "i"
-	xop = "AOP"
+  name = "INVITE"
+  rank = 190
+  level = 5
+  flag = "i"
+  xop = "AOP"
 }
 
 privilege {
-	name = "KICK"
-	rank = 130
-	level = 4
-	flag = "k"
-	xop = "HOP"
+  name = "KICK"
+  rank = 130
+  level = 4
+  flag = "k"
+  xop = "HOP"
 }
 
 privilege {
-	name = "MEMO"
-	rank = 280
-	level = 10
-	flag = "m"
-	xop = "SOP"
+  name = "MEMO"
+  rank = 280
+  level = 10
+  flag = "m"
+  xop = "SOP"
 }
 
 privilege {
-	name = "MODE"
-	rank = 170
-	level = 9999
-	flag = "s"
-	xop = "QOP"
+  name = "MODE"
+  rank = 170
+  level = 9999
+  flag = "s"
+  xop = "QOP"
 }
 
 privilege {
-	name = "NOKICK"
-	rank = 20
-	level = 1
-	flag = "N"
-	xop = "VOP"
+  name = "NOKICK"
+  rank = 20
+  level = 1
+  flag = "N"
+  xop = "VOP"
 }
 
 privilege {
-	name = "OP"
-	rank = 230
-	level = 5
-	flag = "o"
-	xop = "SOP"
+  name = "OP"
+  rank = 230
+  level = 5
+  flag = "o"
+  xop = "SOP"
 }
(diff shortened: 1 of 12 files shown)


jvperrin:
Also I'll be adding documentation in ocfweb after this is all merged, since there are a few things to write on how to add people to the auto-op list, and manage things with the IRC server.

modules/ocf_irc/files/anope/chanserv.conf
```diff
@@ -0,0 +1,484 @@
+# Comments trimmed for easier editing.
```
chriskuehl: Is there any way we can avoid committing this entire massive file? I feel like we only made a couple minor edits. Committing entire files is pretty fragile and has been cause for much pain in the past.

It looks like anope supports includes; maybe you can append an include line to the file, and put our small number of changes in the included file?

Alternatively, if it's just a couple lines changed, `sed` isn't a terrible option. Normally augeas would be preferred, but this looks like a complicated custom config format, so probably not an option here.

modules/ocf_irc/files/anope/global.conf
```diff
@@ -0,0 +1,24 @@
+# Most comments have been stripped for easier editing
+# The original comments can be found on anope's wiki:
+# https://wiki.anope.org/index.php/2.0/Configuration#Global
+
+service {
+	nick = "Global"
+	user = "services"
+	host = "services.irc.ocf.berkeley.edu"
```
chriskuehl: Is this the only line we changed? Can we sed that?

modules/ocf_irc/manifests/init.pp
```diff
@@ -1,4 +1,11 @@
 class ocf_irc {
   include ocf_ssl
+  include ocf_irc::ircd
+  include ocf_irc::services
   include ocf_irc::slack
+
+  # Make the irc user able to read the certs for running the IRCd with SSL
+  user { 'irc':
+    groups => 'ssl-cert',
+  }
```
chriskuehl: You might wish to `require => Package['whatever-package-adds-this-user']` here so that the groups are added after the user is created, rather than creating a user here.

Otherwise, on the initial Puppet run, this user might get created with some default home directory, groups, shell, etc., and the package install won't re-create the user or set the settings properly.

modules/ocf_irc/manifests/ircd.pp
```diff
@@ -0,0 +1,32 @@
...
+    notify  => Service['inspircd'],
+    owner   => 'irc',
+    group   => 'irc',
+  }
+
+  file {
+    '/etc/default/inspircd':
+      content => 'INSPIRCD_ENABLED=1',
+      owner   => 'root',
+      group   => 'root';
```
chriskuehl: nit: we normally use barewords for owner/group. nbd

modules/ocf_irc/manifests/ircd.pp
```diff
@@ -0,0 +1,32 @@
+class ocf_irc::ircd {
+  package { 'inspircd':; }
+
+  service { 'inspircd':
+    ensure  => 'running',
+    restart => 'service inspircd reload',
+    require => Package['inspircd'],
+  }
```
chriskuehl: I feel that `ensure => running` is a smell in Puppet code. Is there a reason Puppet needs to start this service? I'd think systemd should be able to handle it.

Also, why changing `restart`?

modules/ocf_irc/manifests/services.pp
```diff
@@ -0,0 +1,36 @@
...
+  File {
+    require => Package['anope'],
+    notify  => Service['anope'],
+    owner   => 'irc',
+    group   => 'irc',
+  }
+
+  file {
+    '/etc/default/anope':
+      content => 'START=yes',
```
chriskuehl: `content => "START=yes\n"`

modules/ocf_irc/manifests/services.pp
```diff
@@ -0,0 +1,36 @@
...
+      owner   => 'root',
+      group   => 'root';
+
+    '/etc/anope/services.conf':
+      content => template('ocf_irc/services.conf.erb'),
+      mode    => '0640';
+
+    '/etc/anope/services.motd':
+      content => 'Welcome to OCF IRC Services!';
```
chriskuehl: newline at end

modules/ocf_irc/files/anope/chanserv.conf
jvperrin: I definitely agree with your feeling about this, since these are really large config files. I tried to use augeas before, but discovered there is no lens for this format, which really sucks. Sed would work fine, but it's at that awkward point where enough has been changed that a lot of sed commands would need to be written to get the same effect, but that might still be a better option than committing the whole thing.

Includes would be best, not sure if that would work, I hadn't considered it. If later options overwrite earlier config options (or vice-versa), then that would work fine, but I'm not sure how anope behaves on this.

modules/ocf_irc/files/anope/global.conf
jvperrin: No, there's a few more changed, such as the global messages on power cycles, and things like that: https://wiki.anope.org/index.php/2.0/Configuration#Global

modules/ocf_irc/templates/inspircd.conf.erb
```diff
@@ -0,0 +1,202 @@
...
+          server="ldaps://ldap.ocf.berkeley.edu"
+          searchscope="subtree"
+          attribute="uid">
+
+# List of allowed opers
+<oper name="waf"          password="uid=waf"      host="*@*" type="NetAdmin" hash="ldap">
+<oper name="nattofriends" password="uid=tzhu"     host="*@*" type="NetAdmin" hash="ldap">
+<oper name="ckuehl"       password="uid=ckuehl"   host="*@*" type="NetAdmin" hash="ldap">
+<oper name="jvperrin"     password="uid=jvperrin" host="*@*" type="NetAdmin" hash="ldap">
+<oper name="mattmcal"     password="uid=mattmcal" host="*@*" type="NetAdmin" hash="ldap">
```
chriskuehl: I don't really feel we should be using (potentially root) admin passwords for IRC.

modules/ocf_irc/manifests/init.pp
jvperrin: Thanks, forgot about that, that's a great idea. I think both the `inspircd` and `anope` packages add an irc user, but I'll just pick one since that should be enough to create the user properly.

modules/ocf_irc/manifests/ircd.pp
chriskuehl: (for context, I've seen this get out-of-control in some Puppet codebases where Puppet is practically used as a process supervisor; a server isn't considered "ready" after a reboot until Puppet has had a chance to run)

jvperrin: No, it's fine without it, I just thought it would be best if it's running. I'll remove that then.

I've changed restart so that it reloads gracefully when any config files are changed instead of performing a full restart and kicking everyone off the server.

modules/ocf_irc/templates/services.conf.erb
```diff
@@ -0,0 +1,243 @@
...
+
+oper {
+    name = "jvperrin"
+    type = "Services Root"
+}
+
+oper {
+    name = "mattmcal"
+    type = "Services Root"
+}
```
chriskuehl: If you're really bored, you could probably DRY this out by having an array of operators in the Puppet code, and generate these blocks (and the list in `inspircd.conf.erb` above) from that.

```puppet
$operators = [
  {nick => ckuehl, username => ckuehl},
  {nick => nattofriends, username => tzhu},
  # ...etc
]
```

I hear you're good at Ruby ;)

modules/ocf_irc/manifests/services.pp
jvperrin: Haha, thanks for catching that, I did that for the IRCd default file but forgot it here!

modules/ocf_irc/manifests/services.pp
jvperrin: And seem to have forgotten it here too lol

chriskuehl:
Looks pretty reasonable to me, thanks for tackling this! This has definitely reinforced my idea that everything related to IRC is over-engineered :P

Having this puppeted finally is a big improvement.

modules/ocf_irc/templates/inspircd.conf.erb
jvperrin: Not sure what alternative there is besides including encrypted passwords in the config file (what the current configuration does). Maybe this can [use mysql as a backend](https://wiki.inspircd.org/Modules/2.0/mysql) here too? Not sure how well that would work, but it would be preferable to using LDAP if it does work.

I would trust inspircd with my password, but this should at least require using a SSL connection to enter a oper password... We do need this config section here to properly use anope though, since some anope commands require a user to be both an IRCd oper and a services root with anope, which is a pain and badly designed in my opinion. </rant>

modules/ocf_irc/templates/services.conf.erb
jvperrin: Haha, that would be better than repeating it in both the IRCd and anope config files. I'll have a look and see how complicated it gets...

jvperrin:
It's definitely overengineered when your IRC services package ships with 6000+ lines of config files. :frowning: At least most of those lines are comment blocks...

modules/ocf_irc/manifests/ircd.pp
```diff
@@ -0,0 +1,31 @@
...
+  File {
+    require => Package['inspircd'],
+    notify  => Service['inspircd'],
+    owner   => irc,
+    group   => irc,
+  }
+
+  file {
+    '/etc/default/inspircd':
+      content => 'INSPIRCD_ENABLED=1',
```
chriskuehl: newline here too

modules/ocf_irc/manifests/services.pp
```diff
@@ -0,0 +1,36 @@
+class ocf_irc::services {
+  package { 'anope':; }
+
+  service { 'anope':
+    ensure  => 'running',
```
chriskuehl: remove?

jvperrin: Thanks, I missed these ones

modules/ocf_irc/files/anope/chanserv.conf
jvperrin: Well, I tried overridding some config later in the file that had been defined before and anope segfaulted, so I think sed would be the best option apart from including full config files like this.

modules/ocf_irc/templates/inspircd.conf.erb
jvperrin: Ok, I changed this so that it stores oper credentials as sha256 hashed passwords in mysql. Stronger encryption would be better, like [using bcrypt or something](https://github.com/inspircd/inspircd/issues/741), but the strongest option available is sha256. Unfortunately, any opers will have to go in to mysql and add their own hashed password, but at least it's not the same password as for admin/root access, which is definitely better.

modules/ocf_irc/templates/services.conf.erb
jvperrin: Looks like I was bored! I only had one section to replace since we aren't using LDAP for IRCd opers, so it was much simpler and I could just use a regular array instead of an array of hashes.

jvperrin:
Ok, I think I've fixed all the things I can think of here, apart from committing the whole large files, but I think they are large enough that sed wouldn't work very well and would be more complicated than it should be. Includes would be the best option really, but any duplicated options appeared to make anope segfault (of course).

Any other ideas on what should be changed?

chriskuehl:
lgtm!

one final question: dwc mentioned on irc that it might be possible to omit the parts of the config files that we didn't change and maybe it would use the defaults. did you try that?

jvperrin:
Some values seem to have defaults with anope, but others do not. I haven't tried that with most of the config values, but I could try that to see if I can slim down the config files. Unfortunately I think omitting any of [the commands](https://github.com/ocf/puppet/pull/55/files#diff-3222067e9d2f550120d183e37af9bc63R339) will just make them unavailable, so they will have to stay, but some of the config options specifying expiration times or things like that could maybe be removed. As for inspircd, all their config values all have defaults I believe, so they can all be removed if sensible defaults exist.

chriskuehl:
okay, seems reasonable to me. maybe we can make this simpler in the future by moving it to mesos (looks like it has no state besides the database)

nattofriends:
Why would moving it to mesos make it simpler???

2016-07-15 9:39 GMT-07:00 Chris Kuehl <notifications@github.com>:

chriskuehl:
@nattofriends compare our RT config in Puppet with the Dockerfile:
https://github.com/ocf/rt/pull/1/files

it's obviously not going to solve the config files though

modules/ocf_irc/manifests/services.pp
```diff
@@ -0,0 +1,37 @@
+class ocf_irc::services {
+  package { 'anope':; }
+
+  service { 'anope':
+    require => [Package['anope'], Service['inspircd']],
+  }
+
+  $passwords = parsejson(file("/opt/puppet/shares/private/${::hostname}/services-passwords"))
+
+  $root_nicks = ['waf', 'nattofriends', 'ckuehl', 'jvperrin', 'mattmcal']
```
gnowxilef: please to add `faw`, `not-waf`, and `not-faw`?

chriskuehl: please pick one

jvperrin: Alright, I just used `waf` because that was the config from before (current anope server), but adding more is fine.

modules/ocf_irc/files/anope/chanserv.conf
```diff
@@ -0,0 +1,172 @@
...
+privilege { name = "OWNER";         rank = 350; level = "founder"; flag = "q"; xop = "QOP"; }
+privilege { name = "OWNERME";       rank = 340; level = 9999;      flag = "q"; xop = "QOP"; }
+privilege { name = "PROTECT";       rank = 310; level = 9999;      flag = "a"; xop = "QOP"; }
+privilege { name = "PROTECTME";     rank = 300; level = 10;        flag = "a"; xop = "AOP"; }
+privilege { name = "SET";           rank = 320; level = 9999;      flag = "s"; xop = "QOP"; }
+privilege { name = "SIGNKICK";      rank = 140; level = 9999;      flag = "K"; xop = "QOP"; }
+privilege { name = "TOPIC";         rank = 160; level = 5;         flag = "t"; xop = "AOP"; }
+privilege { name = "UNBAN";         rank = 200; level = 4;         flag = "u"; xop = "HOP"; }
+privilege { name = "VOICE";         rank = 70;  level = 4;         flag = "v"; xop = "HOP"; }
+privilege { name = "VOICEME";       rank = 60;  level = 3;         flag = "v"; xop = "VOP"; }
```
jvperrin: This file is a lot shorter now, mostly just because I compacted these lines though. Still, I think it makes it easier to read and edit easily.

jvperrin:
Ok, I cut down on a bunch of the config file lines, so I think this is ready to merge now.

I think the plan should be to merge this, then shut down the old IRC server and services and move them to different (non-standard) ports on flood so that we can still go back to that server and get config options off it easily if need be. Then, the new IRC server will start, using this config and the mysql data, and everyone should automatically join that one since it's running on the same ports the old IRC server ran on. Eventually we can get rid of the old IRC server entirely, or at least shut it down, but I think it would be good to keep around until we have thoroughly tested the new IRC server works well as a replacement.

chriskuehl:
@jvperrin I would suggest to kill flood entirely and bring up a new server. It has tons of cruft lying around on the filesystem.

The way I usually do this is:

```
$ virsh shutdown flood
$ virsh dumpxml flood > old-flood.xml
$ virsh undefine flood
$ lvrename /dev/vg/flood flood-old
$ makevm [...]
```

If something goes wrong, you can easily revert back to the working one with `virsh define old-flood.xml`, adjust the path to the LV via `virsh edit`, and then start it.

When replacing a server, make sure to first connect to lightning and run `puppet cert clean flood.ocf.berkeley.edu` before the new server comes up and tries to run puppet (this will happen at the end of makevm automatically). Otherwise you'll get certificate errors and have to run manual commands to fix on both the puppetmaster and the server itself.

You might like to put a copy of flood in /opt/backups (we usually do this when replacing special snowflake servers that people might have files lying around on), you can just `dd` the disk image of `flood-old` to hal before removing it.

Good luck and thanks for doing this! I've wanted this to happen for years now :P

PR #56, opened 2016-07-17
Add renew to lets-encrypt-update

chriskuehl:


chriskuehl:
Eventually we probably want to relax the prints when acquiring/renewing, but for now I think it's useful when this script is still not super well tested (we only get a couple emails a week anyway).

jvperrin:
Looks good to me! I thought it might be good to add in some error handling for when acme-tiny returns a 500 error from Let's Encrypt, but I don't think that would be so easy given acme-tiny is an external script. An occasional email about it erroring is probably fine, and if it happens a lot it's probably our fault anyway, so it would be good to still have it email us for 500 errors.

chriskuehl:
If it errors a lot, it's probably LE's fault. But yeah, I don't think we can handle errors better than just raising. And if it errors consistently for long enough, we'll lose all our certs and have to start over. So we need to be careful about that.

PR #57, opened 2016-07-20
Add web IRC

jvperrin:
This adds a web IRC page at https://irc.ocf.berkeley.edu that anyone can use to easily connect to our IRC server. It should be useful because there's practically no configuration to do, except entering a nickname and pressing the connect button, so it's easy for anyone to get in contact with us without having to download an IRC client or have to use the terminal. It might be useful for staff as well, since it's quick to connect to, but I don't imagine that's quite as useful, since staff could just use pyrc.


nattofriends:
How does this actually install the node package? :confused:

modules/ocf_irc/manifests/nodejs/apt.pp
```diff
@@ -0,0 +1,18 @@
+class ocf_irc::nodejs::apt {
+  package {
+    'apt-transport-https':;
+  }
+
+  # Install a newer version of Node.js for the Slack bridge and web IRC app
+  apt::key { 'nodejs':
+    id     => '9FD3B784BC1C6FC31A8A0A1C1655A0AB68576280',
+    source => 'https://deb.nodesource.com/gpgkey/nodesource.gpg.key';
+  }
```
chriskuehl: do we need a newer version? i don't really trust the npm folks

chriskuehl: can we backport from stretch instead?

modules/ocf_irc/manifests/nodejs/webirc.pp
```diff
@@ -0,0 +1,55 @@
...
+
+      ssl         => true,
+      ssl_cert   => "/etc/ssl/private/${::fqdn}.bundle",
+      ssl_key    => "/etc/ssl/private/${::fqdn}.key",
+
+      add_header => {
+        'Strict-Transport-Security' => 'max-age=31536000',
+      },
+
+      rewrite_to_https => true;
```
chriskuehl: if this works, then we should update everywhere that we're manually doing 301's. we originally didn't use this because it was broken for some reason

modules/ocf_irc/templates/thelounge-conf.js.erb
```diff
@@ -0,0 +1,20 @@
...
+    name: "OCF IRC",
+    host: "irc.ocf.berkeley.edu",
+    port: 6697,
+    tls: true,
+    nick: "",
+    username: "",
+    realname: "",
+    join: "#rebuild"
+  },
+}
```
chriskuehl: why is this a template?

jvperrin:
@nattofriends I used fpm to package it up into a debian package, so it installs from there.

modules/ocf_irc/manifests/nodejs/apt.pp
jvperrin: Yeah, we just need newer than 4.2, so backporting from stretch would be fine.

nattofriends: I don't recall trouble running shout with unstable nodejs (4.4.7). What issue did you encounter?

modules/ocf_irc/manifests/nodejs/webirc.pp
jvperrin: Yeah, seems to work fine. I just copied this from the webssh config, and it seems to be working fine there too.

modules/ocf_irc/templates/thelounge-conf.js.erb
jvperrin: Oh, it originally had SSL certificates referenced in it that needed the hostname, but I removed that since it wouldn't have a HTTP -> HTTPS redirect. This shouldn't be a template now, thanks for catching that!

modules/ocf_irc/manifests/nodejs/apt.pp
jvperrin: Also I had a feeling you would say this :stuck_out_tongue:

jvperrin: We'd only want to backport for flood though, not for all servers, so doesn't this mean that backporting would have to be fixed first?

chriskuehl: We can probably just backport for all servers? Even if you fix backporting, we only have a jessie (not jessie-backports) distro in our apt repo. 

You could fix backporting and add a new dist, or backport with an equal version number and pin it with puppet on flood (a total hack), or just backport for all the servers. Your call, they all sound fine to me.

jvperrin: I don't think backporting for all servers would be a good idea, especially given the [number of API changes](https://github.com/nodejs/node/wiki/API-changes-between-v0.10-and-v4) between 0.10 and 4.0. We don't want to break any node apps we have running on apphosting just for this. I like the sounds of the first option the best, but I'm not really sure what it would involve.

jvperrin: Since backporting nodejs from stretch looks to be a massive pain because of the number of build dependencies needed to actually build it, I think it would be better to go with the nodesource repo. Besides, I trust them enough to use it, and we can go with the 4.x repo instead of the 6.x repo if that helps at all with stability? When we upgrade to stretch then this won't be a problem any more, since we'll have 4.x nodejs by default, and the nodesource version of nodejs will only be used on flood until then, so it shouldn't really be much of a problem.

PR #58, opened 2016-07-24
Fix dependency cycles with apt sources (mostly mesos)

chriskuehl:
We do stuff like this pretty frequently in our code:

```puppet
File {
  notify  => Service['mesos-slave'],
  require => Package['mesos'],
}
```

We thought this was just convenient shorthand for setting defaults inside a manifest. Turns out it's [way more sinister than that](https://docs.puppet.com/puppet/latest/reference/lang_defaults.html#behavior):

> Puppet still uses dynamic scope for resource defaults, even though it no longer uses dynamic variable lookup. This means that if you use a resource default statement in a class, it has the potential to affect any classes or defined types that class declares. See here for a full description of scope rules.

The reason we had these dependency cycles was because this class happened to get evaluated first, and it included indirectly `ocf::packages::docker::first_stage` which contained the apt repo resources. This lead to files like `/etc/apt/sources.list.d/docker.list` created by the apt plugin having a `require => Package['mesos']`, which is obviously invalid (since something in stage `first` can't depend on something in stage `main`).

This seems totally broken to me, the order manifests get included is an implementation detail and shouldn't affect the final result. And we do this all over our codebase and had no idea this is what it was doing. This is pretty scary; we should audit our uses of this pattern.

When we get to Puppet 4, we can use [per-expression defaults](https://docs.puppet.com/puppet/latest/reference/lang_resources_advanced.html#advanced-examples) which are much saner.

(The benefit of using these `first_stage` classes is that we don't need to add extra apt sources on all the servers, just on the servers that actually install packages from them. This was the best solution @daradib and I could come up with to the problem.)

### DIFF
--- a/modules/ocf/manifests/apt.pp
+++ b/modules/ocf/manifests/apt.pp
@@ -78,21 +78,6 @@ class ocf::apt {
     require  => Apt::Key['mesosphere'],
   }
 
-  apt::key { 'docker':
-    id     => '58118E89F3A912897C070ADBF76221572C52609D',
-    server => 'keyserver.ubuntu.com',
-  }
-
-  apt::source { 'docker':
-    location    => 'http://apt.dockerproject.org/repo',
-    release     => 'debian-jessie',
-    repos       => 'main',
-    include   => {
-      src => false
-    },
-    require     => Apt::Key['docker'];
-  }
-
   file { '/etc/cron.daily/ocf-apt':
     mode    => '0755',
     content => template('ocf/apt/ocf-apt.erb'),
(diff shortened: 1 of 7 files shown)


jvperrin:
Looks great to me, thanks for catching this! I was very confused why the `docker.list` file would require the mesos package, and spent a long time searching for where that relationship came from to no avail. It would definitely be nice if the dependency graph showed the lines that each dependency originated from, but oh well. Dynamic scope is really bad too :(

I hope we get to Puppet 4 soon for per-expression defaults and can also remove run stages for any apt things, since the puppetlabs-apt module is apparently not compatible with run stages.

chriskuehl:
I'm not sure it's true that the apt module is not compatible fwiw. Where do you see that?

(The only I see is many years old.)

jvperrin:
Well, it says more to [not split across run stages](https://github.com/puppetlabs/puppetlabs-apt/blob/master/README.md#limitations) than not to use them at all. I think avoiding run stages in general, except for things like setting the root password, or maybe some other basic setup, would be the simplest and least error-prone way to do things.

PR #65, opened 2016-08-12
Improve backports by giving updates to apt-dater (rt#3106)

jvperrin:
Updates for backported packages were previously done only by puppet, but it's better for apt-dater to perform these updates and only do the first package upgrade (normal -> backported) with puppet.

Also installing without recommends should work as intended now, before the behaviour was reversed, but now the default is to not install any recommends when using `ocf::repackage`.

One issue with changing this is that doing something like `require => Package['blah']` will only work if the package isn't being backported, since otherwise an `exec` is used instead of `package`, so the requirement breaks. I think the best behavior there would be to just require the parent `ocf::repackage` instead, which should work fine.

### DIFF
--- a/modules/ocf/manifests/repackage.pp
+++ b/modules/ocf/manifests/repackage.pp
@@ -1,6 +1,6 @@
 define ocf::repackage(
     $package     = $title,
-    $recommends  = false,
+    $recommends  = true,
     $backport_on = undef,
     $dist        = $::lsbdistcodename,
   ) {
--- a/modules/ocf/manifests/repackage.pp
+++ b/modules/ocf/manifests/repackage.pp
@@ -1,37 +1,29 @@
 define ocf::repackage(
     $package     = $title,
-    $recommends  = true,
+    $recommends  = false,
     $backport_on = undef,
     $dist        = $::lsbdistcodename,
   ) {
   $install_options = $recommends ? {
-    true    => ['--no-install-recommends'],
+    false   => ['--no-install-recommends'],
     default => []
   }
 
   if $backport_on and $dist == $backport_on {
-    #    apt::pin { "bpo-${package}":
-    #      release  => "${dist}-backports",
-    #      priority => 600,
-    #      packages => [$package];
-    #    }
-
-    package { $package:
-      # in case the non-backported package is already present
-      ensure          => latest,
-
-      # even though we've pinned the package, we still need -t dist-backports
-      # so that dependencies don't hold it back
-      install_options => concat($install_options, ['-t', "${dist}-backports"]);
-      #      require         => Apt::Pin["bpo-${package}"];
+    # We can't pin packages, because it won't install required dependencies that
+    # way, so we instead upgrade the package once (as long as it isn't a
+    # backport version already), and then future upgrades are done the normal
+    # way with apt-dater. It would be best if apt-dater did everything,
+    # including the original upgrade, but then a package and all its
+    # dependencies would need to be pinned, which is excessive and prone to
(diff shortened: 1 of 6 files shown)


modules/ocf/manifests/repackage.pp
```diff
@@ -1,37 +1,29 @@
 define ocf::repackage(
     $package     = $title,
-    $recommends  = true,
+    $recommends  = false,
```
chriskuehl: I think this should still default to true, as is the Debian default.

With this change, it's not obvious why we do `ocf::repackage` in some places instead of just `package` to someone looking at the code. With `ocf::repackage { 'package': recommends => false }` it's immediately obvious we're doing it to not install recommends.

Plus, it's weird that we have to set `recommends` back to `true` for packages we want to backport. `ocf::repackage` does multiple things, but it probably shouldn't do any of them by default.

modules/ocf_desktop/manifests/xfce.pp
```diff
@@ -1,7 +1,5 @@
 class ocf_desktop::xfce {
-  ocf::repackage { ['xfce4', 'xfce4-goodies', 'xfce4-notifyd']:
-    recommends => false;
-  }
+  ocf::repackage { ['xfce4', 'xfce4-goodies', 'xfce4-notifyd']:; }
 
   # TODO: figure out why this gets installed anyway on new systems
   # xfce4-power-manager asks for admin authentication on login in order to
```
chriskuehl: I think you figured out why this happens :)

chriskuehl:
> One issue with changing this is that doing something like require => Package['blah'] will only work if the package isn't being backported, since otherwise an exec is used instead of package, so the requirement breaks. I think the best behavior there would be to just require the parent ocf::repackage instead, which should work fine.

Agreed, users of `ocf::repackage` shouldn't be relying on implementation details.

modules/ocf_desktop/manifests/xfce.pp
jvperrin: Accidentally looks like I did!

modules/ocf/manifests/repackage.pp
jvperrin: Yeah, I do agree with that, we don't want to mix behaviors like recommends and backporting, as you said. I'll change it back and add back the `recommends => false` to all those packages.

modules/ocf_desktop/manifests/xfce.pp
```diff
@@ -3,7 +3,6 @@
     recommends => false;
   }
 
-  # TODO: figure out why this gets installed anyway on new systems
   # xfce4-power-manager asks for admin authentication on login in order to
   # "change laptop display brightness"
   package { 'xfce4-power-manager':
```
chriskuehl: we can probably remove this `ensure => absent` now since it won't get installed in the first place?

chriskuehl:
lgtm

modules/ocf_desktop/manifests/xfce.pp
jvperrin: Yeah, that's true, my concern was just that this hadn't run on all the desktops, but that doesn't really make sense, since all the desktops have run puppet pretty recently.

daradib:
Should we set `DEBIAN_FRONTEND=noninteractive`? My guess is that apt will be smart enough not to get stuck, but let's keep it in mind if we later have problems and need it.

chriskuehl:
yes probably

chriskuehl:
07b3e4e17b6f8157a36eecdb9fd19eff5abe9576

PR #66, opened 2016-08-15
Add virtual host support to mail server

chriskuehl:
This adds support to the mail server for serving virtual host mail.

We don't support storing mail (and thus don't have IMAP or webmail or anything); we think people won't really like this anyway. Instead, we provide mail *forwarding* and *sending* (submission).

We'll provide really good instructions on our website about how to configure Gmail to send via our SMTP server for that address. The idea is that people will forward to Gmail, and then use Gmail to send the reply mail (or compose new mail), which will be sent via our SMTP server. (Obviously you can use whatever forwarding address and SMTP client you want, but everyone uses Gmail anyway.)

Information goes into a new MySQL table:
```
MariaDB [ocfmail]> select * from addresses;
+------------------------------------+------------+--------------------------+---------------------+
| address                            | password   | forward_to               | last_updated        |
+------------------------------------+------------+--------------------------+---------------------+
| @derp.com                          | ...        | ckuehl@berkeley.edu      | 2016-08-14 20:40:48 |
| ckuehl2@dev-vhost.ocf.berkeley.edu | ...        | ckuehl+test@berkeley.edu | 2016-08-13 14:15:23 |
| ckuehl@dev-vhost.ocf.berkeley.edu  | ...        | ckuehl+test@berkeley.edu | 2016-08-13 14:15:23 |
| ckuehl@dev-vhost2.ocf.berkeley.edu | ...        | ckuehl+test@berkeley.edu | 2016-08-13 14:15:23 |
+------------------------------------+------------+--------------------------+---------------------+
4 rows in set (0.00 sec)
```

The passwords are salted and hashed using `crypt(3)`. Any `crypt`-format password will work on the verification side, but we'll generate them using [the same strong settings we use for our root password](https://github.com/ocf/utils/blob/ed9418d5eac922a1ead4e2c5c8a7241edafb0f75/staff/puppet/gen-rootpw#L2-L3).

This PR is enough to get the anthrax config totally working. The next part is writing functions in ocflib for querying and updating this database, and then adding a nice UI in ocfweb (this is why I added sessions/logins more prominently a few weeks ago).

For now we won't touch sandstorm, but once everything is in place, it will be easy to import existing `.forward` files, and then we can email groups about the changes.

### DIFF
--- a/modules/ocf_mail/files/site_vhost/trivial-table
+++ b/modules/ocf_mail/files/site_vhost/trivial-table
@@ -1,0 +1,4 @@
+# A trivial regexp table that always returns its key as the value.
+# This is used as the SASL [email] => [username] lookup table, since the
+# username always matches the email.
+/^(.*)$/ $1
--- a/modules/ocf_mail/templates/postfix/main.cf.erb
+++ b/modules/ocf_mail/templates/postfix/main.cf.erb
@@ -15,7 +15,7 @@ mydestination = int.ocf.berkeley.edu
 virtual_mailbox_domains =
 virtual_alias_domains =
     ocf.berkeley.edu,
-    mysql:/etc/postfix/vhost/mysql-alias-domains
+    mysql:/etc/postfix/vhost/mysql-alias-domains,
 
 # logging
 always_bcc = ocflog
@@ -23,8 +23,8 @@ always_bcc = ocflog
 # aliases
 virtual_alias_maps =
     hash:/etc/postfix/ocf/aliases-local,
-	proxy:ldap:/etc/postfix/ldap-aliases.cf,
-    mysql:/etc/postfix/vhost/mysql-alias-map
+    proxy:ldap:/etc/postfix/ldap-aliases.cf,
+    mysql:/etc/postfix/vhost/mysql-alias-map,
 alias_maps =
     hash:/etc/aliases,
     hash:/etc/aliases-group,
@@ -39,7 +39,7 @@ virtual_alias_recursion_limit = 2
 
 # environment variables for LDAP GSSAPI bind
 import_environment = MAIL_CONFIG MAIL_DEBUG MAIL_LOGTAG TZ XAUTHORITY DISPLAY
-	LANG=C KRB5CCNAME=FILE:/var/spool/postfix/krb5-cred
+    LANG=C KRB5CCNAME=FILE:/var/spool/postfix/krb5-cred
 
 # postmaster notifications
 notify_classes = resource, software
@@ -96,7 +96,7 @@ smtpd_client_restrictions =
     permit_dnswl_client list.dnswl.org,
     check_policy_service inet:127.0.0.1:10023,
     sleep 5,
-    permit
+    permit,
 
 # restrictions on HELO hostname
 # reject invalid HELO hostnames
@@ -105,7 +105,7 @@ smtpd_helo_restrictions =
     permit_sasl_authenticated,
     reject_invalid_helo_hostname,
     reject_non_fqdn_helo_hostname,
-    check_helo_access texthash:/etc/postfix/ocf/helo_access
+    check_helo_access texthash:/etc/postfix/ocf/helo_access,
 
 # restrictions on sender addresses
 # reject bad users/auth, non-existant domains, policyd-weight
@@ -117,7 +117,7 @@ smtpd_sender_restrictions =
     reject_authenticated_sender_login_mismatch,
     permit_mynetworks,
     permit_sasl_authenticated,
-    check_policy_service inet:127.0.0.1:12525
+    check_policy_service inet:127.0.0.1:12525,
 
 # restrictions on recipient addresses
 # reject bad users/auth, non-existant domains, open relay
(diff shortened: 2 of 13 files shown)


modules/ocf_mail/files/site_vhost/trivial-table
```diff
@@ -0,0 +1,4 @@
+# A trivial regexp table that always returns its key as the value.
+# This is used as the SASL [email] => [username] lookup table, since the
+# username always matches the email.
+/^(.*)$/ $1
```
chriskuehl: TODO: we can use the existing `mysql:/etc/postfix/vhost/mysql-alias-map` table for this

modules/ocf_mail/templates/postfix/main.cf.erb
```diff
@@ -13,15 +13,21 @@ myorigin = $mydomain
...
 # logging
 always_bcc = ocflog
 
 # aliases
-virtual_alias_maps = hash:/etc/postfix/ocf/aliases-local,
-	proxy:ldap:/etc/postfix/ldap-aliases.cf
-alias_maps = hash:/etc/aliases, hash:/etc/aliases-group
+virtual_alias_maps =
+    hash:/etc/postfix/ocf/aliases-local,
+	proxy:ldap:/etc/postfix/ldap-aliases.cf,
```
chriskuehl: derpy indentation

modules/ocf_mail/templates/site_vhost/mysql-alias-domains.erb
```diff
@@ -0,0 +1,8 @@
+hosts = mysql
+user = ocfmail-ro
+password = <%= @mysql_ro_password %>
+dbname = ocfmail
+query = SELECT domain FROM domains WHERE domain = '%s'
```
chriskuehl: `domains` is a view generated from the `addresses` table, so there is no duplication of data here

chriskuehl:
@daradib you are the expert, pls review?

modules/ocf_mail/templates/postfix/main.cf.erb
```diff
@@ -13,15 +13,21 @@ myorigin = $mydomain
...
+    mysql:/etc/postfix/vhost/mysql-alias-map
+alias_maps =
+    hash:/etc/aliases,
+    hash:/etc/aliases-group,
```
jvperrin: Should this have a trailing comma? None of the other comma-separated fields appear to, so I don't think there should be one here.

modules/ocf_mail/templates/site_vhost/mysql-alias-map.erb
```diff
@@ -0,0 +1,8 @@
+hosts = mysql
+user = ocfmail-ro
```
jvperrin: Could some duplication be removed between this template and the `mysql-alias-domains` template since they have the same content except for the SQL query? Maybe the query could be another variable that they each get?

modules/sandstorm/manifests/init.pp
```diff
@@ -1,5 +0,0 @@
-class sandstorm {
```
jvperrin: I guess sandstorm's manifest can just be removed for now and puppet runs on it will fail until it is completely shut down? I think it would be better to wait to disable sandstorm at all until the new system is fully working.

jvperrin:
Looks good, I'm excited to have better email support for vhosts, nice job! I don't really get how the postfix parts work, but I like the storing of the account data in MySQL, I think that'll work well for access through both ocfweb and SMTP and be easy to update. Also I'm glad we aren't actually storing any mail, since that sounded like a pain to do in the past, so this is a much better setup than that.

modules/ocf_mail/templates/postfix/main.cf.erb
chriskuehl: ah, not intentional. good catch!

though, since it appears to work, we should probably just add trailing commas to the rest :D

modules/sandstorm/manifests/init.pp
chriskuehl: will add it back for now

modules/ocf_mail/templates/site_vhost/mysql-alias-map.erb
chriskuehl: good idea, but i'll wait on this for now (until future parser is ready probably)

afaik it's not possible to pass variables to regular ERB templates explicitly, it just reads from the current scope (and puppet doesn't let you redefine variables in a manifest), so it wouldn't be straightforward to parameterize the query (would need two separate manifests, or a "common" partial, which seems overkill)

with EPP it looks easier, but we don't have this yet:
https://docs.puppet.com/puppet/latest/reference/lang_template.html

modules/ocf_mail/files/site_vhost/trivial-table
chriskuehl: almost, but not quite. i think this is actually better than the alternative (an extra mysql config file, and a useless mysql query)

chriskuehl:
merging now, but feel free to review @daradib :)

PR #69, opened 2016-08-30
Add xclip to desktops

mcint:
Added `xclip` to the xorg section. 

An [SO answer](http://stackoverflow.com/a/15971506) suggests this one might be a little more featureful, and that the extra dependency (libxmu6) over `xsel` is already used by `xterm` (confirmed) and other x-* applications.

### DIFF
--- a/modules/ocf_desktop/manifests/packages.pp
+++ b/modules/ocf_desktop/manifests/packages.pp
@@ -27,7 +27,7 @@ class ocf_desktop::packages {
     # performance improvements
     ['preload', 'readahead-fedora']:;
     # Xorg
-    ['xserver-xorg', 'xscreensaver','xclip']:;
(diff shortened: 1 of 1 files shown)


modules/ocf_desktop/manifests/packages.pp
```diff
@@ -27,7 +27,7 @@
     # performance improvements
     ['preload', 'readahead-fedora']:;
     # Xorg
-    ['xserver-xorg', 'xscreensaver']:;
+    ['xserver-xorg', 'xscreensaver','xclip']:;
```
chriskuehl: nit: space after comma for consistency :)

kpengboy:
Now I feel that xsel has shorter command line options and has approximately
the same amount of necessary functionality...

chriskuehl:
Make a PR adding xsel if you want it? You don't have to choose just one.

PR #71, opened 2016-09-02
Add php dependency to try out xibo

mcint:
[xibo](http://xibo.org.uk/) and see the [layouts](http://xibo.org.uk/layouts) people have created


chriskuehl:
`php` isn't a package on jessie. You probably want some other package, maybe `php5-cli` or something.

mcint:
Not sure what I want yet. Will install on my vm to determine required packages

PR #73, opened 2016-09-20
ocf_www: Move Apache DocumentRoot under /services

matthew-mcallister:
Setting the default DocumentRoot to /services/http/users causes per-directory
PHP config files (.user.ini) to be inherited in subdirectories.

See http://php.net/manual/en/configuration.file.per-user.php.

Test this on your website at dev-death.ocf.berkeley.edu! I don't see any obvious reason this should break anything.

### DIFF
--- a/modules/ocf_www/manifests/site/www.pp
+++ b/modules/ocf_www/manifests/site/www.pp
@@ -28,7 +28,7 @@ class ocf_www::site::www {
     servername      => 'www.ocf.berkeley.edu',
     serveraliases   => ['dev-www.ocf.berkeley.edu'],
     port            => 443,
-    docroot         => '/var/www/html',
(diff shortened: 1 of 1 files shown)


chriskuehl:
I'm not aware of any reason *not* to do this, just want to be careful it doesn't have unintended consequences. @daradib can you think of anything?

daradib:
I vaguely recall docroot headaches. I think we had docroot under /services or /mnt (legacy directory structure when we had /mnt/u{1..8}). Can you think of any issues with apache2-suexec-ocf or suphp? @sjyk concerns? If not go ahead.

chriskuehl:
Probably okay to go ahead in that case. I was most worried about any unintended security concerns of adding stuff to the docroot, but I can't find anything concrete.

PR #75, opened 2016-09-21
add 60s timeout for all print jobs

Baisang:
http://man7.org/linux/man-pages/man5/cupsd.conf.5.html

note that the default is 3 hours before killing a job...

We also have the option to limit the size of jobs submitted to the printer (see LimitRequestBody), but I think for now we should start with just killing jobs that are taking too long. We might be able to fix handling larger (~4MB) submitted jobs to the printer at a later time anyway.

Need to test somehow if possible.

### DIFF
--- a/modules/ocf_printhost/templates/cups/cupsd.conf.erb
+++ b/modules/ocf_printhost/templates/cups/cupsd.conf.erb
@@ -39,6 +39,9 @@ HostNameLookups On
 ServerAlias *
 DefaultAuthType Basic
 
+# Kill jobs after a certain time (stuck)
+MaxJobTime 60
+
 <Location />
   # Allow shared printing and remote administration...
   Order allow,deny
--- a/modules/ocf_printhost/templates/cups/cupsd.conf.erb
+++ b/modules/ocf_printhost/templates/cups/cupsd.conf.erb
@@ -41,6 +41,7 @@ DefaultAuthType Basic
 
 # Kill jobs after a certain time (stuck)
 MaxJobTime 60
+ErrorPolicy abort-job
 
 <Location />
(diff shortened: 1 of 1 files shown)


chriskuehl:
what kind of notification is given to the user when this timeout is hit?

jvperrin:
Nothing, it just kills the job, we're testing it right now. It kills the job but the printer is still hung up on the job, so it doesn't seem to help much. We really need some kind of way of filtering out jobs that will make the printers lock up, which is the main problem here, but it doesn't even seem to be size-based entirely. Maybe using the HP drivers would make it so that the printers get stuck less since they should sent correctly formatted data to the printers? Not sure on that though...

matthew-mcallister:
I predict that using HP drivers won't alleviate this issue, since the same thing happened with regularity on the old print system.

Baisang:
our test was not successful so thinking of adding this `ErrorPolicy` thing

Apparently the default `ErrorPolicy` is `stop-printer`, which will stop the current print queue? so changed to `abort-job` which hopefully just kills the job

http://superuser.com/questions/280396/how-to-resume-cups-printer-from-command-line

matthew-mcallister:
We've already been to this point. We can't change the policy to `abort-job` because a bunch of PDFs seem to mysteriously fail and then immediately succeed on reattempt. A had to patch out a line in enforcer to stop the email spam.

Baisang:
Fair enough. Will do more investigation tonight playing with CUPS settings.

jvperrin:
I'm closing this for inactivity and because while testing this didn't work to clear print jobs that take too long. It would definitely be good to fix the problem with jobs that take too long to process, but this doesn't quite do it unfortunately.

PR #78, opened 2016-09-26
Add npm for node projects

harrisonkuo:


chriskuehl:
Note that this gives you a really ancient version of npm. You probably want to use nodeenv instead for any real development.

PR #79, opened 2016-09-30
Add --environment flag to puppet-trigger

chriskuehl:
This allows you to run `puppet-trigger -e ckuehl` for example.

This should be more convenient than constantly messing with ldapvi and typing your /admin password all the time.

You'll need to put it back on production when you're done still (or just `puppet-trigger -e production`).

### DIFF
--- a/modules/ocf/files/puppet-trigger
+++ b/modules/ocf/files/puppet-trigger
@@ -6,6 +6,7 @@ import re
 import socket
 import subprocess
 import sys
+from textwrap import dedent
 
 
 def log(*args, file=sys.stderr, **kwargs):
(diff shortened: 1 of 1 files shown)


modules/ocf/files/puppet-trigger
```diff
@@ -40,13 +46,60 @@ def is_jessie():
...
+def switch_to_environment(env):
+    log('Switching environment to "{}".'.format(env))
+
+    hostname = socket.gethostname()
+    ldif = '''\
+dn: cn={hostname},ou=Hosts,dc=OCF,dc=Berkeley,dc=EDU
+changetype: modify
+replace: environment
+environment: {environment}
+'''.format(
```
jvperrin: Maybe use `dedent` for this to make it look cleaner (more consistent indentation)?

PR #80, opened 2016-10-07
ocf_www: generate vhosts outside of Puppet

chriskuehl:
Seems to work pretty well. Has pleasant output including a diff from the previous config file, runs a syntax check before reloading Apache, and reloads Apache only on changes.

The cronjob uses [chronic](https://joeyh.name/code/moreutils/) so that we get all this nice output, but only when it fails.

Remaining issues:

* The first vhost becomes the wildcard, but we want www.ocf to be the wildcard domain. The problem is that `conf.d` gets loaded before `sites-enabled.d`.
* Test a little more thoroughly.


kpengboy:
Bug 0: http://calslam.org redirects to http://calslam.org

chriskuehl:
good catch, fixed

modules/ocf_www/files/vhost.jinja
```diff
@@ -0,0 +1,58 @@
...
+
+    {% if vhost.use_ssl %}
+        # SSL
+        SSLEngine on
+        SSLCertificateFile {{vhost.ssl_cert}}
+        SSLCertificateKeyFile {{vhost.ssl_key}}
+        SSLCertificateChainFile {{vhost.ssl_chain}}
+        SSLCACertificatePath /etc/ssl/certs
+
+        # TODO: need the common SSL options here
```
chriskuehl: done

chriskuehl:
This seems to be working well on dev-death, trying it on death now.

chriskuehl:
Seems to work on death!

I'm going to go ahead and merge this because it's a large branch and don't want to leave it around for too long. Comments still welcome as usual :)

If you notice problems, reverting this branch and running Puppet on death should be sufficient to fully revert.

PR #81, opened 2016-10-12
added lab_close_notify script

dkess:
This is a Python script that runs in the background and sends notifications to the desktop user when the lab is about to close (discussed in [RT ticket #4560](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=4560)).

If the script is all good, the next step is actually getting it to run on login.


chriskuehl:
If you accept the invitation to the @ocf org then ocfbot will trust you and run automated builds for you :)

modules/ocf_desktop/files/xsession/lab_close_notify
```diff
@@ -0,0 +1,37 @@
...
+
+# 0 minutes, 2 minute, 5 minutes, 10 minutes, 15 minutes
+WARNING_DURATIONS = [0, 120, 300, 600, 900]
+
+def notify_user(seconds_left):
+    title = 'Lab closure'
+    if seconds_left != 0:
+        msg = 'Lab closing in {0:g} minutes'.format(seconds_left / 60)
+    else:
+        msg = 'Lab is closing NOW!'
```
chriskuehl: Maybe worth adding a bit more here to make it clear people need to be gone *now*?

e.g. `The lab is now closed. Please log out and leave right away.`

modules/ocf_desktop/files/xsession/lab_close_notify
```diff
@@ -0,0 +1,37 @@
...
+
+    # TODO: pick an icon
+    subprocess.call(['notify-send', title, msg])
+
+today = datetime.datetime.today()
+todays_hours = hours.REGULAR_HOURS[(today.weekday() - 1) % 7]
```
chriskuehl: This won't take into account holidays when the lab is sometimes open but with adjusted hours. I think you can use this instead:

```python
from ocflib.lab.hours import Day

...

todays_hours = Day.from_date().hours
```

That should do the right thing on both regular days and holidays.

modules/ocf_desktop/files/xsession/lab_close_notify
```diff
@@ -0,0 +1,37 @@
...
+closing_hours = [x.close for x in todays_hours]
+
+# unix timestamps (seconds since epoch) for the closing times today
+closing_times = [datetime.datetime.combine(today.date(), x).timestamp()
+        for x in closing_hours]
+
+s = sched.scheduler(time.time)
```
chriskuehl: neat, I hadn't used this module before

modules/ocf_desktop/files/xsession/lab_close_notify
```diff
@@ -0,0 +1,37 @@
...
+    # TODO: pick an icon
+    subprocess.call(['notify-send', title, msg])
+
+today = datetime.datetime.today()
+todays_hours = hours.REGULAR_HOURS[(today.weekday() - 1) % 7]
+closing_hours = [x.close for x in todays_hours]
+
+# unix timestamps (seconds since epoch) for the closing times today
+closing_times = [datetime.datetime.combine(today.date(), x).timestamp()
+        for x in closing_hours]
```
chriskuehl: i'd probably just combine with the above:

```python
closing_times = [
    datetime.datetime.combine(today.date(), hour.close).timestamp()
    for hour in Day.from_date().hours
]
```

your call

kpengboy:
I'd name the script `lab-close-notify` to be in line with the rest of the scripts we have in that directory.

modules/ocf_desktop/files/xsession/lab_close_notify
```diff
@@ -0,0 +1,37 @@
...
+    title = 'Lab closure'
+    if seconds_left != 0:
+        msg = 'Lab closing in {0:g} minutes'.format(seconds_left / 60)
+    else:
+        msg = 'Lab is closing NOW!'
+
+    # TODO: pick an icon
+    subprocess.call(['notify-send', title, msg])
+
+today = datetime.datetime.today()
```
kpengboy: Typically we wrap all of the main code into a `main()` method and then write at the bottom of the file

    if name == "__main__":
        main()

This way, if someone in the future wants to import your script as a module, the import statement won't suddenly start running code.

You can look at some of the other Python scripts here or the stuff in the `utils` repository to see examples.

modules/ocf_desktop/files/xsession/lab_close_notify
```diff
@@ -0,0 +1,37 @@
...
+
+def notify_user(seconds_left):
+    title = 'Lab closure'
+    if seconds_left != 0:
+        msg = 'Lab closing in {0:g} minutes'.format(seconds_left / 60)
+    else:
+        msg = 'Lab is closing NOW!'
+
+    # TODO: pick an icon
+    subprocess.call(['notify-send', title, msg])
```
kpengboy: As an alternative to calling an external process, another option is to use libnotify's Python bindings. The latter is what the `auto-lock` script currently uses.

modules/ocf_desktop/files/xsession/lab_close_notify
dkess: I usually like to split up long lines into shorter ones since it tends to make the code clearer and self-documenting-- but since this makes the line more concise anyways I'll go with your suggestion. Thanks!

modules/ocf_desktop/files/xsession/lab_close_notify
dkess: I get the reasoning for this, but it seems a bit overkill for a small script that I certainly don't expect anyone to import. However if this is really the style for all Python code at the OCF then I wont push it.

chriskuehl: The main benefit is that you can import the code without running it. This is how some linters (pylint) work (though it kinda sucks and is generally unsafe).

It's also somewhat useful during development because you can run the code in interactive mode (with `-i`) to get a shell without any of the code actually executing.

chriskuehl:
For the next steps, you'll want to:

* Provide this file on all of the desktops
* Modify the xsession script to launch it on login.

These should be pretty simple, but it might be useful to pair with another staff member during staff hours or after BoD if you'd like more context about the changes.

`notify.sh` should be a pretty good template to copy; it does almost the same thing as your script. [Here's](https://github.com/ocf/puppet/blob/216a58c9c671c26a36b2c9bc19297c568a9da83f/modules/ocf_desktop/manifests/xsession.pp#L11-L14) how it provides the file on the desktops, and [here's](https://github.com/ocf/puppet/blob/216a58c9c671c26a36b2c9bc19297c568a9da83f/modules/ocf_desktop/files/xsession/Xsession#L111-L112) how it launches itself on login. Both of these are also in the puppet repo, so you can do it on the same branch if you'd like.

chriskuehl:
lgtm minus some minor lint issues jenkins is complaining about: https://jenkins.ocf.berkeley.edu/job/puppet-test-pr/123/console

(you can use "make test" locally to run these as well. after the first time pre-commit should be installed and will in the future run when committing, and only on the files you change)

modules/ocf_desktop/manifests/xsession.pp
```diff
@@ -12,6 +12,10 @@
     '/opt/share/puppet/notify.sh':
       mode    => '0755',
       source  => 'puppet:///modules/ocf_desktop/xsession/notify.sh';
+	# script for warning users when the lab is about to close
```
matthew-mcallister: Style issue: watch the indentation.

dkess:
Sorry for being slow on these fixes-- it should all be done now. Let me know if there's anything else I forgot about.

As a side note, I got weird errors when trying to run `make test` on my own machine. [here's](http://sprunge.us/HUOg) the error log. I can't decipher it at all.

chriskuehl:
I assume that's your own personal computer and not one of the OCF machines? If so, how did you install pre-commit?

What's happening in that log is that you're using the latest version of pre-commit, but for Puppet we use my fork of pre-commit which has a few tweaks that make it better for the Puppet repo (most notably, it is able to detect the language of files without extensions based on their shebang, which is really important here since most of our scripts don't have a file extension and thus would miss out on most hooks otherwise). There are more details about that [here](https://github.com/pre-commit/pre-commit/pull/329). I'd still really like to upstream it eventually.

Assuming you installed it with pip, you can most likely fix it by `sudo pip3 uninstall pre-commit` and then `sudo pip3 install ckuehl-pre-commit-types==0.7.6.dev1`. (I don't really recommend `sudo pip install`-ing things, but it should mostly work.)

On an OCF machine (e.g. your staff VM), you should be able to just `sudo apt-get install pre-commit` instead of using pip since we install it as a Debian package.

chriskuehl:
lgtm! maybe work with a root staffer to test this on a desktop, if you haven't already?

dkess:
I tried to test it out in the lab today but ocflib wasn't updated so the notification never changed. Still, I am pretty sure it works since I tested it on my own machine.

I also added some icons that look like clocks (appointment-soon and appointment-missing).

chriskuehl:
cool, still lgtm, but i obviously haven't tested in the lab :)

PR #82, opened 2016-10-13
ocf_mail: tag mail sent by vhost users with group uid (rt#5376)

chriskuehl:
@jvperrin @matthew-mcallister @dkess thoughts?

Context: we keep a log of mail messages that we send so that we can easily spot trouble users (usually hijacked users sending thousands of mails per day). Each time a message is sent, we call this script (`log-mail`), which parses some metadata from the headers and then adds a line to the end of the file `/var/log/ocfmail.log`.

We then have a cronjob that runs each morning which parses that file and looks for problems (sending too much mail, sending mail claiming to be from other users, etc).

Right now for mail sent by vhost users using SMTP, we don't tag the messages with any `uid`, so we wouldn't notice if a group was sending spam mail this way.

### DIFF
--- a/modules/ocf_mail/files/spam/logging/log-mail
+++ b/modules/ocf_mail/files/spam/logging/log-mail
@@ -9,26 +9,53 @@ import email
 import email.utils
 import io
 import json
+import pwd
 import re
 import sys
 from datetime import datetime
 
+from ocflib.vhost.mail import get_mail_vhosts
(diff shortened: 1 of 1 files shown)


jvperrin:
Looks great to me, I didn't know this wasn't already accounted for in our current logs, but it definitely should be. :+1:

PR #83, opened 2016-10-13
Add more www server aliases

kpengboy:
In preparation for redirecting unrecognized vhosts to a vhost
disabled page


kpengboy:
One alternative proposal I can think of is listing out every disabled vhost using a special syntax in vhosts.conf, and having them redirect to a special "vhost disabled" page. But I personally think our current plan of making the vhost disabled page the default vhost is less maintenance.

chriskuehl:
> But I personally think our current plan of making the vhost disabled page the default vhost is less maintenance.

agreed, i like this approach better

kpengboy:
Actually I'll squash the two commits here together and directly commit the result to master.

PR #84, opened 2016-10-13
Preload ocf.io HSTS

daradib:
Before merging, will need to deal with hello.ocf.io and stats.ocf.io which don't have proper certificates.

This is for [HSTS Preload List Submission](https://hstspreload.appspot.com/).

### DIFF
--- a/modules/ocf_www/manifests/site/ocfweb_redirects.pp
+++ b/modules/ocf_www/manifests/site/ocfweb_redirects.pp
@@ -114,7 +114,6 @@ class ocf_www::site::ocfweb_redirects {
       'dev-hello.ocf.berkeley.edu',
       'dev-staff.ocf.berkeley.edu',
       'hello',
-      'hello.ocf.io',
       'staff.ocf.berkeley.edu',
     ],
     port            => 80,
--- a/modules/ocf_www/manifests/site/shorturl.pp
+++ b/modules/ocf_www/manifests/site/shorturl.pp
@@ -85,7 +85,7 @@ class ocf_www::site::shorturl {
       {rewrite_rule => '^/~?([a-z]{3,16}(?:/.*)?)$ https://www.ocf.berkeley.edu/~$1 [R]'},
     ],
 
-    headers       => ['always set Strict-Transport-Security max-age=31536000'],
+    headers       => ['always set Strict-Transport-Security "max-age=31536000; includeSubDomains; preload"'],
   }
 
   # canonical redirects


chriskuehl:
imo let's just not support subdomains on ocf.io, we don't (or shouldn't -- sometimes someone emails them) advertise them, and i don't like the added complexity of getting *.ocf.io certs

staff can use them for VMs if they want, but then it's their responsibility to get https working

daradib:
> imo let's just not support subdomains on ocf.io

+1, we should use ocf.io/shortname for that purpose. I believe we have publicized hello.ocf.io though so we might want to get a certificate for it and deprecate it.

chriskuehl:
yeah, unfortunately it's in some recent presentations we've given :(

what about just setting it up as a regular vhost for the `staff` user with a `.htaccess` redirect so that an LE cert will be acquired automatically for it?

chriskuehl:
we could replace this entire module with a simple regular ~staff vhost and `.htaccess`: https://github.com/ocf/puppet/blob/master/modules/ocf_www/manifests/site/ocfweb_redirects.pp

is that a good approach?

pros:
* much simpler
* free ssl

cons:
* config no longer is source control

anything else?

daradib:
> what about just setting it up as a regular vhost for the `staff` user with a `.htaccess` redirect so that an LE cert will be acquired automatically for it?

Done. Manually invocated `build-vhosts` but I'll wait for `lets-encrypt-update` in a few minutes.

> we could replace this entire manifest with a simple regular ~staff vhost and `.htaccess`

I didn't do it now, especially because of the `dev-` vhost handling, but seems like a reasonable idea. We may want to version control ~staff, especially vhost.conf and even bod minutes.

chriskuehl:
cool thanks. the https version is live (but without hsts currently -- i commented on rt#5378 with details)

i don't think that blocks you from submitting to the preload list, though?

daradib:
> i don't think that blocks you from submitting to the preload list, though?

Not blocked, we can go ahead! Let's wait for an ack from the SMs, paging @jvperrin, @matthew-mcallister.

jvperrin:
Sounds good to me, feel free to submit it. I do think versioning vhosts and some of the staff directory would be a good thing to do first, so once we do that then we can change this manifest to just be another vhost.

daradib:
https://hstspreload.appspot.com/?domain=ocf.io
> ocf.io is now pending inclusion in the HSTS preload list!
> 
> Please make sure that ocf.io continues to satisfy all preload requirement, or it will be removed. Please revisit this site over the next few weeks to check on the status of your domain.

PR #91, opened 2016-10-23
Automatically clean docker cruft

kpengboy:
This sets up cronjobs to remove containers older than 2 weeks, dangling images, and dangling volumes.

### DIFF
--- a/modules/ocf/manifests/packages/docker.pp
+++ b/modules/ocf/manifests/packages/docker.pp
@@ -37,4 +37,22 @@ class ocf::packages::docker($admin_group = 'docker') {
     require => Package['docker-engine'],
     notify  => Exec['docker-socket-update'];
   }
+
+  cron {
+    'clean-old-docker-containers':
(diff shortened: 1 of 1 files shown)


modules/ocf/manifests/packages/docker.pp
```diff
@@ -37,4 +37,23 @@
     require => Package['docker-engine'],
     notify  => Exec['docker-socket-update'];
   }
+
+  cron {
+    'clean-old-docker-containers':
+      command => "docker ps -a --filter status=exited | grep -E '(weeks|months|years) ago' | awk '{print \$1}' | chronic xargs -r docker rm",
```
chriskuehl: i guess that should be `(weeks|month|year)` so that singular of the latter two get caught?

i'd be okay with anything at least a day old and exited, but this seems fine too

kpengboy:
Do you think it's worthwhile or a regrettable idea to automatically prune docker volumes? I haven't really used Docker volumes other than bind mounts, and apparently volumes are designed to persist data independently of containers.

chriskuehl:
I think it's okay. We can always change it later, but so far I have almost never used Docker volumes either.

PR #93, opened 2016-10-31
U2F

mcint:
This diff enables u2f usage on desktops, per https://www.yubico.com/faq/enable-u2f-linux/

It is based on rules here, which is ahead of (but included in) the u2f-host deb package
https://github.com/Yubico/libu2f-host/blob/master/70-u2f.rules

### DIFF
--- a/modules/ocf_desktop/manifests/udev.pp
+++ b/modules/ocf_desktop/manifests/udev.pp
@@ -1,0 +1,8 @@
+class ocf_desktop::udev {
+  package { ['u2f-host']:; }
+
+  file { '/etc/udev/rules.d/70-u2f-fido.rules':
+    source    => 'puppet:///modules/ocf_desktop/70-u2f-fido.rules',
+    mode      => '0644',
+  }
+}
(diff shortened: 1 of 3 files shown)


nattofriends:
This is the version of 70-u2f.rules included in libu2f-host0 for jessie-backports:

https://anonscm.debian.org/git/pkg-auth/libu2f-host.git/tree/70-u2f.rules

Is support for "JaCarta U2F" and "HS ePass FIDO" important enough to copy this file in instead of installing libu2f-host0?

mcint:
All I really care about is the FIDO U2F one. I also submitted a pull request for it to Yubico/libu2f-host today. It looks like they're fairly fast, 1 week, and 1 day for the last successful pull requests. Not sure how long debian will take from there though.

modules/ocf_desktop/manifests/udev.pp
```diff
@@ -0,0 +1,8 @@
+package { ['libu2f-host']:; }
```
chriskuehl: this should be inside the class

matthew-mcallister:
Consider squashing some of the minor commits.

modules/ocf_desktop/manifests/udev.pp
```diff
@@ -0,0 +1,8 @@
+class ocf_desktop::udev {
+  package { ['libu2f-host']:; }
```
jvperrin: This looks like it should be `libu2f-host0` since `libu2f-host` isn't in the package archives.

mcint:
@jvperrin Good call. Meant to add the `u2f-host` package, not `libu2f-host0`, which even says "you don't normally install this manually"

PR #95, opened 2016-11-17
add new alias to get the weather

Baisang:
saw this on reddit and thought it was neat because i'm too lazy to open a browser and google weather berkeley all the time


chriskuehl:
imo this belongs in a binary in utils and not in every user's shell environment

Baisang:
Makes sense. So basically make a script called weather that does the same command?

chriskuehl:
yup. you get the added benefit of it working in all shells

PR #96, opened 2016-11-20
Add hiera and systemd dropins and mesos node attributes

chriskuehl:
Hiera is the big important change I guess. It's not used for much yet, but we can use it to refactor some of the yuckier parts of our Puppet code.


matthew-mcallister:
Well, with three distinct and significant changes, this PR is kind of hard to evaluate, but here it goes.

Adding hiera is a pretty unambiguously useful change. It would allow us to basically clear out most of `nodes.pp` and get rid of all `hostname.startswith('dev-')` junk. I'm not very familiar with hiera myself, but it can't hurt to add it, though it will take more scrutiny when rewriting old code.

Overriding default systemd services sounds nice, though it seems like it's only needed for the mesos module.

Support for changing mesos node attributes and refreshing the slave state I can't really weigh in on, but I'm glad we will be able to run create on marathon instead of on supernova.

chriskuehl:
> Overriding default systemd services sounds nice, though it seems like it's only needed for the mesos module.

There are likely to be other situations in the future, e.g. @kpengboy wanted to change the default storage driver for Docker.

PR #97, opened 2016-11-20
Use hiera for more things

chriskuehl:
Currently tested on:
* raptors (my staff VM)
* supernova
* eruption
* anthrax
* hal
* jaws
* pandemic
* werewolves
* tsunami
* lightning
* destruction

This will require changing `type` to `staffvm` for all staff VMs in LDAP.


chriskuehl:
This also fixes the remaining issue from ocf.io/rt/5412 (ntp syncing on physical machines which are not jaws/hal/pandemic)

chriskuehl:
@jvperrin @matthew-mcallister I think this is ready for review.

Things I'd still like to do:

* Use hiera to declare class defaults, e.g. nginx (finally getting rid of the deprecation warnings it gives us)
* Move `puppetClass` and `puppetVar: owner=x` from LDAP into hiera so that it's actually visible and version-controlled along with the puppet environment

chriskuehl:
I think we can also use this to simplify run stages elsewhere...

modules/ocf/manifests/motd.pp
```diff
@@ -1,12 +1,12 @@
-class ocf::motd {
-  $is_virtual = str2bool($::is_virtual)
-
-  if tagged('ocf_ssh') or tagged('ocf_apphost') {
+class ocf::motd($nfs_motd = false) {
```
jvperrin: I don't think I'd name it this, since not all servers mounting NFS are using this motd, only the public-facing ones. Maybe something like `$public` or `$public_motd` would be more appropriate?

modules/ocf/manifests/packages/ntp.pp
```diff
@@ -1,16 +1,16 @@
-class ocf::packages::ntp {
+class ocf::packages::ntp($master = false, $peers = []) {
```
jvperrin: Nice, this is much better

modules/ocf/manifests/staff_users.pp
```diff
@@ -1,11 +1,10 @@
...
 
-class ocf::staff_users {
+class ocf::staff_users($noop = false) {
   # Only create home directories if:
   #   - we don't mount NFS
-  #   - we're a server (not a desktop)
-  #   - we're not the file server
-  if !str2bool($::ocf_nfs) and $::type == 'server' and !tagged('ocf_filehost') {
+  #   - it hasn't been turned off in Hiera
+  if !str2bool($::ocf_nfs) and !$noop {
```
jvperrin: This does what's wanted, but it just seems more fragile to me, since it's essentially defined in two different places which systems don't create staff directories (facts and hiera). In my opinion it was nicer when it was restricted to servers only, since now the noop must explicitly opt-out, which seems like an issue waiting to happen. Maybe it could be made opt-in instead and then enabled for all servers? Not sure that's much better since then jaws is still an exception, but this current system seems like it could sometimes have problems otherwise.

modules/ocf/templates/ntp.conf.erb
```diff
@@ -15,16 +15,8 @@ server 1.pool.ntp.org iburst
 server 2.pool.ntp.org iburst
 server 3.pool.ntp.org iburst
 
-# TODO: this is terrible
```
jvperrin: Yes, yes it was

hieradata/common.yaml
```diff
@@ -1 +1,19 @@
...
+    - ocf::munin::node
+    - ocf::networking
+    - ocf::packages
+    - ocf::puppet
+    - ocf::rootpw
+    - ocf::staff_users
+    - ocf::systemd
+    - ocf::utils
+
+packages: {}
```
jvperrin: Meh, seems like this shouldn't be duplicated both here and in the `ocf::packages` class. I can just see people adding packages here instead of in the correct class, when it would be much better to have that all in the manifests if possible.

modules/ocf/manifests/packages.pp
```diff
@@ -10,14 +10,14 @@
...
-  include ocf::packages::memtest
-  include ocf::packages::microcode
   include ocf::packages::ntp
   include ocf::packages::rsync
   include ocf::packages::shell
-  include ocf::packages::smart
   include ocf::packages::ssh
 
+  # any packages from hiera
+  create_resources(package, hiera('packages'))
```
jvperrin: Not a huge fan of `create_resources` as it's often abused. I do like that it at least makes it obvious that the data is coming from hiera though, that's nice.

modules/ocf/manifests/rootpw.pp
```diff
@@ -5,7 +5,7 @@
 # there is no other option.
 #
 # To regenerate the root password, see /opt/puppet/scripts/gen-rootpw
-class ocf::rootpw {
+class ocf::rootpw($stage = 'first') {
```
jvperrin: Having the stage here definitely clarifies when it runs, so that's a lot nicer.

hieradata/type/staffvm.yaml
```diff
@@ -0,0 +1,3 @@
+ocf::auth::ulogin: [['%{owner}', 'ALL']]
+ocf::auth::usudo: ['%{owner}']
+ocf::auth::nopasswd: true
```
jvperrin: Cool, these are perfect values to go in hiera in my opinion, they make it a lot more clear who is getting these settings and who isn't.

hieradata/nodes/dev-anthrax.yaml
```diff
@@ -0,0 +1 @@
+ocf::packages::postfix::manage: false
```
jvperrin: Can we do this based on tagged class instead of hostname so that production and `dev-` servers get the same configuration with less duplication? (unless overridden by hostname of course)

modules/ocf/manifests/packages/postfix.pp
```diff
@@ -1,39 +1,36 @@
-class ocf::packages::postfix {
+class ocf::packages::postfix($manage = true) {
```
jvperrin: I think it would be clearer if these kinds of variables had their default values defined in `common.yaml`, and then were loaded with a `hiera()` call to make it more explicit where the value is coming from. Otherwise I can see someone (or me) searching all the manifests trying to find where this is passed as an argument and being quite confused until they realize that hiera is overwriting this value. Or worse, sometimes this might be passed as an argument in some places and in other places as a hiera value, which would be even harder to find.

modules/ocf/manifests/staff_users.pp
chriskuehl: i think this entire design is bad; we should instead just add pam_mkhomedir which fixes this

jvperrin: Yeah, good point, that would be much nicer than either method

hieradata/common.yaml
chriskuehl: the key has to be defined for all hosts, so we either don't have it at all, or it has to be here

hieradata/nodes/dev-anthrax.yaml
chriskuehl: i think using tagged classes is an antipattern; they're not particularly reliable (depend on parse order, etc.)

if we used a role pattern, that would be an ideal place to put it

hieradata/common.yaml
jvperrin: I'd vote for not having it at all then, seems like a rabbit hole I really don't want to go down.

modules/ocf/manifests/motd.pp
chriskuehl: maybe we should have an attribute for whether a host is public, and then calculate this based on "public && nfs" (we already have an nfs fact)

hieradata/nodes/dev-anthrax.yaml
jvperrin: That's a good point, I hadn't thought about the parse order issues there. This would be fine for now then, a little duplication doesn't hurt. Would be something to think about, maybe having a role pattern would be the way to go then...

hieradata/common.yaml
chriskuehl: how do you suggest to do this: https://github.com/ocf/puppet/blob/789dec481116bebdf69620d0014dd059155ec8a3/hieradata/is_virtual/false.yaml

make a manifest for each package we might want to install?

modules/ocf/manifests/motd.pp
jvperrin: Alright, sounds like a good idea to me, that seems to cover all the times we'd want to potentially show that motd, and I'm sure having a public attribute would be useful elsewhere.

hieradata/common.yaml
jvperrin: I'm not sure what the best way would be, but I think making a separate manifest for it, or a conditional somewhere would be better than doing this. Maybe I'm exaggerating how much this would be abused, but I do see it as a source of confusion because it could be so broadly used.

hieradata/nodes/dev-anthrax.yaml
jvperrin: Things are a lot nicer now, that does look like a bit of a mess. `tagged` works fine, but I see your point that it's brittle, so it would be good to use roles instead, since we're most of the way there. I'll look into being able to have multiple roles somehow, it could be useful.

modules/ocf/manifests/staff_users.pp
chriskuehl: to summarize from irc: agreed that this sucks, let's improve it

opened #98

PR #99, opened 2016-11-26
Refactor ocf_ssl to allow providing additional bundles

chriskuehl:
Sometimes we have the need to provide provide additional SSL certs/keys besides the main one based on the vhost. (In the past we did this for virtual hosts; another reason now is so that we can use a wildcard cert, since InCommon doesn't seem to have an option to use both a wildcard and other SANs on the same cert.)

This lets us easily define additional bundles, e.g.:

```puppet
ocf_ssl::bundle { 'wildcard.agent.mesos.ocf.berkeley.edu':; }
```

The default bundle is also provided this way now.

### DIFF
--- a/hieradata/nodes/fireball.yaml
+++ b/hieradata/nodes/fireball.yaml
@@ -1,6 +1,6 @@
 classes:
     - ocf::packages::docker
     - ocf_ocfweb::dev_config
-    - ocf_ssl
+    - ocf_ssl::default_bundle
 
 owner: jvperrin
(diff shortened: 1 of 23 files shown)


hieradata/nodes/fireball.yaml
```diff
@@ -1,6 +1,6 @@
 classes:
     - ocf::packages::docker
     - ocf_ocfweb::dev_config
-    - ocf_ssl
+    - ocf_ssl::default_bundle
```
chriskuehl: I think this is a good example of why managing classes via hiera is much nicer than LDAP. Previously I would have had to try to coordinate this change when pushing this to master, making sure people didn't revert it in their branches if they changed the environment without pulling master, etc.

chriskuehl:
merging because this blocks me, feel free to comment though

PR #103, opened 2016-12-06
puppet-trigger: show syslog from only puppet-agent

kpengboy:
Comments? Good idea? Bad idea?

### DIFF
--- a/modules/ocf/files/puppet-trigger
+++ b/modules/ocf/files/puppet-trigger
@@ -17,7 +17,7 @@ def log(*args, file=sys.stderr, **kwargs):
 
 def trigger_run():
     if subprocess.call(
-            ('/usr/sbin/service', 'puppet', 'status'),
+            ('systemctl', 'status', 'puppet.service'),
             stdout=subprocess.DEVNULL,
     ) == 0:
         with open('/var/run/puppet/agent.pid') as f:
@@ -33,10 +33,8 @@ def trigger_run():
         sys.exit(1)
 
 
-def tail_syslog():
-    # call less in "secure" mode since this is triggered by passwordless sudo
-    os.environ['LESSSECURE'] = '1'
-    os.execvp('less', ('less', '+F', '/var/log/syslog'))
+def show_journal():
+    os.execvp('journalctl', ('journalctl', '-f', '-u', 'puppet.service'))
 
 
 def switch_to_environment(env):
@@ -92,7 +90,7 @@ def main(argv=None):
     trigger_run()
 
     if args.tail:
-        tail_syslog()
+        show_journal()
 
 
 if __name__ == '__main__':


modules/ocf/files/puppet-trigger
```diff
@@ -17,7 +17,7 @@ def log(*args, file=sys.stderr, **kwargs):
 
 def trigger_run():
     if subprocess.call(
-            ('/usr/sbin/service', 'puppet', 'status'),
+            ('systemctl', 'status', 'puppet.service'),
```
chriskuehl: this seems good

modules/ocf/files/puppet-trigger
```diff
@@ -33,10 +33,8 @@ def trigger_run():
         sys.exit(1)
 
 
-def tail_syslog():
-    # call less in "secure" mode since this is triggered by passwordless sudo
-    os.environ['LESSSECURE'] = '1'
-    os.execvp('less', ('less', '+F', '/var/log/syslog'))
+def show_journal():
+    os.execvp('journalctl', ('journalctl', '-f', '-u', 'puppet.service'))
```
chriskuehl: i think you still want the `LESSSECURE=1` environment variable since journalctl invokes `less` (well, `PAGER`, which will default to `less`)

kpengboy: Although `journalctl -f` doesn't seem to invoke `less`...

chriskuehl:
I'm -1.5 on this change overall:

* **-1** because this means you only see the Puppet logs.
  
  I know that was your intention, but I think this is a real downside (and is why e.g. we always tailed `syslog` rather than just `daemon.log`, which contains less other things).
  
  There are lots of things that can go wrong without seeing it in Puppet, and I typically find it really useful to see logs from other services as Puppet restarts them. Consider the case where you break a config file: it depends entirely on the service whether you'll even see that starting the service failed in the Puppet log, e.g.:
    * some services exit zero on some errors
    * some services fork and systemd doesn't get their true exit status
    * some services take too long to exit nonzero, so puppet will still see them as starting successfully)

  Overall I think it's better to be watching all of the system logs when making Puppet changes, as Puppet can have impact every service on the system.

* **-1** because `journalctl` has a really annoying bug where you can't stop tailing logs with `less` (it kills `less` and itself when you `Ctrl-C` (interrupt to stop tailing). I have [a shitty workaround](https://gist.github.com/chriskuehl/9ab4c74c19f2f4cb883744171335ac44) with some more details, but it doesn't solve all the problems (still can't scroll up forever like you can with `less` on a file, or resume tailing). You can also work around the buggy behavior by piping `journalctl` to `less` yourself, but with the same problems as my hack.

  I find it really useful to be able to stop tailing a log and scroll back up to read something that got printed, then later resume tailing.

* **+0.5** because `journalctl` has some nice colors.

chriskuehl:
I think a better approach here would be to add a mode that runs the Puppet agent with `--no-daemonize --debug --show_diff --onetime` once in the foreground (you'll need to stop Puppet first, though, and probably start it again at the end).

This would be pretty useful because it provides more detail than we currently get (e.g. augeas errors), and nice coloring (better than journalctl, since puppet chooses the coloring rather than journalctl which just greps for "error" and similar). I wouldn't use it all the time, but there are definitely times it would have been handy (usually I stop puppet and run it manually with the flags above).

modules/ocf/files/puppet-trigger
chriskuehl: true, apparently if you tail it doesn't

though i think that's another regression since now you can't scroll :(

kpengboy:
I had some of the same concerns with regards to log selection. I'll leave the question open. As for your other idea, I might let another staffer do it or get around to it eventually.

PR #106, opened 2016-12-16
ocf_desktop: rewrite notify to use ocflib

matthew-mcallister:


modules/ocf_desktop/files/xsession/notify
```diff
@@ -0,0 +1,67 @@
...
+QUOTA_FORMAT = """\
+    → {semesterly} remaining this semester
+    → {daily} remaining today"""
+
+
+old_quota = None
+
+
+def show_quota_loop(scheduler):
+    global old_quota
```
matthew-mcallister: This could optionally be passed in as a mutable dict instead.

modules/ocf_desktop/files/xsession/notify
```diff
@@ -0,0 +1,67 @@
...
+from ocflib.printing.quota import get_connection
+from ocflib.printing.quota import get_quota
+
+
+NOTIFICATION_ICON = '/opt/share/xsession/images/ocf-color.png'
+
+# Format taken from paper command
+QUOTA_FORMAT = """\
+    → {semesterly} remaining this semester
+    → {daily} remaining today"""
```
chriskuehl: should we just subprocess out to paper rather than duplicating its logic?

modules/ocf_desktop/files/xsession/notify
```diff
@@ -0,0 +1,67 @@
...
+    scheduler.run()
+
+
+def main():
+    Notify.init('OCF')
+
+    staff = [
+        session.user
+        for session in staff_in_lab()
+        if session.user not in {'pubstaff', 'optstaff'}
```
chriskuehl: is "optstaff" a typo?

also, i thought they weren't going to have an account (and just use their own accounts)?

modules/ocf_desktop/files/xsession/notify
```diff
@@ -0,0 +1,67 @@
...
+    scheduler.enter(60, 0, show_quota_loop, argument=(scheduler,))
+    scheduler.run()
+
+
+def main():
+    Notify.init('OCF')
+
+    staff = [
+        session.user
+        for session in staff_in_lab()
```
chriskuehl: wonder if it's better to use the ocfweb endpoint? idk
https://www.ocf.berkeley.edu/stats/api/staff-in-lab

it would reduce load on mysql, idk

modules/ocf_desktop/manifests/xsession.pp
```diff
@@ -11,9 +11,9 @@
       source  => 'puppet:///modules/ocf_desktop/xsession/Xsession',
       require => File['/opt/share/xsession'];
     # printing and other notification script daemon
-    '/opt/share/puppet/notify.sh':
+    '/opt/share/puppet/notify':
       mode    => '0755',
-      source  => 'puppet:///modules/ocf_desktop/xsession/notify.sh';
+      source  => 'puppet:///modules/ocf_desktop/xsession/notify';
```
chriskuehl: should we ensure => absent the old path for a bit?

matthew-mcallister: We'll be installing stretch on the desktops soon enough, so I don't think it really matters.

modules/ocf_desktop/files/xsession/notify
matthew-mcallister: Yeah, I'll delete that

modules/ocf_desktop/manifests/xsession.pp
chriskuehl: ok

modules/ocf_desktop/files/xsession/notify
matthew-mcallister: Nah, the color formatting used by paper only works in a shell, plus we already get the username and quota so we can tell if it changed. The only code copied is the use of the "→" character.

modules/ocf_desktop/files/xsession/notify
matthew-mcallister: Well, it would put that API to some good use, but then the script will break if ocfweb changes or goes down. Also, this script already queries mysql once every minute, so the load difference sounds very marginal.

I don't really know, maybe @jvperrin can decide.

modules/ocf_desktop/files/xsession/notify
```diff
@@ -0,0 +1,67 @@
...
+                time=datetime.datetime.now().strftime('%-I:%M%P')
+            ),
+            QUOTA_FORMAT.format(semesterly=quota.semesterly, daily=quota.daily),
+            NOTIFICATION_ICON,
+        ).show()
+        old_quota = quota
+
+    # Schedule self to repeat next minute
+    scheduler.enter(60, 0, show_quota_loop, argument=(scheduler,))
+    scheduler.run()
```
jvperrin: Any reason to have this separate like this instead of just sleeping for 60 seconds in a loop? I think this is fine as it is, but it's not like it needs to run precisely every minute (which scheduler is better for), so just sleeping seems like it would get rid of the global variable and clean this up a little. Thoughts?

modules/ocf_desktop/files/xsession/notify
chriskuehl: we should fix ocflib not to output colors when stdout isn't a tty anyway. doesn't that solve the problem? (just assert the entire output of the script, like the old one did)

modules/ocf_desktop/files/xsession/notify
chriskuehl: ok, let's delete from ocfweb though if nothing is using it

modules/ocf_desktop/files/xsession/notify
```diff
@@ -0,0 +1,67 @@
...
+
+
+def main():
+    Notify.init('OCF')
+
+    staff = [
+        session.user
+        for session in staff_in_lab()
+        if session.user != 'pubstaff'
```
jvperrin: Could do something like `if session.user != 'pubstaff' and not is_staff(user, 'opstaff')` if you want to exclude opstaff here. Maybe we should make this into a separate ocflib method if we're going to keep [repeating](https://github.com/ocf/ocflib/blob/7100e2c4a7342462649b62a3ddcbcea73f450814/ocflib/printing/quota.py#L63) it though?

modules/ocf_desktop/files/xsession/notify
matthew-mcallister: That does sound better, it's not like it needs to run at a set time or anything. I guess `sched` sounded more proper to me, but it's probably overkill.

modules/ocf_desktop/files/xsession/notify
jvperrin: Yeah, sounds good, sorry about that since you just added it yesterday... :/

matthew-mcallister: Yeah, I admit I didn't even realize @chriskuehl had made an API for it since puppet hadn't run on eruption so the script was still broken.

modules/ocf_desktop/files/xsession/notify
jvperrin: It would be better if it's running at a set time, since otherwise the time will drift by however long it takes to execute each time, but this is a pretty short script so that shouldn't matter. Thankfully timing definitely doesn't matter much for this script...

modules/ocf_desktop/files/xsession/notify
matthew-mcallister: We could add a function, but pubstaff is supposed to go away in a couple of months, so we'll be getting rid of it soon.

jvperrin: Alright, good point, sounds fine then.

modules/ocf_desktop/files/xsession/notify
matthew-mcallister: OK, that sounds like a plan.

modules/ocf_desktop/files/xsession/notify
chriskuehl: its ok

chriskuehl: i'll wait to delete until you merge though

matthew-mcallister:
Merge pending approval

matthew-mcallister:
No need to modify paper, going through subprocess strips out the formatting characters.

jvperrin:
Huh, that's cool, I didn't know that it did that.

chriskuehl:
subprocess doesn't strip anything; it returns you bytes without knowing anythng about the context. the stripping happens in ocflib where it correctly doesn't print color if the stdout is not a tty (you can compare `paper` and `paper | cat` in your terminal): https://github.com/ocf/ocflib/commit/a90df9387d4e49267abb275be0706ff10b0e8a6f

PR #107, opened 2016-12-20
Install firefox beta from experimental

daradib:
mozilla.debian.net does not support stretch yet,
so we can use firefox 45esr, firefox release from unstable,
or firefox beta from experimental. Once mozilla.debian.net
supports stretch, switch back to them.


modules/ocf/manifests/packages/firefox.pp
```diff
@@ -3,14 +3,20 @@
     stage => first,
   }
 
-  $firefox_pkg = $::lsbdistcodename ? {
-    'jessie' => 'firefox',
-    default  => 'firefox-esr',
+  if $::lsbdistcodename == 'jessie' {
```
jvperrin: I don't think you need the if statement here, `ocf::repackage` will install the regular way (using a `package` resource) if it doesn't match the distribution to "backport" on (not actually backporting in this case), so just having the `ocf::repackage` resource below should be enough.

modules/ocf/manifests/repackage.pp
```diff
@@ -2,22 +2,22 @@
     $package     = $title,
     $recommends  = true,
     $backport_on = '',
-    $dist        = $::lsbdistcodename,
+    $dist        = "${::lsbdistcodename}-backports",
```
jvperrin: Nice, I think this will be a good way to do this and make it easier to install from different sources.

modules/ocf/manifests/packages/firefox.pp
daradib: I agree, I think an array makes sense here.

modules/ocf/manifests/repackage.pp
```diff
@@ -17,16 +17,18 @@
...
+      before      => Package[$package];
     }
-  } else {
-    package { $package:
-      install_options => $install_options;
-    }
+  }
+
+  package { $package:
+    install_options => $install_options;
```
jvperrin: That's really nice, this has been mildly annoying for a while, so this will be much better!

modules/ocf/manifests/packages/firefox.pp
daradib: Good catch. We can probably do this with `aptitude`, `apt-show-versions`, or `apt-cache policy`.

jvperrin: From my limited testing right now `apt-show-versions` would be easiest to work with, so I think using that it would be doable.

daradib: I'm trying with `apt-cache policy`. `apt-show-versions` is easier but installs a daily cronjob and an apt hook, so it seemed a bit heavy-weight for our purposes. But it's a great package to see where packages are installed from.

jvperrin: Alright, I didn't notice that before, so that is good to know. I still think it would make things a lot easier though, even if it has a cronjob and apt hook. The issue with using `apt-cache policy` is that it shows all the sources that it could be installed from, not just the one it actually is installed from, which sounds a lot harder to work with.

daradib: It marks the currently-installed one with `***`, so I'm using `grep -A1` to get the line after that, but I agree that this is...a bit cumbersome.

daradib: Feel free to change it. I was reluctant to add a new package (and its cron and apt hooks) to every box, but SM should not feel reluctant!

daradib:
@chriskuehl any objections regarding `ocf::repackage`? I'll merge for now since this is not master yet.

modules/ocf/manifests/packages/firefox.pp
jvperrin: No, I think this is fine for now, but I'll keep looking at `apt-show-versions`, maybe it would be worth using.

PR #109, opened 2016-12-31
Add weechat to extra packages

dkess:
Irssi and ZNC are in there, so I figured Weechat should as well.


jvperrin:
Nice, looks good!

PR #110, opened 2017-01-01
ocf_desktop: switch from lightdm-gtk-greeter-ocf to pam_trimspaces

kpengboy:


modules/ocf_desktop/manifests/xsession.pp
```diff
@@ -93,7 +93,21 @@
...
+  # add pam_trimspaces to lightdm PAM stack
+  augeas { 'lightdm-pam_trimspaces':
+    context => '/files/etc/pam.d/lightdm',
+    changes => [
+      'ins #comment after #comment[1]',
+      'set #comment[2] "Strip leading and trailing space from username"',
+      'ins 01 after #comment[2]',
+      'set 01/type auth',
+      'set 01/control requisite',
+      'set 01/module pam_trimspaces.so',
```
chriskuehl: some days i think augeas is pretty cool and other days i think it's insane

jvperrin: Same

PR #111, opened 2017-01-02
Fix issue with rotating log files

jvperrin:
This fixes the [issue](https://munin.ocf.berkeley.edu/static/dynazoom.html?plugin_name=ocf.berkeley.edu%2Fdemocracy.ocf.berkeley.edu%2Fdf_inode&start_iso8601=2016-12-23T19%3A35%3A52-0700&stop_iso8601=2017-01-01T13%3A30%3A02-0700&start_epoch=1482546952.08&stop_epoch=1483302602.16&lower_limit=&upper_limit=&size_x=800&size_y=400&cgiurl_graph=%2Fmunin-cgi%2Fmunin-cgi-graph) with log rotation on democracy.

I also renamed the log files to end in `.log` and make the log files not world-readable but still readable to `ocfstaff` (slightly more secure, since before they were world-readable and could contain sensitive data).

### DIFF
--- a/modules/ocf_syslog/files/logrotate
+++ b/modules/ocf_syslog/files/logrotate
@@ -4,7 +4,6 @@
     rotate 14
     compress
     delaycompress
-    create 0640 root ocfstaff
 
     sharedscripts
     postrotate
--- a/modules/ocf_syslog/files/logrotate
+++ b/modules/ocf_syslog/files/logrotate
@@ -1,9 +1,10 @@
-/var/log/remote/* {
-	daily
-	missingok
-	rotate 14
-	compress
+/var/log/remote/*.log {
+    daily
+    missingok
+    rotate 14
+    compress
     delaycompress
+    create 0640 root ocfstaff
 
     sharedscripts
     postrotate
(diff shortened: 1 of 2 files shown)


modules/ocf_syslog/files/logrotate
```diff
@@ -1,9 +1,10 @@
...
-	missingok
-	rotate 14
-	compress
+/var/log/remote/*.log {
+    daily
+    missingok
+    rotate 14
+    compress
     delaycompress
+    create 0640 root ocfstaff
```
chriskuehl: imo we probably don't want this, since it means any stream that ever existed will continue to exist forever (in empty rotated log files), even stuff like typo'd stream names (that's why i originally left it out)

jvperrin: Ah I see, I wondered why you left out `notifempty` too, but that makes a lot more sense now.

Sounds fine to me, looks like I can change some rsyslog stuff to get the same behavior without the issue with typos.

jvperrin:
Ok, now rsyslog does the permissions changing, and logrotate doesn't.

modules/ocf_syslog/files/logrotate
jvperrin: Ok, now rsyslog does the permissions changing, and logrotate doesn't.

PR #112, opened 2017-01-04
Generalize systemd override

kpengboy:
Generalize `ocf::systemd::override` to handle any type of unit, and make use of that to set the docker socket group with systemd drop-ins.

Please review.

### DIFF
--- a/modules/ocf/manifests/systemd/override.pp
+++ b/modules/ocf/manifests/systemd/override.pp
@@ -21,4 +21,11 @@ define ocf::systemd::override(
     notify  => Exec['systemd-reload'],
     require => Package['systemd-sysv'],
   }
+
+  if $unit =~ /(.*)\.service$/ {
+    ensure_resource('service', $1)
+
+    File["/etc/systemd/system/${unit}.d/${title}.conf"] ~> Service[$1]
+    Exec['systemd-reload'] -> Service[$1]
+  }
 }
--- a/modules/ocf/manifests/systemd/override.pp
+++ b/modules/ocf/manifests/systemd/override.pp
@@ -7,21 +7,18 @@
 #     ExecStart=
 #     ExecStart=/path/to/my/thing
 define ocf::systemd::override(
-    $service,
+    $unit,
     $source = undef,
     $content = undef,
 ) {
-  ensure_resource('file', "/etc/systemd/system/${service}.service.d", {
+  ensure_resource('file', "/etc/systemd/system/${unit}.d", {
     ensure => directory,
   })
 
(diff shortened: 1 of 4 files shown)


modules/ocf_irc/manifests/ircd.pp
```diff
@@ -8,10 +8,10 @@
   }
 
   ocf::systemd::override { 'start-inspircd-before-anope':
-    service => 'inspircd',
-    content => "[Unit]\nBefore=anope.service",
+    unit    => 'inspircd.service',
+    content => "[Unit]\nBefore=anope.service\n",
     require => Package['inspircd', 'anope'],
-    before  => Service['inspircd'],
+    notify  => Service['inspircd'],
```
chriskuehl: is there a guarantee that this notify happens after systemd gets reloaded? that's the reason `before` was used instead of notify.

modules/ocf/manifests/systemd/override.pp
```diff
@@ -7,21 +7,18 @@
...
-    content  => $content,
-    notify   => [
-      Exec['systemd-reload'],
-      Service[$service],
-    ],
-    require  => Package['systemd-sysv'],
+  file { "/etc/systemd/system/${unit}.d/${title}.conf":
+    source  => $source,
+    content => $content,
+    notify  => Exec['systemd-reload'],
```
kpengboy: It's up to the caller to notify any service that needs to be reloaded now. Is this too prone to programmer error? Are we worried about any potential race conditions, of `Service` being refreshed before `systemd-reload` happens?

modules/ocf_irc/manifests/ircd.pp
kpengboy: Well, `notify` implies `before`, but I did touch upon your point in your previous comment.

modules/ocf/manifests/systemd/override.pp
kpengboy: Also, is the status quo not subject to the same potential race?

modules/ocf_irc/manifests/ircd.pp
chriskuehl: ok

modules/ocf/manifests/systemd/override.pp
jvperrin: I think it's fine if the caller is left to notify the service it overrides, but that should probably be mentioned at the top of this file so that it is clear it is necessary.

kpengboy:
Rebasing on top of current master

kpengboy:
Comments?

kpengboy:
None? OK, I'm merging...

PR #113, opened 2017-01-05
Only allow certain names for remote syslog files

jvperrin:
The security scanner was probing our rsyslog server and made files like `#000#000#000@#000#003#000#000user#000pgbouncer#000UTF8#000.log` (except longer) that we don't really want, so now any files not matching a similar regex to the one in stdin2syslog will be put in a generic `errors.log` file.

This seems a reasonable amount more secure too, since there aren't as many arbitrary files that someone could create.

### DIFF
--- a/modules/ocf_syslog/files/ocf.conf
+++ b/modules/ocf_syslog/files/ocf.conf
@@ -12,6 +12,19 @@ template(name="ocf-remote" type="list") {
     constant(value="\n")
 }
 
+# template for errored remote log lines
+template(name="ocf-errors" type="list") {
+    constant(value="ERRORED_TAG[")
+    property(name="syslogtag")
+    constant(value="]  MESSAGE[")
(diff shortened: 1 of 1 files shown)


modules/ocf_syslog/files/ocf.conf
```diff
@@ -23,5 +36,17 @@ template(name="ocf-remote-file" type="list") {
 
 # rules for storing the logs
 ruleset(name="ocf-remote") {
-    action(type="omfile" dynaFile="ocf-remote-file" template="ocf-remote" fileGroup="ocfstaff" fileCreateMode="0640")
+    # This regex was fiendish, since the hyphen must be next to the end bracket
+    # for it to work.
```
chriskuehl: lmao

jvperrin: Yeah, took me a while to figure that one out, fun stuff

PR #114, opened 2017-01-06
Better mirror monitoring & Kali push mirroring

jvperrin:
This separates out the monitoring of ftpsync mirrors from the rest of the sync process. This should make it easier to add new monitoring for non-ftpsync projects by just making a different heath-checking script for that project.

This also makes the hour for ftpsync configurable in case syncing should be run less often than every hour.

Finally, this adds push syncing for Kali, which has a long SSH key, but I don't think there's a better way to add it really.


modules/ocf_mirrors/manifests/ftpsync.pp
```diff
@@ -1,20 +1,16 @@
 # ftpsync is a Debian tool for reliably mirroring Debian archives:
 # https://www.debian.org/mirror/ftpmirror
 define ocf_mirrors::ftpsync(
-    $cron_minute,
     $rsync_host,
+    $cron_minute = undef,
```
chriskuehl: is there ever a case we won't want a cron minute? i figure even with push mirroring, we'd still want to sync at least once a day, just in case

jvperrin: No, if isn't set, then [cron isn't even used at all](https://github.com/ocf/puppet/pull/114/files#diff-8c8f6ab7251650b2233929d55bdb08f1R45), which makes it possible to define a push-only triggered ftpsync for example. Not super useful right now, but I thought we might want to do this sometime with Kali if push mirroring works fine.

jvperrin: Whoops, read the last part of your comment lol.

Yeah, probably true on that, we would want to sync periodically anyway, so I'll change this back, good point.

PR #115, opened 2017-01-12
Upgrade puppet pre-commit hooks

chriskuehl:
We're using puppet 4 for validation now, but that's probably fine since soon we'll be on Puppet 4.


jvperrin:
I did update Jenkins plugins today, so maybe that's why this is failing? I'll look into it a bit more.

jvperrin:
Alright, I upgraded a couple remaining plugins and it appears to be running now, so hopefully it's fixed again.

PR #116, opened 2017-01-14
Replace Rack puppet master with puppetserver

jvperrin:
Puppetserver is not my favorite, mostly because it's a scary all-in-one JVM application, but it does have some especially nice features, like being able to support both Puppet 3 and Puppet 4 clients in the same server, and being able to use PuppetDB. The puppet documentation does say that the Rack puppet master will be removed in the future, so it seems like a good idea to migrate to puppetserver now, since it also allows for using Puppet 4 with no barriers.

Also all the current errors being emailed are coming from [this issue](https://tickets.puppetlabs.com/browse/MODULES-3310), which appears to be some kind of issue when using Puppet 3 agents with puppetserver and the concat module, which we do almost everywhere now, so anything using the concat module (pretty much just used for joining SSL certs right now into bundles) will error until the concat module is downgraded.

### DIFF
--- a/hiera.yaml
+++ b/hiera.yaml
@@ -10,7 +10,7 @@
     - common
 
 :yaml:
-    :datadir: /opt/puppet/env/%{environment}/hieradata
+    :datadir: /etc/puppetlabs/code/environments/%{environment}/hieradata
 
 :puppet:
     :datasource: data
--- a/modules/ocf_puppet/files/ldap-enc
+++ b/modules/ocf_puppet/files/ldap-enc
@@ -35,21 +35,16 @@ def main():
         # Could not find a unique node to classify, exit
         sys.exit(1)
 
-    output = {'parameters': {}}
+    output = {'parameters': {}, 'environment': host['environment'][0]}
 
     for key, values in host.items():
-        # The environment is a special parameter that needs some conversion to
-        # work correctly.
-        if key == 'environment':
-            output['environment'] = values[0]
-
         # Remove the value from its list only if it is a singular value
         if len(values) == 1:
             output['parameters'][key] = values[0]
         else:
             output['parameters'][key] = values
 
-    print(yaml.dump(output, default_flow_style=False))
+    print(yaml.safe_dump(output, default_flow_style=False))
 
 
 if __name__ == '__main__':
--- a/modules/ocf_puppet/files/ldap-enc
+++ b/modules/ocf_puppet/files/ldap-enc
@@ -10,13 +10,13 @@ nodes like the built-in LDAP classifier.
 This is that classifier! It takes in node FQDNs, and looks them up in LDAP
 to get variables and classes to give to Puppet.
 """
-
 import argparse
 import sys
-import yaml
 
-from ocflib.infra.hosts import hosts_by_filter
+import yaml
 from ocflib.infra.hosts import hostname_from_domain
+from ocflib.infra.hosts import hosts_by_filter
+
 
 def main():
     parser = argparse.ArgumentParser(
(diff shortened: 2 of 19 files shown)


hiera.yaml
```diff
@@ -10,7 +10,7 @@
     - common
 
 :yaml:
-    :datadir: /opt/puppet/env/%{environment}/hieradata
+    :datadir: /etc/puppetlabs/code/environments/%{environment}/hieradata
```
chriskuehl: lol i kinda hate this new path

modules/ocf_puppet/files/ldap-enc
```diff
@@ -0,0 +1,56 @@
...
+        if key == 'environment':
+            output['environment'] = values[0]
+
+        # Remove the value from its list only if it is a singular value
+        if len(values) == 1:
+            output['parameters'][key] = values[0]
+        else:
+            output['parameters'][key] = values
+
+    print(yaml.dump(output, default_flow_style=False))
```
chriskuehl: i'd s/dump/safe_dump/ as a general best practice

!sux pyyaml

modules/ocf_puppet/files/ldap-enc
```diff
@@ -0,0 +1,56 @@
...
+        # Could not find a unique node to classify, exit
+        sys.exit(1)
+
+    output = {'parameters': {}}
+
+    for key, values in host.items():
+        # The environment is a special parameter that needs some conversion to
+        # work correctly.
+        if key == 'environment':
+            output['environment'] = values[0]
```
chriskuehl: can you just move this out of the loop?

```python
output['environment'], = host['environment']

for key, values in host.items():
    if len(values) == 1:
        ...
```

modules/ocf_puppet/manifests/puppetmaster.pp
```diff
@@ -1,52 +1,75 @@
...
-    ssl_verify_depth  => 1,
-    ssl_options       => ['+StdEnvVars', '+ExportCertData'],
-
-    rack_base_uris => ['/'];
+  # Set correct memory limits on puppetserver so that it doesn't run out
+  augeas { '/etc/default/puppetserver':
+    context => '/files/etc/default/puppetserver',
+    changes => [
+      "set JAVA_ARGS '\"-Xms1g -Xmx1g -XX:MaxPermSize=512m\"'",
+    ],
```
chriskuehl: lol thanks java

hiera.yaml
jvperrin: Same, it's really long, but there is a symlink still in place if you want to edit things at the old location

modules/ocf_puppet/files/ldap-enc
jvperrin: Yeah, sounds fine, they should all have an environment set in the LDAP data, so that's a fair assumption to make

modules/ocf_puppet/manifests/puppetmaster.pp
jvperrin: Yeah, not a fan of the memory usage and startup time of puppet server, but it does serve requests faster, so I guess that's a reasonable tradeoff? /shrug

modules/ocf_puppet/files/ldap-enc
jvperrin: Alright, didn't know that was a thing, but it certainly sounds safer :stuck_out_tongue:

chriskuehl: if the pyyaml people were sane, they'd have renamed `load` to `very_dangerous_load` and `safe_load` to `load`

afaik `dump` isn't all that dangerous, but it can generate output that can't be parsed by `safe_load` (or by non-pyyaml yaml parsers), so i'd still suggest to avoid it

jvperrin:
Thanks, I just wish it didn't take so long to get puppeted and cleaned up, because stretch is coming very soon now...

PR #118, opened 2017-01-18
Add nvidia-settings to ocf_desktops

harrisonkuo:
Need nvidia-settings to remove screen tearing script, not sure about NUC.

### DIFF
--- a/modules/ocf/manifests/extrapackages.pp
+++ b/modules/ocf/manifests/extrapackages.pp
@@ -85,6 +85,7 @@ class ocf::extrapackages {
     'nasm',
     'nodejs',
     'npm',
+    'nvidia-settings',
     'octave',
     'pandoc',
     'pdfchain',


chriskuehl:
also probably only on desktops with nvidia? we have some logic for that already: https://github.com/ocf/puppet/blob/a4cda0ac13a0ac17abc58f82466f38b4b320b34c/modules/ocf_desktop/manifests/drivers.pp#L5-L14

jvperrin:
`nvidia-settings` has a backported version too, so that one would probably work better and can just be added to this line: https://github.com/ocf/puppet/blob/master/modules/ocf_desktop/manifests/drivers.pp#L8

harrisonkuo:
Woops, forgot about that one.

matthew-mcallister:
Please don't close PRs prematurely, it's better to work on the existing branch so the discussion is preserved.

PR #120, opened 2017-01-19
Stretch desktops

jvperrin:
This is pretty much done now, it pretty much just needs the network interface naming to be fixed up, which can be done now that we have regular access to the lab, unlike over break, where something could have easily broken and not been fixable until the semester started. A few packages also should be figured out, but most of them are compatible already and appear to be working fine.

My plan is to gradually enable this on desktops in the lab by installing stretch on them, starting with the row eruption is on, and growing from there if there aren't any issues.


jvperrin:
Interesting, so it makes a merge commit to resolve the conflicts, hadn't tried that before...

jvperrin:
Alright, I think this is pretty much all done now, and three desktops are now on stretch! (`eruption`, `asteroid`, and `chaos`)

jvperrin:
It sorta does, but it didn't have the intended effect. Instead of pinning only `unstable`, it pinned everything from anywhere in `mirrors` to 150, which meant that packages were installed from unstable when they shouldn't have been. I changed it to be in an `apt::pin` instead with the release name and that seems to help.

Strange behavior though, not at all what I would expect it to do...

jvperrin:
Any comments on this? I'm thinking of merging this soon if not since more than half the lab is using stretch now and I haven't heard of any problems recently.

jvperrin:
Now named "Stretch desktops and one KVM server"

jvperrin:
Ok, I'll merge this then, since there don't seem to be any suggestions and it's pretty stable.

matthew-mcallister:
~Awesome~

PR #121, opened 2017-01-20
Update DRI and AccelMethod.

harrisonkuo:


chriskuehl:
what does this do?

modules/ocf_tv/files/X11/xorg.conf
```diff
@@ -1,5 +1,7 @@
 Section "Device"
    Identifier  "Intel Graphics"
    Driver      "intel"
+   Option      "AccelMethod"    "sna"
```
daradib: Thanks! This is only needed on jessie though. It is the default in jessie-backports and stretch. I'd suggest upgrading.

modules/ocf_tv/files/X11/xorg.conf
```diff
@@ -1,5 +1,7 @@
...
    Option      "TearFree"    "true"
+   Option      "DRI"    "3"
```
daradib: I don't think this will work on jessie (see "reverting to software rendering" in xorg log) and shouldn't be needed in jessie-backports and stretch, where it sets the maximum DRI level I think (default is no maximum).

jvperrin:
@chriskuehl This is supposed to prevent screen tearing on the TV, but testing for it is ongoing I believe.

daradib:
Have you tried installing xserver-xorg-video-intel from backports (assuming you're on jessie)?

jvperrin:
It is from jessie, but I don't believe that's been tried, that's a good idea.

harrisonkuo:
Yeah I'll try the backported version of the package when I get the chance.

jvperrin:
@harrisonkuo Now that the TV is on stretch, this should be a little easier (no backport needed).

harrisonkuo:
Closed, flags default on Stretch packages.

PR #123, opened 2017-01-30
[ocf_www]: Add ocf-gmail link as /gmail

mcint:


matthew-mcallister:
Perhaps this would be better used to point to https://www.ocf.berkeley.edu/docs/services/vhost/mail/ or maybe https://www.ocf.berkeley.edu/account/vhost/mail/.

matthew-mcallister:
You may as well put `/mail` to some use while you're at it.

PR #124, opened 2017-02-04
Set bridge-nf-call-iptables=0 on stretch (rt#5849)

chriskuehl:


chriskuehl:
heh weird, it does seem to work on riptide, but i didn't test it thoroughly. i just picked the most popular sysctl module from the forge

for now i guess i'll merge and we can re-think if it causes errors

PR #125, opened 2017-02-05
Add getty on serial ports

kpengboy:
Comments? Criticism?

### DIFF
--- a/modules/ocf/manifests/serial_getty.pp
+++ b/modules/ocf/manifests/serial_getty.pp
@@ -1,8 +1,10 @@
 class ocf::serial_getty {
   service { 'serial-getty@ttyS0':
-    ensure   => running,
     enable   => true,
     provider => systemd,
     require  => Package['systemd-sysv'],
+  } ~>
+  exec { 'systemctl start serial-getty@ttyS0':
+    refreshonly => true,
   }
 }
--- a/modules/ocf/manifests/serial_getty.pp
+++ b/modules/ocf/manifests/serial_getty.pp
@@ -1,0 +1,8 @@
+class ocf::serial_getty {
+  service { 'serial-getty@ttyS0':
+    ensure   => running,
+    enable   => true,
+    provider => systemd,
+    require  => Package['systemd-sysv'],
+  }
+}
--- a/modules/ocf/manifests/serial_getty.pp
+++ b/modules/ocf/manifests/serial_getty.pp
@@ -1,10 +1,12 @@
 class ocf::serial_getty {
-  service { 'serial-getty@ttyS0':
-    enable   => true,
-    provider => systemd,
-    require  => Package['systemd-sysv'],
-  } ~>
-  exec { 'systemctl start serial-getty@ttyS0':
-    refreshonly => true,
+  if str2bool($::is_virtual) or member(hiera('classes'), 'ocf_kvm') {
+    service { 'serial-getty@ttyS0':
+      enable   => true,
+      provider => systemd,
+      require  => Package['systemd-sysv'],
+    } ~>
+    exec { 'systemctl start serial-getty@ttyS0':
+      refreshonly => true,
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


jvperrin:
Do we want to restrict it to VMs or run on all systems? Right now it's on all, but I thought we just wanted it on VMs?

jvperrin:
I don't know much about getty though, so I think @chriskuehl would be able to review this more effectively.

chriskuehl:
I also don't know much about getty, but it seems like we'd only want VMs + the physical servers? If it doesn't hurt on other things (desktops, nuc, etc.) then I'm okay with leaving it.

modules/ocf/manifests/serial_getty.pp
```diff
@@ -0,0 +1,8 @@
+class ocf::serial_getty {
+  service { 'serial-getty@ttyS0':
+    ensure   => running,
```
chriskuehl: is it necessary to `ensure => running`? generally i am not a fan of this pattern. i explained in some other PR, but i think it's a slippery slope that leads to puppet being used as a shitty service supervisor rather than just to manage configuration

i have seen really bad cases where e.g. a server is not ready for use after boot until puppet has run

kpengboy: The intention of this line is to actually start getty on the serial line after it's been enabled. Otherwise we'd need to manually start it or wait for a reboot.

Do you have a better proposal for how we should write this?

jvperrin: I think it's probably fine to leave it disabled as long as it will run on the next restart, but I dunno...

kpengboy:
I haven't seen any immediately obvious adverse effects from enabling getty on the (I'm pretty sure nonexistent) serial lines on the desktops and the NUC. However, it wouldn't be too difficult to restrict this to only VMs and physical servers...

modules/ocf/manifests/serial_getty.pp
kpengboy: ...it can often be months between server reboots, though, and an emergency could potentially develop during this time which forces us to use the serial console...

chriskuehl: I don't have a good alternative if we care that it gets started immediately, unfortunately. On stretch you could do

```puppet
exec { 'systemctl enable --now serial-getty@ttyS0':
  creates => '/etc/systemd/system/getty.target.wants/serial-getty@ttyS0.service',
}
```

but jessie doesn't support `--now`

jvperrin: I mean we should have physical access for the next few months, and the VMs can be accessed with VNC too, so I don't see this as something that needs to be running straight off. It would be nice though to have running now, so maybe add `ensure => running` for now and put a TODO to remove it later? I don't think I have a better solution than that.

daradib: Can you enable with an exec and notify the service from the exec (or subscribe the other way)?

chriskuehl: @daradib that's what i tried, but it didn't actually start it :\

```puppet
exec { 'systemctl enable --now serial-getty@ttyS0':
  creates => '/etc/systemd/system/getty.target.wants/serial-getty@ttyS0.service',
} ~>
service { 'serial-getty@ttyS0': provider => systemd }
```

does enable it, but just leaves it stopped (though it does notify it, not sure exactly what it's doing)

daradib: Does the debug log have any helpful info?

modules/ocf/manifests/serial_getty.pp
```diff
@@ -1,10 +1,12 @@
 class ocf::serial_getty {
-  service { 'serial-getty@ttyS0':
-    enable   => true,
-    provider => systemd,
-    require  => Package['systemd-sysv'],
-  } ~>
-  exec { 'systemctl start serial-getty@ttyS0':
-    refreshonly => true,
+  if str2bool($::is_virtual) or member(hiera('classes'), 'ocf_kvm') {
```
kpengboy: Would it be better instead to directly add `ocf::serial_getty` to the hypervisor nodes in hiera?

chriskuehl: probably not. i wish we had a "role" hiera layer, but such a thing is hard to implement

can you write `member(hiera('classes'), 'ocf_kvm') ` as `tagged('ocf_kvm')` though? i think that is slightly cleaner

PR #126, opened 2017-02-07
ocf_docker: restart docker registry after docker engine is upgraded

kpengboy:
Concerns?


chriskuehl:
did you test it? if so, lgtm

PR #127, opened 2017-02-08
Install docker on staff VMs

matthew-mcallister:
I think staff will appreciate getting to play around with docker on their VMs.

### DIFF
--- a/hieradata/nodes/fireball.yaml
+++ b/hieradata/nodes/fireball.yaml
@@ -1,5 +1,4 @@
 classes:
-    - ocf::packages::docker
     - ocf_ocfweb::dev_config
     - ocf_ssl::default_bundle
 
--- a/hieradata/type/staffvm.yaml
+++ b/hieradata/type/staffvm.yaml
@@ -1,3 +1,6 @@
+classes:
+  - ocf::packages::docker
+
 ocf::auth::ulogin: [["%{hiera('owner')}", 'ALL']]
 ocf::auth::usudo: ["%{hiera('owner')}"]
 ocf::auth::nopasswd: true


chriskuehl:
I kind of feel like we should leave staff VMs as blank as possible, and instead encourage people to add classes like this via hiera. For example, why install docker but not extrapackages?

Also this module might conflict with config people already have on their VMs.

matthew-mcallister:
> I kind of feel like we should leave staff VMs as blank as possible, and instead encourage people to add classes like this via hiera. For example, why install docker but not extrapackages?

I agree with this, and I'm sure not every person cares to have docker on their VM, but i think new staff would be much more encouraged to play around with it if they didn't have to make a pull request to puppet or manage their own apt sources just to get it installed. I like the idea that staff can autonomously use their VMs as sandboxes without having to learn and go through all our systems and infrastructure (puppet, github) to get things going.

> Also this module might conflict with config people already have on their VMs.

This may be a concern, so I suppose it would be wise to hold back on merging this. This would most likely conflict with any staff who have the older `docker.io` package installed on their VMs, which would be best solved by uninstalling `docker.io` first. I don't know if there are any more complicated cases though.

chriskuehl:
What about just asking the people you're making new VMs for whether they'd like you to add Docker? I guess I'm not sure why Docker is special compared to everything else we could install.

matthew-mcallister:
I was just thinking the same thing and had the idea to add it to the form just sent out. I think that sounds like a better solution at the moment.

For my closing comment, I still think that docker _is_ particular in that it's broadly useful (not some esoteric service or util), probably wanted by a lot of people, and, most importantly, requires _outside intervention_ for staff to install, meaning they have to make a PR (or they can manage apt manually, but that's kind of higher-maintenance). This hinders staff from using their VMs independently, which I think is the strongest draw.

Also, as you know, we do install a few smaller unnecessary things on staff VMs already, mostly for admin convenience, such as gist, vim, python, ocflib, and the utils repo. I do get that these are less intrusive than docker, however.

PR #128, opened 2017-02-10
Allow debian-security-support local login

daradib:
This is needed for the dpkg hook installed by the package
of the same name which runs check-support-status as that user:
/usr/share/debian-security-support/check-support-status.hook

Technically the origin is "???" which works in access.conf
but I think allowing all LOCAL logins is more readable.
The user's shell is already /bin/false (overriden with su command).

This should fix these messages during apt/dpkg operations:

    su: Permission denied
    (Ignored)


chriskuehl:
this is great, thanks!

PR #129, opened 2017-02-13
upload-to-box: don't create a new thread every second

kpengboy:
Is there some reason we were doing the old behavior?

### DIFF
--- a/modules/ocf_backups/files/upload-to-box
+++ b/modules/ocf_backups/files/upload-to-box
@@ -218,48 +218,48 @@ def make_box_folder(folder, auth):
         sys.exit(3)
 
 
-def print_progress(multi, stats, prev_total):
+def print_progress(multi, stats):
     """Show an updating progress bar for all ongoing uploads"""
-    width = get_terminal_size((80, 20)).columns  # Default size as a fallback
+    prev_total = 0
+    while True:
+        width = get_terminal_size((80, 20)).columns  # Default size as a fallback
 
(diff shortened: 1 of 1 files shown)


kpengboy:
@jvperrin ^

jvperrin:
Before there was a problem that if it slept at all it would upload very slowly, because it would spend most of its time sleeping and not actually sending data. That's why it's using the timer instead of sleeping.

This change should work, since the progress-printing thread is separate from the data uploading one, but I'd need to try it and see (unless you have already). I don't think it should use a timer though, because if it does, then it will still start a new printing thread every second, which is not desirable. Just calling the `print_progress` function in the new thread should be sufficient.

modules/ocf_backups/files/upload-to-box
```diff
@@ -353,7 +353,9 @@ def main(args):
     # Clear the screen and start printing the progress bar every second
     if not args.quiet:
         subprocess.call(('clear',))
-        print_progress(multi, stats, 0)
+        progress_thread = threading.Timer(1, print_progress, (multi, stats))
```
jvperrin: This is the line I meant, I think this should just be a regular thread, not a timer now.

kpengboy: Timers are merely threads that start after an initial delay.

jvperrin: Ok, but then why have the delay? It's not like it does anything here.

kpengboy: Fair point.

PR #130, opened 2017-02-15
ocf_printhost: redo document processing

matthew-mcallister:
The first major change we've had to the printhost in a while. This branch mainly switches the document preprocessing stage that comes before `enforcer` and the printer itself to use the existing CUPS facilities instead of our half-implementation of the same thing. The filter `ocfps` first passes documents through the user command `pdftops` (which is not part of the CUPS internal filter system), which is a rasterizing converter, then passes them through `pstops` which uses ghostscript to convert to printer-suitable postscript with duplex and economode instructions inserted according to print job options. Other changes are mainly superficial improvements.

This is running on `dev-printhost`. It could use plenty of testing.

TODO (after merging):
- Containerize
- Write documentation


modules/ocf_printhost/files/ocfps
```diff
@@ -0,0 +1,16 @@
...
+if [ "$#" -ne 5 ] && [ "$#" -ne 6 ]; then
+    echo "Usage: $0 job-id user title copies options [file]"
+    exit
+fi
+
+if [ "$#" -eq 5 ]; then
+    $PDFTOPS - - | pstops "$@"
+else
+    $PDFTOPS "$6" - | pstops "${@:1:5}"
+fi
```
chriskuehl: can you explain what is happening here?

matthew-mcallister: Of course. `ocfps` is a CUPS filter, and all CUPS filters have to have the same signature as in the usage string above. If the 6th arg is present, then the filter should take job data (the document to print) from the given file. Otherwise, the filter takes data from stdin. In either case, `ocfps` runs the actual document through `pdftops`, a rasterizing converter provided by the `poppler-utils` package (this is unrelated to the CUPS filter also named `pdftops`). Finally, this modified document is passed to the CUPS filter `pstops` along with the unaltered environment vars and remaining command line arguments provided by cups.

`pstops` is a CUPS filter located in `/usr/lib/cups/filter/`, which is added to the PATH by CUPS. Originally, I specified the full path in a variable `PSTOPS`, but undid that during debugging in case CUPS was adding a different location to the PATH. I can redefine the variable to clarify if that seems useful.

This explanation will be in the documentation.

jvperrin: Maybe include some of that in a comment, or at least the part about the arguments, because that's the important part in this section. Seems better than putting all that in documentation, the overall process and architecture should be in the documentation, but specifics like arguments and stuff seem like they belong here more.

matthew-mcallister: While I agree this stuff should be in a comment on the top, I was planning for our documentation to briefly explain the basics of CUPS architecture as well as our own setup, since the online documentation on CUPS is severely lacking (our setup makes no sense unless you understand CUPS somewhat anyways). I myself believe it would be helpful for OCFers and the general public to state the call signature and location of custom CUPS filters as part of an overview.

On the other hand, it is true that those details are unnecessary to the basic understanding of our setup, since they only matter within this file. There is an argument to be made that OCFers should just read the comments in the script to learn how it's supposed to work.

jvperrin: No, that's fine, it would be good to have documentation on CUPS stuff or anything you had to struggle to find or figure out, that stuff is all good to have, I meant more the specifics that relate to the code itself.

I definitely don't think having either everything in the documentation or everything in the script's comments is ideal though, a mix of the two is probably best, with most stuff in the documentation and anything else in the script perhaps?

matthew-mcallister: Oops, I think I deleted part of my response. I originally said that the stuff in my explanation should definitely be in a comment in the script, but that much of it, particularly the generic stuff, should still be in documentation because it's mostly basic info that anyone working with CUPS should know.

PR #131, opened 2017-02-17
add useful R packages so they don't have to be rebuilt everytime

tahabi:
it takes a lot of time to compile some of these packages from source, like ggplot2. The rest are just useful in general.


chriskuehl:
some of these don't exist on jessie (or only exist in jessie-backports), so you may need to special case it

i think it will install the backport if it's the only available version, can you confirm this works on jessie (try on tsunami or something) as well as on stretch?

matthew-mcallister:
Keep in mind that these packages will get old pretty fast. If you use R's package manager to install new packages, they may pull in newer versions as dependencies and your install times will get slow again (I don't know anything about R though).

tahabi:
I'll try this on tsunami to confirm it works

Also @matthew-mcallister these packages are pretty mature, even if they get out of date it's not going to be too big a deal. ggplot2 only released two updates last year, in May and November. These packages also represent like 95% of what the average person needs out of RStudio (80% of that being ggplot2 alone) so if they want to use the internal package manager that will be a special case.

tahabi:
this doesn't work on jessie because the dependencies for dplyr are outdated, but it works fine in stretch, except for readr, which apparently isn't packaged for debian at all despite being one of the core recommended R packages on tidyverse...

I've removed readr and tested it with RStudio on the desktops, it works.

jvperrin:
Keep in mind `extrapackages` is not only used on desktops, so if some don't work on jessie, then you'll have to put those in an if statement to make sure they are only attempted to be installed when on stretch. I'm assuming you knew that already though since you tested this on tsunami.

tahabi:
Your assumption was generous because I neglected to keep that in mind, thanks
Is there a file for packages that are installed only on the desktops? I don't imagine the login servers need rstudio installed, for example.

jvperrin:
[Yup](https://github.com/ocf/puppet/blob/master/modules/ocf_desktop/manifests/packages.pp), the same file is where [rstudio is installed](https://github.com/ocf/puppet/blob/master/modules/ocf_desktop/manifests/packages.pp#L49), but you should probably put these packages in the top section under their own line with a comment saying they are R packages that go with rstudio.

Baisang:
good choice of packages, that's basically all I used when I took stat133 :)

shex1627:
Is is possible to have just the desktops(maybe staff VMs) have those packages, because we only use R on those places anyway?

jvperrin:
@shex1627 Yes, this is changing the `ocf_desktop` module, so it will only be installed on desktops.

chriskuehl:
I would prefer to only install on stretch instead of only on desktops/staff VMs:

1. staff VMs shouldn't be special cases as much as possible, just empty servers
2. the point of "extrapackages" is to avoid surprises when something is available on desktops but not login servers or death

Is there any reason someone wouldn't use R on a login server?

tahabi:
They could, but it wouldn't be very straightforward to use the graphics libraries through the R console. These are most useful on the desktops because R seems to be built for more interactive use, and if someone is running a R job on their Staff VM they can install the packages they need from inside R itself pretty easily.

matthew-mcallister:
But can't you run R studio from tsunami? I think that would make use of the graphics libraries.

tahabi:
I mean RStudio is an IDE, you could in theory forward it over X, I guess,
but that would be really slow.

matthew-mcallister:
I thought it was a server you access over a web interface?

tahabi:
That's Shiny, which is really an app server for R projects that has a
little interaction built in. Not really like Jupyter where you can run code
remotely, AFAIK.

matthew-mcallister:
No? I'm pretty sure @shex1627 has this running on his staff VM:
https://www.rstudio.com/products/RStudio/#Server

If you can run this on tsunami, then you can ssh forward it and get graphics in your browser.

tahabi:
I don't think any edition of RStudio is even packaged for Stretch - we do
it manually in `ocf_desktops` IIRC.

matthew-mcallister:
So? You can compile it and run it from your home directory. Or we can package it and install it on tsunami like we do for desktops.

tahabi:
Seems like a lot of additional effort for an impractical use case. Much
simpler to just install these packages on desktops where they're much more
likely to be used rather than trying to install them everywhere just
because.

chriskuehl:
It doesn't seem like any additional effort? It's just where you drop the install line in Puppet (extrapackages instead of ocf_desktop).

tahabi:
I mean repackaging RStudio and RStudio Server to run on the login servers

matthew-mcallister:
We already package RStudio, but we wouldn't want to install it on the login
server because it's a desktop application. We would only have  to package
RStudio server.

chriskuehl:
Why would you need to repackage it?

matthew-mcallister:
I addressed the outstanding concerns and merged this in with #157.

PR #132, opened 2017-02-17
add tvhours shorturl

tahabi:


matthew-mcallister:
How about instead do ocf.io/tv -> ocf.b.e/tv

jvperrin:
I agree with @matthew-mcallister, it would make it possible to do `ocf.io/tv/hours` then instead to get to the same page (which I think I'd remember more, since it's the same as the main page URL), and just make it easier for any future TV pages.

tahabi:
oops

PR #134, opened 2017-02-19
ocf_mysql: Listen on all interfaces and on IPv6 (rt#5935)

jvperrin:
I also removed all the chaining arrows from `ocf_mysql/init.pp`, because they are confusing and we use them very rarely. There's a few others scattered around that should probably be removed, although obviously not if there are objections :P

### DIFF
--- a/modules/ocf_mysql/files/99-ocf.cnf
+++ b/modules/ocf_mysql/files/99-ocf.cnf
@@ -1,5 +1,5 @@
 [mysqld]
-bind-address = 0.0.0.0
+bind-address = ::
 
 # https://mariadb.com/kb/en/mariadb/optimizing-key_buffer_size/
 key_buffer_size = 128M
--- a/modules/ocf_mysql/manifests/init.pp
+++ b/modules/ocf_mysql/manifests/init.pp
@@ -17,19 +17,8 @@ class ocf_mysql {
     require        => File['/opt/share/puppet/mariadb-server-10.0.preseed'],
   }
 
-  # Listen on all interfaces and on both IPv4/IPv6
-  augeas { '/etc/mysql/mariadb.conf.d/50-server.cnf':
-    lens    => 'MySQL.lns',
-    incl    => '/etc/mysql/mariadb.conf.d/50-server.cnf',
-    changes => [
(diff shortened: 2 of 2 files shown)


chriskuehl:
The chaining operators grow on you :P

I didn't like them at all at first either, but we use them almost exclusively (at yelp), and I've started to find them more readable. Admittedly it's not at all clear what `->` or `~>` do, but after figuring that out, it sort of "linearizes" the dependency graph so that you can read the order things are happening from top-to-bottom, almost like a dockerfile. Contrast that with `notify`/`subscribe`/`require`, where the dependencies and notifications can be anywhere in the file, and so you might have to jump around to understand the order in which things are happening.

Anyway, I'm happy with removing them here as well. They are indeed confusing to newcomers even though I think maybe overall they might improve the readability (and that is probably questionable).

chriskuehl:
I notice we already set the bind-address in 99ocf.cnf and that it actually conflicts with the one we set here: https://github.com/ocf/puppet/blob/master/modules/ocf_mysql/files/99ocf.cnf

Presumably the ordering changed in stretch so that 99ocf.cnf isn't the last thing loaded, maybe it'd be better to fix that instead of using augeas on the dist-provided files? In any case we should delete the `bind-address` there.

jvperrin:
Ah, good point, it does override earlier statements with later ones, so that makes sense, not sure why it would successfully be listening on IPv6 right now though if it did that, since the later address is IPv4 only? Anyway, good catch on that

chriskuehl:
`mariadb.conf.d` gets included last:

```
# Import all .cnf files from configuration directory
!includedir /etc/mysql/conf.d/
!includedir /etc/mysql/mariadb.conf.d/
```

Probably we can just move `99ocf` into that directory.

jvperrin:
@chriskuehl Yeah, that's a better way to do things, I went ahead and did that and it works.

jvperrin:
This doesn't appear to actually work, I set up dev-maelstrom to restore a group's database, and when it was set up, it didn't listen on the correct ports unless `bind-address` was changed in the file provided by mariadb.

PR #136, opened 2017-02-20
Puppet ldap better (and upgrade firestorm to stretch)

jvperrin:
This isn't quite done, it needs some changes still since the config location has changed and all that, but there's some changes here so far.


modules/ocf_ldap/templates/slapd.conf.erb
```diff
@@ -28,18 +26,13 @@ authz-regexp ^uid=([^,/]+),cn=OCF.BERKELEY.EDU,cn=GSSAPI,cn=auth$ uid=$1,ou=Peop
 authz-regexp ^uid=host/([^,/]+)\.ocf\.berkeley\.edu,cn=OCF.BERKELEY.EDU,cn=GSSAPI,cn=auth$ cn=$1,ou=Hosts,dc=OCF,dc=Berkeley,dc=EDU
 
 ## Database general configuration
-moduleload   back_bdb
-sizelimit    unlimited
```
jvperrin: Haha, looks like I removed this before when I should have kept it, this was what caused those errors on anthrax with fetching sorried users. Whoops...

jvperrin:
None of the changes in `slapd.conf` have any effect right now, they are all stored in `slapd.d` now in LDIF format, which isn't ideal unless they can be put there with Puppet

jvperrin:
Also would be nice if ldap-git-backup actually backed up the new config location...

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=721155

jvperrin:
Ok, I think this is ready to merge now. It's not great, because the config changes are stored in LDAP, but they are backed up a couple times, once by `ldap-git-backup`, then again with `ocf_backups` when it backs up `/etc/ldap/slapd.d`, so I think this should be good. If anything breaks. It should always be possible to get it working again by just editing the slapd.d config files (because the config in LDAP can't be edited when it isn't able to start because of a config issue), so that's still good, but storing configuration in Puppet doesn't feel great...

Thoughts?

modules/ocf_kerberos/manifests/init.pp
```diff
@@ -28,6 +33,22 @@
...
+      'set service[last()]/socket stream',
+      'set service[last()]/protocol tcp6',
+      'set service[last()]/wait nowait',
+      'set service[last()]/user root',
+      'set service[last()]/command /usr/sbin/tcpd',
+      'set service[last()]/arguments/1 /usr/lib/heimdal-servers/kadmind',
+    ],
+    onlyif  => "match service[. = 'kerberos-adm'] size == 1",
+    require => Package['heimdal-kdc'],
+    notify  => Service['inetd'],
```
chriskuehl: lol how many tries did this take to get right

modules/ocf_ldap/manifests/init.pp
```diff
@@ -1,33 +1,17 @@
...
   service { 'slapd':
-    subscribe => File[
-      '/etc/ldap/slapd.conf',
-      '/etc/ldap/schema/ocf.schema',
-      '/etc/ldap/schema/puppet.schema',
-      '/etc/default/slapd',
-      '/etc/ldap/sasl2/slapd.conf'];
+    subscribe => [
+      File['/etc/ldap/krb5.keytab',
+        '/etc/ldap/sasl2/slapd.conf'],
```
chriskuehl: is this weird indentation intentional?

modules/ocf_kerberos/manifests/init.pp
chriskuehl: (i meant the entire augeas block, not the 4 lines github chose to highlight)

jvperrin: Haha, far too long, augeas always seems to work that way...

modules/ocf_ldap/manifests/init.pp
jvperrin: Yes, mostly because it subscribes to both `File` and `Augeas`, maybe that could be made more clear by making the first file string on a new line?

PR #138, opened 2017-02-25
Install dirmngr first in setup

kpengboy:
This way, on stretch hosts, we can add apt keys for third-party
repositories before installing packages on a host for the first time.

Comments?

### DIFF
--- a/modules/ocf/manifests/packages.pp
+++ b/modules/ocf/manifests/packages.pp
@@ -16,10 +16,6 @@ class ocf::packages {
   include ocf::packages::shell
   include ocf::packages::ssh
 
-  class { 'ocf::packages::gnupg':
-    stage => first,
-  }
-
   # Packages to automatically update to be the latest version. This should be
   # kept short, since apt-dater should be used to update almost all packages.
   package {
--- a/modules/ocf/manifests/packages.pp
+++ b/modules/ocf/manifests/packages.pp
@@ -16,6 +16,10 @@ class ocf::packages {
   include ocf::packages::shell
   include ocf::packages::ssh
 
+  class { 'ocf::packages::gnupg':
+    stage => first,
+  }
+
   # Packages to automatically update to be the latest version. This should be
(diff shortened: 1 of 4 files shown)


modules/ocf/manifests/packages.pp
```diff
@@ -16,6 +16,10 @@
   include ocf::packages::shell
   include ocf::packages::ssh
 
+  class { 'ocf::packages::gnupg':
+    stage => first,
+  }
+
```
chriskuehl: I'd probably move this to the manifest that needs it, e.g. `ocf::apt`` probably? (granted, there are other `apt::key`s, but that's the main place)

There are definitely exceptions but mostly this list is "general" packages rather than things required by other manifests.

modules/ocf/manifests/packages/gnupg.pp
```diff
@@ -0,0 +1,5 @@
+class ocf::packages::gnupg {
+  if $::lsbdistcodename != 'jessie' {
+    package { 'dirmngr':; }
+  }
+}
```
chriskuehl: seems fine to me, but perhaps leave a comment about why this is special or the reason will be forgotten eventually

jvperrin: Agreed, a comment here (and perhaps in `site.pp`) would help to make it clearer why we need this package specifically.

modules/ocf/manifests/packages.pp
kpengboy: I guess that's fair, considering that `ocf::apt` is one of the common classes included. I've reworked some changes and I'll test them in the next few days when I have time.

kpengboy:
I made some changes. Now the code's inside `ocf::apt` (which is included as a common class), and it seems to work on a test hozer I spun up and auto-installed docker on. Comments? Does anyone object to the use of resource collectors?

kpengboy:
Squashing and merging.

PR #142, opened 2017-03-02
Remove the stupid Debian vim config

chriskuehl:
Normally I do not like overriding these things, but the new Debian config is completely unusable. I can't even copy-paste text to my system clipboard because the idiotic mouse mode takes over.

I also added [vim-sensible](https://github.com/tpope/vim-sensible) to our default config, which seems reasonable since lots of our users might occasionally use vim, but most won't take the time to install their vimrc (or even have one to install).

I don't really care about that change though, so if people object, I'll delete all the vim-sensible stuff and we can just keep the first 2 lines.


NickImpicciche:
+inf

PR #143, opened 2017-03-04
auto log out for staff

NickImpicciche:


modules/ocf_desktop/files/xsession/Xsession
```diff
@@ -114,7 +114,10 @@ xset m 0 0
 sleep 8 && /usr/local/bin/fix-audio &
 
 # Auto logout inactive users
-/usr/bin/xautolock -locker /usr/local/bin/auto-lock -time 7 -noclose &
+# killer is called killtime minutes after locker is called
+# any user input including shaking the mouse then not clearing the lockscreen is considered user input
```
jvperrin: So if anyone moves the mouse to see if they are logged in, and then leaves it then the timer resets? If so, the timer on this should be much much shorter than 90 minutes total, because that will happen a lot.

Maybe a different method should be used if this is what it does, like a background script that logs the user out after a certain amount of inactivity? Seems like that would be pretty simple at least, and could be forked off `auto-lock` once it actually locks the screen.

modules/ocf_desktop/files/xsession/Xsession
```diff
@@ -114,7 +114,10 @@ xset m 0 0
...
+# locker must still be running for killer to be called, so if auto-lock is killed then killer will not be called
+/usr/bin/xautolock -locker /usr/local/bin/auto-lock -time 7 -killtime 83 -killer 'pkill xfce4-session' -noclose &
```
jvperrin: I know `83` was chosen to make it add up to `90`, but I think that needs a comment, because it seriously took me like 30 seconds to realize why that was the case.

modules/ocf_desktop/files/xsession/auto-lock
```diff
@@ -139,6 +139,9 @@ def main():
         sys.exit(1)  # Because we didn't lock/logout
     else:
         timebomb_explode()
+        # since xautolock killer will only be called while this script is running we sleep to
+        # keep it in the background
+        time.sleep(86400)
```
jvperrin: Meh, if this is what it needs then I suppose this is fine.

modules/ocf_desktop/files/xsession/Xsession
NickImpicciche: I did comment that already, "killer is called killtime minutes after locker is called".

modules/ocf_desktop/files/xsession/Xsession
NickImpicciche: I've thought about this, not sure what the best way to do it would be. This method definitely has problems but seems to be the most natural implementation using xautolock.

NickImpicciche: Also while 90 minutes would miss a lot of cases, it covers more than what we do now, and we can always make this longer/shorter depending on experience with it.

jvperrin: I just think that 90 minutes would miss pretty much all the cases we care about, like when the lab is open, and would only cover cases where OCF staff stay logged in overnight or something.

modules/ocf_desktop/files/xsession/auto-lock
NickImpicciche: Yeah, not a huge of this either. This is the case is that normally you put something like xscreensaver here, but we use our script which calls xflock4. However this method definitely does not work without this.

modules/ocf_desktop/files/xsession/Xsession
jvperrin: I think this could just be made clearer, or `killtime` could be a round number, because my first inclination was to try to look up if `83` corresponded to some special code that `xautolock` needed. Not a big deal though, just a little more confusing than it could be.

modules/ocf_desktop/files/xsession/Xsession
NickImpicciche: yeah I'm not opposed to lowering it. I just picked a number, what do you think would be a good number? I do not have strong feelings towards 90 at all.

modules/ocf_desktop/files/xsession/Xsession
NickImpicciche: What way of explaining it do you think would make more sense?

modules/ocf_desktop/files/xsession/Xsession
jvperrin: I'm not sure, but I feel like 30 minutes might be better, although I think it'll run into the same problem where people would wiggle the mouse often enough to make it last forever essentially, at least when the lab is open...

NickImpicciche: Possibly, but we can always set it to that and then if we're not happy with it then we can change it later

modules/ocf_desktop/files/xsession/Xsession
jvperrin: Not sure, maybe an extra line that just says "time and killtime are both in minutes" maybe, or something about them adding to up a round number?

modules/ocf_desktop/files/xsession/Xsession
jvperrin: Alright, fair enough, good idea to try it first.

matthew-mcallister:
Will this still kill sessions which were manually locked? If not, I think this won't be as helpful as we need it to be overall.

NickImpicciche:
Yes it will, but once again any input while locked will reset the timer

matthew-mcallister:
OK, I can accept this limitation; it's better than nothing anyways.

jvperrin:
Will it though, doesn't that just call `xflock4` as a script separately from `xautolock`?

I'm beginning to think we need some kind of background listener that checks if `xflock4` is running for some number of consecutive checks and kills the session if so. Seems messy but that would fix the issue with the mouse movement too...

matthew-mcallister:
> Will it though, doesn't that just call xflock4 as a script separately from xautolock?

Yes, but if xautolock keeps running anyways, it will still eventually kill the session if there's no activity.

NickImpicciche:
I can check again to confirm, but I am almost confident it will. Actually right now xautolock actually runs the notifier again while the screen is locked if you move the mouse but do not unlock leading to a notifier that says "you have been inactive" on the lock screen

jvperrin:
Ok, sounds fine

NickImpicciche:
I agree something more complicated may have more ideal functionality, but this wasn't very hard to do, so I figured it could do for now

jvperrin:
:+1:

NickImpicciche:
nvm, this actually does seem to break it. Not sure why...

PR #144, opened 2017-03-04
staff auto logout

NickImpicciche:


chriskuehl:
Maybe I'm being optimistic, but isn't this a problem that can be solved with non-technical means?

modules/ocf_desktop/files/xsession/staff-logout
```diff
@@ -0,0 +1,33 @@
...
+from subprocess import Popen
+from time import sleep
+
+from ocflib.lab.hours import Day
+
+# delay before killing session in minutes
+delay = 45
+# extra delay if we will be closed delay minutes from now
+bonus = 30
+now = datetime.datetime.now()
```
chriskuehl: this code all needs to be guarded inside an `if __name__ == '__main__'` so that the file can be imported without side effects

modules/ocf_desktop/files/xsession/staff-logout
```diff
@@ -0,0 +1,33 @@
...
+if not Day.from_date(now).is_open(now + datetime.timedelta(minutes=delay)):
+    delay += bonus
+
+sleep(60 * delay)
+
+
+cmd = 'xscreensaver-command -display :0.0 -time'.split()
+p = Popen(cmd, stdout=PIPE)
```
chriskuehl: don't split the string like that :(

if you really insist on splitting, do it properly with `shlex.split` which understands shell syntax. preferably just write out the arguments:
```python
proc = subprocess.Popen(
    ('xscreensaver-command', '-display', ':0.0', '-time'),
    stdout=subprocess.PIPE,
)
```

however you don't need Popen here at all

```python
output = subprocess.check_output(('xscreensaver-command', '-display', ':0.0', '-time'))
```

modules/ocf_desktop/files/xsession/staff-logout
```diff
@@ -0,0 +1,33 @@
...
+output = p.communicate()[0].decode('utf-8').strip().split()
+
+if output[3] == 'locked':
```
chriskuehl: you can avoid decoding which assumes the encoding and throws exceptions if it isn't
```python
if output.split()[3] == b'locked':
```

modules/ocf_desktop/files/xsession/staff-logout
```diff
@@ -0,0 +1,33 @@
...
+    # format is day, hour:minute:second year
+    time_of_lock = datetime.datetime.strptime(' '.join(output[-4:]), '%b %d %H:%M:%S %Y')
+    now = datetime.datetime.now()
+    diff = now - time_of_lock
+    # -1 prevents any slight misses
+    if diff >= datetime.timedelta(minutes=(delay - 1)):
+        cmd = 'pkill xfce4-session'.split()
+        call(cmd)
```
chriskuehl:
```python
subprocess.call(('pkill', 'xfce4-session'))
```

modules/ocf_desktop/files/xsession/xflock4
```diff
@@ -0,0 +1,52 @@
...
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+#
```
chriskuehl: given that we can assume the locker, can't this entire script be replaced with
```bash
#!/bin/bash -eu
/usr/local/bin/staff-logout &
exec xscreensaver-command -lock
```

modules/ocf_desktop/files/xsession/xflock4
```diff
@@ -0,0 +1,52 @@
...
+
+# Lock by xscreensaver or gnome-screensaver, if a respective daemon is running
+for lock_cmd in \
+    "xscreensaver-command -lock" \
+    "light-locker-command --lock" \
+    "gnome-screensaver-command --lock"
+do
+    /usr/local/bin/staff-logout &
+    $lock_cmd >/dev/null 2>&1 && exit
+done
```
chriskuehl: doesn't this launch the logout process 3 times?

matthew-mcallister:
Why would we want to check in xflock4 when it's already installed?

modules/ocf_desktop/files/xsession/xflock4
matthew-mcallister: It works no matter what you have installed. It exits after the first command that succeeds.

matthew-mcallister:
> Maybe I'm being optimistic, but isn't this a problem that can be solved with non-technical means?

Sure, but so can users leaving themselves logged in, in principle, yet I still think auto logout was a worthwhile idea (even if our script is too complicated). I'm pretty sure with there will always be times when staff accidentally leave computers locked all day. This happened even back when we were in Hearst.

modules/ocf_desktop/files/xsession/xflock4
chriskuehl: no it does not. if the only one you have installed is gnome-screensaver, it will launch 3 copies of our script in the background.

NickImpicciche:
>Maybe I'm being optimistic, but isn't this a problem that can be solved with non-technical means

I wouldn't say it's a huge problem or anything, but evidence seems to say otherwise

modules/ocf_desktop/files/xsession/staff-logout
```diff
@@ -0,0 +1,33 @@
...
+from subprocess import PIPE
+from subprocess import Popen
+from time import sleep
+
+from ocflib.lab.hours import Day
+
+# delay before killing session in minutes
+delay = 45
+# extra delay if we will be closed delay minutes from now
+bonus = 30
```
kpengboy: Do we really need to kick staff off computers when we're closed?
IMO if we're closed, we should let them be logged in until ~20 minutes after the lab reopens.

kpengboy:
One thing: this doesn't prevent people from manually running `xflock4` and keeping computers locked indefinitely

NickImpicciche:
Yes it does because this replaces xflock4. They could run
xscreensaver-command -lock manually, but I don't see this as an issue.

modules/ocf_desktop/files/xsession/xflock4
matthew-mcallister: Oh, ha, I didn't even realize @nickimp added a line to this file. It looked exactly like you were questioning whether xflock4 works. Seriously, why are we checking this in?

modules/ocf_desktop/files/xsession/staff-logout
```diff
@@ -0,0 +1,33 @@
...
+now = datetime.datetime.now()
+if not Day.from_date(now).is_open(now + datetime.timedelta(minutes=delay)):
```
kpengboy: This won't work if `now + datetime.timedelta(minutes=delay)` happens to be on the next day. I think you meant `Day.from_date(now + datetime.timedelta(minutes=delay)).is_open(now + datetime.timedelta(minutes=delay))` (of course you can simplify that expression)

modules/ocf_desktop/files/xsession/staff-logout
kpengboy: That backfires with the next line of code, though...

chriskuehl: true

modules/ocf_desktop/files/xsession/staff-logout
NickImpicciche: It won't error, but will just always return that it is closed in my understanding. I've fixed anyway as this is more reasonable.

modules/ocf_desktop/files/xsession/staff-logout
NickImpicciche: Not really sure why you'd want to import this file, but sure

modules/ocf_desktop/files/xsession/staff-logout
NickImpicciche: I think even if we're closed logging out after a reasonable amount of time makes sense. However I see your point and I will update to slightly lower delay and greatly increase bonus.

NickImpicciche:
>Why would we want to check in xflock4 when it's already installed

To change its behavior to run our script. At first I just added in the part we needed keeping the rest of it the same, but now its very simplified.

modules/ocf_desktop/files/xsession/staff-logout
kpengboy: rip, I sometimes liked to lock the screen, go away and do stuff for a long while, and come back and resume my work...

Or sometimes, I liked to do all nighters in the lab, fall asleep, wake up again and resume my work...

NickImpicciche: okay well if you know you're explicitly going to do that then you can run xscreensaver-command -lock.

kpengboy: Well, people don't always know when they are going to collapse of exhaustion.....

tahabi:
I just kill xautolock (thanks mcint for showing me that trick) when I don't
want it to close on me randomly, you could probably do something similar
here if you're working/reading/watching a video and don't want to keep
touching the mouse to stop the screen from suspending or killing your
session.

On Mar 4, 2017 8:03 PM, "kpengboy" <notifications@github.com> wrote:

*@kpengboy* commented on this pull request.

In modules/ocf_desktop/files/xsession/staff-logout
<https://github.com/ocf/puppet/pull/144#discussion_r104302463>:

> @@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+import datetime
+from subprocess import call
+from subprocess import PIPE
+from subprocess import Popen
+from time import sleep
+
+from ocflib.lab.hours import Day
+
+# delay before killing session in minutes
+delay = 45
+# extra delay if we will be closed delay minutes from now
+bonus = 30

rip, I sometimes liked to lock the screen, go away and do stuff for a long
while, and come back and resume my work...

Or sometimes, I liked to do all nighters in the lab, fall asleep, wake up
again and resume my work...

—
You are receiving this because you are subscribed to this thread.
<https://github.com/ocf/puppet/pull/144#discussion_r104302463>, or mute the
thread

modules/ocf_desktop/files/xsession/staff-logout
NickImpicciche: I think you may have larger issues than your session being killed then

NickImpicciche:
Any more comments? Requested changes?

chriskuehl:
I sort of agree with Kevin. Besides making it harder to cheat on lab hours, what's the benefit of kicking out staff when the lab is closed? Seems like it's only going to annoy our own people.

NickImpicciche:
My main annoyance is when people leave specific computers I want to use
(dual monitors, eruption, standing desk) locked for hours without being
there, but if I am outnumbered I will change behavior to Kevin's suggestion.

chriskuehl:
I don't actually care anymore :) It probably would've annoyed me at some point, though.

NickImpicciche:
I'm going to bring up at bod to decide on desired behavior, so if there are
no more requested changes I will merge when ready

modules/ocf_desktop/manifests/xsession.pp
```diff
@@ -162,6 +162,7 @@
     owner   => root,
     group   => ocfstaff,
     mode    => '0754',
+    source  => 'puppet:///modules/ocf_desktop/xsession/xflock4',
```
matthew-mcallister:
```
    require => Package['xfce4-session'],
```

mcint:
Let's implement the staff auto-logout (in off hours) on just dual-monitor computers.
If we want to address excessive locking on certain computers, target them directly.

NickImpicciche:
This seems like an unnecessary addition, I don't agree

chriskuehl:
something really wrong has happened to this git history

NickImpicciche:
I'm going to fix it

PR #145, opened 2017-03-07
update u2f/fido/hardware 2fa udev rules

tahabi:
it works now

### DIFF
--- a/modules/ocf_desktop/files/70-u2f-fido.rules
+++ b/modules/ocf_desktop/files/70-u2f-fido.rules
@@ -1,14 +1,43 @@
-# From yubikey:
-# https://www.yubico.com/faq/enable-u2f-linux/
-# https://github.com/Yubico/libu2f-host/blob/master/70-u2f.rules
+# Copyright (C) 2013-2015 Yubico AB
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1, or (at your option)
+# any later version.
+#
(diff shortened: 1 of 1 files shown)


tahabi:
I'm not entirely sure this is necessary because we also have libu2f-host0 installed but it doesn't seem to hurt

tahabi:
Oh this is just copied directly from the yubikey example config file. I
just `curl -o`'d it.

jvperrin:
Yeah, I just think it would help to include a link to where it is from so if it needs updating it can be more easily done.

tahabi:
Doesn't the terms of the license say we must include that disclaimer, especially since our code is licensed under GPLv3 and not 2.1, even though v3 includes the same disclaimer? I'm removing it per your recommendation, but I'm not sure. Something to ask @kpengboy perhaps?

jvperrin:
Hmm, that's a good point, I guess it's fine to include then. I'd just prefer not to have extra license comments if we can link to the original and essentially say it's not ours in the first place, but the license is probably necessary?

Honestly I don't know much about all this licensing stuff though.

tahabi:
According to [this](https://www.gnu.org/licenses/gpl-faq.en.html#v2v3Compatibility) we can include the code because it says GPVv2.1 or later, but I'm not sure if that means we still have to include the preamble. Honestly this tiny set of rules isn't even worth the time dedicated to figuring this licensing stuff out, lol.

jvperrin:
Wait, so what is going on with this, is there a package that fixes it that could be used instead, or does it just work now?

PR #148, opened 2017-03-17
add gdrive and absa shorturls

tahabi:
absa link is edit iff in ocf.berkeley.edu


jvperrin:
Google Drive sounds good, but why make one for ABSA when it'll be turned in soon anyway and then become obsolete until next year? I feel like short URLs work better for long-term things, like GitHub links, shortcuts to things in the docs, blog posts that are going to stay around and are published offline so that people can more easily type the links, things like that.

tahabi:
useful for reference and future use, in the same vein as putting the stf-cost-breakdown and senate-resolution links.

jvperrin:
I mean those were more so that we could publish them easily in applications rather than have them easily available for the future. Although, having ABSA easily available might be nice, especially when doing it next year, so maybe having the link and updating it for that would be ok?

tahabi:
The link is to the ABSA folder, which I imagine we'd want to keep using for the foreseeable future, unless I'm misinterpreting what you're saying

jvperrin:
Oh ok, sorry, I should have actually tried the link, I thought it was going to the current ABSA document, not the folder. Seems fine to me then, since it shouldn't change then.

PR #149, opened 2017-03-17
redirect for youtube

Baisang:
since it'll likely be a bit until we can get a custom url for youtube


NickImpicciche:
I would add /yt as well, fits with /github and /gh

PR #152, opened 2017-03-26
ocf: add iptables rules for forwarding packets on bridge

kpengboy:


chriskuehl:
Did you consider the security implications of this? Correct me if I'm wrong, but it appears to me that your patch is taking the worst of both worlds (effectively allowing all FORWARD traffic, but still setting up complicated rules based on Docker changing the policy to DROP). It is very easy to verify that these rules do not add security via the repro steps in the issue (it took me about 30 seconds to test them).

Instead I think we should either:

1. **If we don't care about the security implications and are okay with the pre-1.13 behavior:**
  
     We can tell Docker to shove off and not change the policy by turning on ip_forward before starting Docker. I tested that this works on riptide. This is a much simpler change than your iptables one and doesn't change any actual settings (Docker normally turns on ip_forward anyway).

    I wish there was a more explicit way to do this, though.

2. **If we do care about the security implications:**
  
      I think we should stop depending on quirky Docker behavior that could change at any time and instead set the policy to DROP ourselves, and then explicitly add rules to allow VM traffic without allowing arbitrary traffic to the containers.

Personally, I'm convinced based on the issue (https://github.com/docker/docker/issues/14041) that we *do* either want a DROP policy or a rule to specifically drop incoming packets headed to containers. What do you think?

I don't think this is a good approach to using the firewall module, but given my thing is sort-of vaporware until I get it into a presentable state and share it, I can't justify blocking you.

kpengboy:
I've tightened the rules. Further review would be welcome.

modules/ocf/manifests/firewall.pp
```diff
@@ -0,0 +1,15 @@
...
+  # calls iptables-save, which saves all iptables rules when it runs,
+  # including any transient rules added by Docker that happen to be
+  # present. These transient rules then get saved and loaded at every boot.
+  # Docker is designed to clear away or work with most existing iptables
+  # rules. However, it doesn't clear POSTROUTING rules as we'd like it to,
+  # so we have to manually ensure those don't get persisted.
+  Firewall <| |> -> exec { 'Delete Docker POSTROUTING rules in iptables persistent config':
+    command => "sed -i '/^-A POSTROUTING / d' /etc/iptables/rules.v4",
+    onlyif  => "grep '^-A POSTROUTING ' /etc/iptables/rules.v4",
+  }
```
chriskuehl: I really don't like the sed-ing out here and would prefer you to not do it and just live with the fact that Puppet will save the rules, knowing that eventually we'll fix that. But it's not a big deal so I leave it up to you.

chriskuehl:
rules look good, thanks for that

modules/ocf/manifests/firewall.pp
kpengboy: I don't like this very much either, but neither do I feel comfortable with random POSTROUTING rules being accidentally persisted. If anyone has a concrete better proposal in the future, please suggest it.

PR #154, opened 2017-03-28
Run puppet uniformly as cronjob and update puppet-trigger

kpengboy:
This configures puppet to be run as a cronjob on all hosts,
including jessie hosts. It also modifies `puppet-trigger` in light of
this to trigger a single puppet run synchronously. The
`puppet-trigger -f` functionality is removed, and `--debug` no longer
requires `--test`.

### DIFF
--- a/modules/ocf/files/puppet-trigger
+++ b/modules/ocf/files/puppet-trigger
@@ -3,7 +3,6 @@
 import argparse
 import os
 import re
-import signal
 import socket
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


modules/ocf/files/puppet-trigger
```diff
@@ -60,12 +34,12 @@ def switch_to_environment(env):
...
     # TODO: use subprocess.run on python3.5; we don't actually care about stdout here
     subprocess.check_output(
-        (
+        [
             'kinit',
             '-t', '/etc/krb5.keytab',
             'host/{}.ocf.berkeley.edu@OCF.BERKELEY.EDU'.format(hostname),
             'ldapmodify',
-        ),
+        ],
```
jvperrin: ?

kpengboy: For consistency with the `os.execvp` above

jvperrin: Sure, I mean I don't think that matters too much. Anyway, it's not a big deal either way if this is a tuple or list to my mind.

chriskuehl: consistency with the os.execvp which you unnecessarily changed :)

kpengboy: Actually, the way we build the list of arguments in this script now, it does make a little more sense for it to be a list...

PR #158, opened 2017-04-10
Update docker GC cronjobs

kpengboy:
Now automatically removes all images older than 4 weeks on hosts other
than Jenkins, which gets a daily prune instead.

Currently blocked on awaiting Docker >=17.04 to be on all machines.


kpengboy:
Rebasing.

PR #159, opened 2017-04-12
Add staff to docker on staffvms

kpengboy:
I never felt too comfortable with giving every staffer root on every other staffer's VM. This change aims to preserve the intent of 29b8882 while narrowing the grant of root privileges.

### DIFF
--- a/hieradata/type/staffvm.yaml
+++ b/hieradata/type/staffvm.yaml
@@ -9,5 +9,5 @@ ocf::auth::nopasswd: true
 # let any staff use docker on staff VMs without sudo
 classes:
     - ocf_staffvm::docker_group
-# temporary
+# TODO: temporary
 ocf::packages::docker::admin_group: docker
--- a/hieradata/type/staffvm.yaml
+++ b/hieradata/type/staffvm.yaml
@@ -7,4 +7,7 @@ ocf::auth::nopasswd: true
 
 
 # let any staff use docker on staff VMs without sudo
-ocf::packages::docker::admin_group: ocfstaff
+classes:
+    - ocf_staffvm::docker_group
+# temporary
+ocf::packages::docker::admin_group: docker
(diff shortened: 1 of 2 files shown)


chriskuehl:
I think you are doing this at the wrong layer. A manifest for installing a package shouldn't know or care what type of host it is or have special rules for staff VMs. Compare for example with `ocf::auth` which doesn't know anything about staff VMs, but which has a flexible config (we've gone through many iterations of the rules without changing the class) and can be made to work for staff VMs via hiera.

I suggest instead to either:

* Factor out this special behavior into another class (e.g. `ocf_staffvm::docker_group` or something).
* Add another parameter to `ocf::packages::docker` like `admin_users` or something, where each user in that list gets added to the group. I think this is somewhat dangerous though (e.g. if the admin group is `ocfroot`, you wouldn't want puppet trying to add users to that group).

kpengboy:
What do you think of adding a parameter to specify which users are to be added to the `docker` group, without regard to what the actual docker socket group is? Granted, it wouldn't make sense to set both this parameter and the `admin_group` parameter, but at least it doesn't create a special class just for this...

chriskuehl:
Adding a parameter sounds which does that sounds preferable to this branch as-is.

Though it's not clear to me why you prefer to have one large class with multiple responsibilities as opposed to several small classes with single purposes.

hieradata/type/staffvm.yaml
```diff
@@ -7,4 +7,7 @@ ocf::auth::nopasswd: true
 
 
 # let any staff use docker on staff VMs without sudo
-ocf::packages::docker::admin_group: ocfstaff
+classes:
+    - ocf_staffvm::docker_group
+# temporary
```
chriskuehl: can you write TODO like we usually do so it's easier to grep for?

PR #160, opened 2017-04-17
Defaults sharing of sudo-right between terminals

mcint:


tahabi:
+1

chriskuehl:
I don't know much about tty tickets, but I assume it's defaulted to off for a reason? Do we know what that reason is?

kpengboy:
From the Arch wiki:

> **Warning:** This will let any process use your sudo session.

chriskuehl:
Does that mean any process running as your user? If so I'm not overly concerned about it (they could also just write malicious stuff into your .bashrc for the same effect).

If it's all processes then it's obviously an issue, but I can't see how that would be the case.

kpengboy:
Then again, if they wrote malicious stuff in your bashrc, they still
haven't gotten root...

chriskuehl:
That is false for all useful intents and purposes.

mcint:
Apple defaults it disabled. I think it's just a defense in-depth idea. If you run semi-trusted scripts from elsewhere...they could call sudo without you realizing, and run with your privileges. If you're concerned about staffers running `curl ... | sh`, then maybe this should move to a different module, but I don't really see the point.

I remember that finding some very interesting webpage related to this just before I submitted the request...I may have just been absorbing the best practices suggested at the foot of the arch wiki.

chriskuehl:
If that's the only security concern, then I'm +1 for shipping this.

Root staffers should understand that anything they run as their user effectively has the ability to escalate to root. Whether that's by calling sudo directly or by more creative means (like adding stuff to your PATH, writing to bashrc, etc.), I don't think there is a meaningful security distinction there. Similar argument for why having passwordless sudo isn't nearly as bad as people sometimes think it is.

jvperrin:
There do seem to be a lot of people upset about this being the default in Mac OS X for a while, but I don't see it as a problem really since there's still a time-based cache, so you'd have to reauthenticate every so often anyway. I think the main concern people have about this is where you are logged into multiple ttys and maybe remotely execute something with `sudo` and then someone locally can use it without a password, but I agree once someone has access to your user account, then this is a really small concern compared to all the other things they can do anyway.

jvperrin:
Thoughts on this? I'm in favor of just merging it, seems reasonable and nobody is especially opposed to it. Speak now or forever hold your peace (until you want to revert it)! :P

jvperrin:
(bumping again)

chriskuehl:
+1 from you, me, abizer, and mcint. Who are you waiting for? :P

jvperrin:
lol ok, I didn't want to merge it because I wanted to let @mcint do it, but I forgot that @mcint can't merge in this repo, so fair enough...

kpengboy:
Interesting, did we forget to add him to the admin team?

PR #161, opened 2017-04-21
add more irc -> slack channels

gnowxilef:


jvperrin:
Sure, sounds fine, eventually this'll be replaced with something that automatically mirrors all channels, but until then this looks good!

PR #162, opened 2017-04-26
ocf: don't install AUFS

kpengboy:
Docker 1.13+ falls back to overlay2 if AUFS is not installed, but the Debian package installs AUFS as a recommends. Rather than unnecessarily installing AUFS and then manually configuring docker to use overlay2, let's not install AUFS to begin with, and let Docker naturally fall back to overlay2.

### DIFF
--- a/modules/ocf/manifests/packages/docker.pp
+++ b/modules/ocf/manifests/packages/docker.pp
@@ -16,8 +16,20 @@ class ocf::packages::docker($admin_group = undef) {
   package {
     'docker.io':
       ensure  => purged;
+  }
+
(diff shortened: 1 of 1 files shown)


jvperrin:
This breaks docker running on a few virtual machines, however [this comment](https://github.com/moby/moby/issues/14026#issuecomment-128055691) fixes it (essentially need to remove old aufs stuff from VMs that were upgraded in-place using `sudo rm -rf /var/lib/docker/aufs`)

I've run this on fireball and raptors, the only two that appear to be affected.

PR #166, opened 2017-05-09
Set PulseAudio default sinks on snowflake nodes

abizer:
On cyclone, invasion, and tornado, there are multiple PulseAudio sinks. cyclone because we want to use the USB headphone jack, invasion because it has a different? motherboard, and tornado because it's the TV. 

This sets the correct default sinks on those special nodes.

### DIFF
--- a/hieradata/nodes/cyclone.yaml
+++ b/hieradata/nodes/cyclone.yaml
@@ -1,1 +1,2 @@
+# get this value by running `pactl list sinks`
 pulse_sink: alsa_output.usb-C-Media_Electronics_Inc._USB_Audio_Device-00.analog-stereo
--- a/hieradata/nodes/cyclone.yaml
+++ b/hieradata/nodes/cyclone.yaml
@@ -1,0 +1,1 @@
+pulse_sink: alsa_output.usb-C-Media_Electronics_Inc._USB_Audio_Device-00.analog-stereo
--- a/modules/ocf_desktop/manifests/pulse.pp
+++ b/modules/ocf_desktop/manifests/pulse.pp
@@ -19,4 +19,10 @@ class ocf_desktop::pulse {
     require => Package['libasound2-plugins'],
   }
 
+  $pulse_sink = lookup({name => 'pulse_sink', default => nil})
+
+  if $pulse_sink {
+    exec { "pactl set-default-sink ${pulse_sink}": }y
+  }
+
 }
--- a/modules/ocf_desktop/manifests/pulse.pp
+++ b/modules/ocf_desktop/manifests/pulse.pp
@@ -19,11 +19,12 @@ class ocf_desktop::pulse {
     require => Package['libasound2-plugins'],
   }
 
-  $pulse_sink = lookup({'name' => 'pulse_sink', 'default_value' => nil})
+  $pulse_sink = lookup('pulse_sink')
   if $pulse_sink {
-    exec { 'set-pulse-default-sink':
-      command => "sed -i 's/#set-default-sink output/set-default-sink ${pulse_sink}/' /etc/pulse/default.pa",
-      unless => 'grep ^set-default-sink /etc/pulse/default.pa > /dev/null',
+    augeas { 'set-pulse-default-sink':
+      incl => '/etc/pulse/default.pa',
+      lens => 'Spacevars.simple_lns',
+      changes => "set set-default-sink ${pulse_sink}",
     }
   }
 }
--- a/modules/ocf_desktop/manifests/pulse.pp
+++ b/modules/ocf_desktop/manifests/pulse.pp
@@ -19,10 +19,11 @@ class ocf_desktop::pulse {
     require => Package['libasound2-plugins'],
   }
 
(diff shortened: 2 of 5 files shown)


modules/ocf_desktop/manifests/pulse.pp
```diff
@@ -19,4 +19,11 @@
     require => Package['libasound2-plugins'],
   }
 
+  $pulse_sink = lookup({'name' => 'pulse_sink', 'default_value' => nil})
+  if $pulse_sink {
+    exec { 'set-pulse-default-sink':
+      command => "sed -i 's/#set-default-sink output/set-default-sink ${pulse_sink}/' /etc/pulse/default.pa",
+      unless => 'grep ^set-default-sink /etc/pulse/default.pa > /dev/null',
```
tahabi: this doesn't seem like a particularly elegant way of doing this

jvperrin: Augeas might work well here instead

hieradata/nodes/cyclone.yaml
```diff
@@ -0,0 +1 @@
+pulse_sink: alsa_output.usb-C-Media_Electronics_Inc._USB_Audio_Device-00.analog-stereo
```
jvperrin: I'd document where you got these from, if they ever need to be updated or changed in any way.

jvperrin: (and by document I mean add a comment lol)

modules/ocf_desktop/manifests/pulse.pp
```diff
@@ -19,4 +19,11 @@
     require => Package['libasound2-plugins'],
   }
 
+  $pulse_sink = lookup({'name' => 'pulse_sink', 'default_value' => nil})
```
jvperrin: Does this need an explicit default? I was under the impression it did not, although I have not worked that closely with hiera.

tahabi: I got that from here: https://www.devco.net/archives/2016/03/13/the-puppet-4-lookup-function.php
I'm just starting to figure out how all this works together, so I'm not sure.

hieradata/nodes/cyclone.yaml
tahabi: good idea

modules/ocf_desktop/manifests/pulse.pp
tahabi: yeah, that's probably advisable. I guess I can use Spacevars.simple_lns

modules/ocf_desktop/manifests/pulse.pp
jvperrin: Ah, alright then, I'd try it without the default then, so just `lookup('pulse_sink')` and I think that should work fine.

chriskuehl:
I guess this approach is fine but I really wish we could instead figure out how to automate this based on some rules. If it's something you can figure out by looking at the output of a command, couldn't a script also figure it out by looking at that output?

tahabi:
Sure, but I can't think of an a priori way of knowing (or elegantly
scripting) which sink is correct just by listing all the sinks attached to
the system. In cyclone's case for example, there are two valid sinks, the
USB jack and the 3.5mm port on the front of the case, and Pulse reasonably
chooses the case front port as default by itself, but we want to override
this.

On May 9, 2017 9:05 AM, "Chris Kuehl" <notifications@github.com> wrote:

I guess this approach is fine but I really wish we could instead figure out
how to automate this based on some rules. If it's something you can figure
out by looking at the output of a command, couldn't a script also figure it
out by looking at that output?

—
<https://github.com/ocf/puppet/pull/166#issuecomment-300213707>, or mute
the thread

chriskuehl:
I'm okay with some OCF-specific logic, e.g.:

1. If it has a USB sink, assume we put it there for a reason, so use that.
2. If it has <whatever special TV sink>, assume we want to use that
3. Otherwise do the default thing.

etc

I think long-term this is a better solution.

jvperrin:
I do agree that having OCF-specific logic is fine here and would work better for the future, since we'd be able to customize it however we want, so it shouldn't have any problems.

modules/ocf_tv/manifests/init.pp
```diff
@@ -63,4 +63,11 @@
       User['ocftv'],
     ],
   }
+
+  $pulse_sink = lookup('pulse_sink')
+
+  exec { "pactl set-default-sink ${pulse_sink}":
```
jvperrin: I believe this will run every puppet run, which isn't great (although not terrible either since I assume it's fast), is there some other way this could be set, or maybe some trigger that could be made so that it only runs when needed?

tahabi: we can do the same thing as in ocf_desktop as use augeas, but that's a bit overkill for this I think. We can make it run less frequently as like a daily cronjob or something but this is a very lightweight and idempotent operation.

tahabi:
I'll do some research on scripting this instead of going through hiera

PR #167, opened 2017-05-27
Implement Let's Encrypt for app vhosts

matthew-mcallister:
This combines the config generation scripts for death and the apphost
and allows death to request Let's Encrypt certs for app vhosts. It also
allows app vhosts to have aliases and sets up HTTPS redirects for them
on the web and app vhost.


Baisang:
:D

modules/ocf_apphost/files/vhost-app.jinja
```diff
@@ -0,0 +1,38 @@
+# {{vhost.comment}}
+server {
+    listen {{vhost.port}};
+    listen [::]:{{vhost.port}};
+    server_name "{{vhost.fqdn}}";
+
+    # Proxy /.well-known/ to www.o.b.e/.well-known/ so death can request
+    # Let's Encrypt certificates for app vhosts
```
chriskuehl: this is kinda weird. imo servers should request their own certificates (this is what the RFC i sent out suggested)

why centralize everything on death? what if we have servers that want to use LE but don't mount NFS?

modules/ocf_apphost/manifests/proxy.pp
```diff
@@ -16,15 +16,32 @@
...
 
-  cron {
-    'rebuild-vhosts':
-      command => '/usr/local/sbin/rebuild-vhosts > /dev/null',
-      minute  => '*/10',
-      require => Package['nginx'];
+  if $dev_config {
+    cron {
+      'build-vhosts':
+        command => '/usr/local/sbin/build-vhosts --dev app > /dev/null',
```
chriskuehl: i'd suggest using chronic instead of always piping to /dev/null

this way we get the output on failures, but otherwise it gets sent to /dev/null

modules/ocf_www/files/build-vhosts
```diff
@@ -14,190 +11,350 @@ from pathlib import Path
...
-LETS_ENCRYPT_SSL = Path('/services/http/ssl')
+class SSL(namedtuple('SSL', ('fqdn',))):
+    """Tells where to find SSL files for a given domain."""
+    @property
+    def is_valid(self):
+        return os.path.exists(self.cert)
+
+    @property
+    def key(self):
+        return '/etc/ssl/lets-encrypt/le-vhost.key'
```
chriskuehl: heh, you don't need a `@property` function for a constant :)

```python
class SSL(namedtuple('SSL', ('fqdn','))):
    __slots__ = ()
    key = '/etc/ssl/lets-encrypt/le-vhost.key'
    chain = '/etc/ssl/certs/lets-encrypt.crt'
```

modules/ocf_www/files/build-vhosts
```diff
@@ -14,190 +11,350 @@ from pathlib import Path
...
-        'fqdn',
-        'user',
-        'relative_docroot',
-        'redirect_dest',
-        'ssl_cert',
-        'ssl_key',
-        'ssl_chain',
-        'use_hsts',
-        'disabled',
+    'fqdn',
```
chriskuehl: double indent is what pylint suggests /shrug

modules/ocf_www/files/build-vhosts
```diff
@@ -14,190 +11,350 @@ from pathlib import Path
...
 ))):
-    """A single <VirtualHost>.
+    """A single virtual host directive in one of the webserver configs.
 
-    The VirtualHost has a one-to-one mapping to <VirtualHost> directives and is
-    rendered by one invocation of the template.
+    A VirtualHost has a one-to-one mapping to <VirtualHost> (apache) or
+    server (nginx) directives and is rendered by one invocation of the
+    template. This means that some vhost lines might emit multiple
+    VirtualHost instances (redirect domains, etc.).
```
chriskuehl: i think it's a smell when you have one class with multiple responsibilities like this

any time you have multiple uses for a class and a large number of the fields end up being `None`, that's a pretty good indicator that you actually want _two_ classes, and possibly using another to store the common data, e.g. I would probably create one class to hold the common logic and data, e.g.

```python
class Vhost(collections.namedtuple('Vhost', ('user', 'comment', 'ssl' ...))):
    ...

ApacheVhost = collections.namedtuple('ApacheVhost', ('vhost',))
NginxVhost = collections.namedtuple('NginxVhost', ('vhost', 'socket'))
```

modules/ocf_www/files/build-vhosts
```diff
@@ -14,190 +11,350 @@ from pathlib import Path
...
 
         # primary vhost
-        user = vhost['username']
         primary = VirtualHost(
             fqdn=domain,
             user=user,
-            relative_docroot=vhost['docroot'],
+            comment='{} (user {})'.format(domain, user),
+            ssl=ssl if use_ssl else None,
+            # only one of {socket,docroot} will ever be present, so this
```
chriskuehl: assert this?

modules/ocf_www/files/build-vhosts
```diff
@@ -14,190 +11,350 @@ from pathlib import Path
...
-        ):
+        # if it has aliases defined, add HTTP and HTTPS domains redirects too
+        redirects = set()
+        if use_ssl:
+            redirects.add((primary.fqdn, None))
+        for alias in vhost['aliases']:
+            redirects.add((alias, None))
+            ssl = SSL(alias)
+            if ssl.is_valid:
+                redirects.add((alias, ssl))
```
chriskuehl: why these changes? this seems a lot less obvious than what was there before

modules/ocf_www/files/build-vhosts
```diff
@@ -14,190 +11,350 @@ from pathlib import Path
...
+        os.rename(new_path, config_path)
+
+        report('Performing config test.')
+        if target == 'web':
+            ret = subprocess.call(('apachectl', 'configtest'))
+        else:
+            # nginx is moronic and prints "test successful" on stderr,
+            # so echo to stderr only if nginx -t failed
+            p = subprocess.Popen(('nginx', '-t'), stderr=subprocess.PIPE)
+            ret = p.wait()
```
chriskuehl: this can deadlock if nginx produces too much output (it fills the output buffer but nothing is reading from it yet; python is waiting for the process to exit, which it never will because it's trying to write output)

instead you probably want something like...

```python
stdout, stderr = p.communicate()
ret = p.returncode  # communicate() does a wait(), so you don't need to do it again
```

modules/ocf_www/files/build-vhosts
```diff
@@ -14,190 +11,350 @@ from pathlib import Path
...
+            if ret:
+                report(p.stderr.read().decode('utf-8'), file=sys.stderr)
+
+        if ret:
+            report('Test failed! Replacing with old config.')
+            os.rename(old_path, config_path)
+
+        return not ret
```
chriskuehl: i think `return ret == 0` is more readable (and less prone to bugs)

modules/ocf_www/files/build-vhosts
```diff
@@ -14,190 +11,350 @@ from pathlib import Path
...
+                "Warning: user '{}' not in group ocfdev but has app vhost '{}'"
+                .format(user, domain)
+            )
+
+        # Fix socket dir
+        socket_dir = str(APP_DIR / user)
+        try:
+            os.makedirs(socket_dir, 0o755)
+        except FileExistsError:
+            pass
```
chriskuehl: nice :)

modules/ocf_www/files/build-vhosts
```diff
@@ -14,190 +11,350 @@ from pathlib import Path
...
+        if 'nossl' not in vhost['flags'] and ssl.is_valid and (
+            not os.path.exists(ssl.bundle) or
+            os.path.getmtime(ssl.bundle) < os.path.getmtime(ssl.cert)
+        ):
+            report('Generating SSL bundle for ' + domain)
+            f = open(ssl.bundle, 'w')
+            with open(ssl.cert) as f1, open(ssl.chain) as f2:
+                shutil.copyfileobj(f1, f)
+                shutil.copyfileobj(f2, f)
+            f.close()
```
chriskuehl: use a context manager for f as well? you can even do this if you like:

```python
with open(ssl.bundle, 'w') as f, open(ssl.cert) as f1, open(ssl.chain) as f2:
    shutil.copyfileobj(f1, f)
    shutil.copyfileobj(f2, f)
```

modules/ocf_www/files/build-vhosts
matthew-mcallister: Yeah, this is probably better done somehow with polymorphism. I did have a similar class hierarchy earlier but it had large, repetitive constructors; I bet I can improve on that though. Something like this would probably also work but I think it might result in either a bunch of boilerplate functions to delegate to the `vhost` object or I would have to stick `.vhost` accessors in a bunch of places in the template files.

modules/ocf_apphost/files/vhost-app.jinja
matthew-mcallister: It does make sense for the apphost to request its own certs, I might as well fix that.

I'm not really sure what scenario you have in mind where a server besides death or werewolves would want to request LE certs however.

modules/ocf_www/files/build-vhosts
matthew-mcallister: Previously we didn't set up redirects for aliases on port 443, i.e. if you visited `http://xkcd.berkeley.edu` it would redirect you to `xkcd.o.b.e` but not if you visited `https://xkcd.berkeley.edu`.

modules/ocf_www/files/build-vhosts
matthew-mcallister: Nice tip, thanks.

modules/ocf_www/files/build-vhosts
matthew-mcallister: Sorry, that was copypasta from SO.

modules/ocf_apphost/files/vhost-app.jinja
chriskuehl: Check the Google Doc I shared, I listed quite a few scenarios.

modules/ocf_www/files/build-vhosts
matthew-mcallister: I think that this is probably almost as good as it could be. It'd feel kind of awkward to subtype a whole class when only one component needs polymorphism. If this had to scale, I would define subtemplates so the main class and template don't have to know what kind of thing the vhost binds to and choose between Apache and Nginx configs by changing the jinja environment.

chriskuehl: i'm not suggesting subclassing -- but i think having one class represent both a `<VirtualHost>` and nginx `server` block is tying a class too closely to two totally distinct implementations

i would either:
* create a data-only object (like in my snippet) which holds all the common data, then two separate classes for apache/nginx which use that, or
* remove references in the docstring to specific implementations, and just say it represents a logical virtual host; you've already sort of gone this direction by moving the implementation (templating) outside of the class (which now pretty much just holds data and some common logic)

matthew-mcallister: Ah, I see your motivation now. It makes sense, though I still think there is the trouble of having to delegate common methods of the two classes, which sort of leads to inheritence anyways.

Also, I pushed new changes which I was referring to, you just happened to comment where the last four lines stayed the same.

I dunno, I thought of the class as being intentionally generic as to which server was being targeted. Like, if you really wanted, you could actually have apache bind to a socket or have nginx serve up a public_html directory. Given that there is no class specific to nginx or apache, I don't think the docstring will cause confusion, but I do agree it would be more accurate to call it by what it is rather than how it's being used.

modules/ocf/manifests/lets_encrypt.pp
```diff
@@ -0,0 +1,22 @@
+class ocf::lets_encrypt {
```
matthew-mcallister: @chriskuehl: This might be relevant to https://ocf.io/rt/6193.

modules/ocf_www/files/build-vhosts
```diff
@@ -16,188 +13,357 @@ import jinja2
...
+        use_ssl = 'nossl' not in vhost['flags'] and ssl.is_valid
+
+        # Exactly one must be true
+        assert ('socket' in vhost) != ('docroot' in vhost)
+        if 'socket' in vhost:
+            bind_type = 'socket'
+            bind_dest = vhost['socket']
+        else:
+            bind_type = 'docroot'
+            bind_dest = vhost['docroot']
```
jvperrin: Minor thing, but you could simplify this slightly:

```python
if 'socket' in vhost:
    bind_type = 'socket'
else:
    bind_type = 'docroot'
bind_dest = vhost[bind_type]
```

modules/ocf_www/files/build-vhosts
jvperrin: lol

modules/ocf_www/files/build-vhosts
```diff
@@ -16,188 +13,357 @@ import jinja2
...
+    if target == 'app' and not args.dry_run:
+        process_app_vhosts()
 
-    fh, path = tempfile.mkstemp(dir='/etc/apache2', prefix='ocf-vhost.conf')
-    try:
-        os.chmod(path, 0o644)
-        os.write(fh, config_file.encode('utf8'))
-        os.close(fh)
-        report(path)
+    global dev_config
```
jvperrin: I feel like this could be removed and replaced with an argument since it's only read in `dev_alias` from what I can see.

modules/ocf_www/files/build-vhosts
```diff
@@ -16,188 +13,357 @@ import jinja2
...
+        report('Copied existing config to ' + old_path)
+
+        # Overwrite with new config and test
+        os.rename(new_path, config_path)
+
+        report('Performing config test.')
+        if target == 'web':
+            ret = subprocess.call(('apachectl', 'configtest'))
+        else:
+            ret = subprocess.call(('nginx', '-t'))
```
jvperrin: This is cool, nice!

modules/ocf_www/files/build-vhosts
matthew-mcallister: It could, but as a general rule it kinda sucks to pass around global behavior options (`--debug`, `--verbose`) everywhere you go and insert them into your data types. :(

In this case it's pretty easy to do, though, so I may as well.

modules/ocf_www/files/build-vhosts
matthew-mcallister: Oh, was I supposed to use `exist_ok=True`?

modules/ocf_www/files/build-vhosts
jvperrin: Yeah, I do agree it sucks to pass around stuff like that, but globals are generally not great either, and as you said in this case it's pretty easy to change anyway.

modules/ocf_www/files/build-vhosts
chriskuehl: i didn't actually realize that existed for `os.makedirs` -- but since it does, that's actually probably a good idea

(my comment was not at all sarcastic, i was commenting on how this _was_ the correct way to do that. i didn't realize python 3.2 added this option)

jvperrin: Ah, I think I misinterpreted that then, might be good to have a short comment here saying why it can be ignored then? Whenever I see an error that is caught and then ignored I assume the worst :P

matthew-mcallister:
Pushed changes. @jvperrin any more comments?

jvperrin:
Nope, I don't think I have any, thanks for doing this!

PR #168, opened 2017-06-13
Make makevm use an SSH key

kpengboy:


modules/ocf_kvm/manifests/init.pp
```diff
@@ -42,5 +42,14 @@
       ensure => link,
       links  => manage,
       target => '/opt/share/utils/staff/sys/makevm';
+
+    '/opt/share/kvm':
+      ensure => directory;
+
+    '/opt/share/kvm/makevm-ssh-key':
+      content   => file('/opt/puppet/shares/private/makevm-ssh-key'),
```
chriskuehl: what if we just generated this? i don't think we really care about having a persistent key between kvm hosts or if we rebuild the hosts?

something like

```puppet
exec { 'ssh-keygen -t ed25519 -f /opt/share/kvm/makevm-ssh-key -N ""':
  creates => '/opt/share/kvm/makevm-ssh-key',
}
```

or whatever

kpengboy: We'd have to somehow get the public key(s) into the netboot image or preseed file though, which would greatly complicate the task.

chriskuehl: ah that's true. shipit

PR #169, opened 2017-06-17
mirrors stats script

tahabi:


modules/ocf_mirrors/manifests/init.pp
```diff
@@ -45,6 +47,10 @@
       source => 'puppet:///modules/ocf_mirrors/README.html',
       owner  => mirrors,
       group  => mirrors;
+
+    '/usr/local/sbin/record-mirrors-stats':
+      content => template('ocf_mirrors/record-mirrors-stats.py.erb'),
+      mode    => '0644';
```
chriskuehl: if it's in sbin it should be executable. additionally, if it has secrets, it shouldn't be world-readable and should have `show_diff => false`

ideally don't template a python file though. store the config separately.

modules/ocf_mirrors/templates/record-mirrors-stats.py.erb
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/python
```
chriskuehl: python3?

modules/ocf_mirrors/templates/record-mirrors-stats.py.erb
```diff
@@ -0,0 +1,54 @@
...
+
+import os
+import pymysql
+from datetime import date
```
chriskuehl: group imports into stdlib / 3rd party / application per PEP8: https://www.python.org/dev/peps/pep-0008/#imports

pre-commit should enforce this for you?

modules/ocf_mirrors/templates/record-mirrors-stats.py.erb
```diff
@@ -0,0 +1,54 @@
...
+
+OCFSTATS_PWD = "<%= @ocfstats_password %>"
+MIRRORS_PATH = '/opt/mirrors/ftp'
+LOG_PATH = '/var/log/apache2'
+LOG_NAME = 'mirrors.ocf.berkeley.edu_access.log'
+
+sources = [mirrored for mirrored in os.listdir(MIRRORS_PATH)
+           if os.path.isdir(os.path.join(MIRRORS_PATH, mirrored)) and not mirrored.startswith('.')]
+
+sources.append('other') # catchall
```
chriskuehl: 2 spaces before inline comments per PEP8: https://www.python.org/dev/peps/pep-0008/#inline-comments

pre-commit isn't enforcing this?

modules/ocf_mirrors/templates/record-mirrors-stats.py.erb
```diff
@@ -0,0 +1,54 @@
...
+
+def build_dists():
+    return { mirrored: { 'up': 0, 'down': 0 } for mirrored in sources }
+
+def process_file(fn):
```
chriskuehl: 2 blank lines before/after lines per PEP8 (going to stop commenting these, pre-commit should do this for you)

modules/ocf_mirrors/manifests/init.pp
matthew-mcallister: 1. I don't think this should be world-readable to be consistent with our use of secrets elsewhere.
2. Shouldn't this be executable?

modules/ocf_mirrors/templates/record-mirrors-stats.py.erb
```diff
@@ -0,0 +1,54 @@
...
+    conn = pymysql.connect(
+        host = 'mysql.ocf.berkeley.edu',
+        user = 'ocfstats',
+        password = OCFSTATS_PWD,
+        db='ocfstats',
+        autocommit = True,
+        cursorclass=pymysql.cursors.DictCursor,
+    )
+
+    c = conn.cursor()
```
chriskuehl: use a context manager

```python
with conn as cursor:
   ...
```

modules/ocf_mirrors/templates/record-mirrors-stats.py.erb
```diff
@@ -0,0 +1,54 @@
...
+def build_dists():
+    return { mirrored: { 'up': 0, 'down': 0 } for mirrored in sources }
+
+def process_file(fn):
+    dists = build_dists()
+    with open(fn, 'r') as f:
+        for line in f:
+            stats = line.split()
+            dist = stats[6]
+            dist = dist.split('/')[1] if '/' in dist else dist
```
chriskuehl: can you include a comment with an example line so people reading the script can see what is being parsed?

modules/ocf_mirrors/manifests/init.pp
```diff
@@ -156,4 +162,12 @@
     ssl_cert  => "/etc/ssl/private/${::fqdn}.crt",
     ssl_chain => '/etc/ssl/certs/incommon-intermediate.crt',
   }
+
+  cron {
+    'mirrors-stats':
+      command => '/usr/local/sbin/record-mirrors-stats',
```
chriskuehl: don't know how this works if it's not executable?

modules/ocf_mirrors/templates/record-mirrors-stats.py.erb
```diff
@@ -0,0 +1,54 @@
...
+
+            if stats[8][0] in ('2', '3') and dist in dists: # http status code is 2xx/3xx
```
matthew-mcallister: Comment this as well as the above lines, please.

modules/ocf_mirrors/manifests/init.pp
tahabi: Yes, one sec will fix

matthew-mcallister: You should probably store config separately as @chriskuehl says and as we do for all of our other programs.

tahabi: yeah, my mistake 

I modeled this after how it's done in the labstats script, I guess I'll fix this and then fix that as well

modules/ocf_mirrors/manifests/init.pp
tahabi: seems to work fine for the last two weeks I let it run, but yeah I need to make it +x

modules/ocf_mirrors/templates/record-mirrors-stats.py.erb
tahabi: pre-commit didn't say anything about this, I'll re-install it I guess

modules/ocf_mirrors/manifests/init.pp
matthew-mcallister: Maybe cron runs the crunchbang without checking if it's +x? Sounds fishy to me.

modules/ocf_mirrors/manifests/init.pp
tahabi: it now uses environment variables, seeded by cron. The password is stored in the root crontab so it should be secure enough + no need to add another config file. Does that seem reasonable?

modules/ocf_mirrors/files/record-mirrors-stats.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/python
```
chriskuehl: python files should either drop the `.py` if they're not intended to be imported (and only executed), or use underscores so they can be imported

our tradition is typically to drop the extension

modules/ocf_mirrors/files/process-mirrors-logs
```diff
@@ -0,0 +1,82 @@
...
+import argparse
+import os
+from datetime import date
+
+import pymysql
+
+OCFSTATS_PWD = os.environ['OCFSTATS_PWD']
+MIRRORS_PATH = '/opt/mirrors/ftp'
+LOG_PATH = '/var/log/apache2'
+LOG_NAME = 'mirrors.ocf.berkeley.edu_access.log'
```
jvperrin: I think I forgot to mention this before, but the reason why I was using `mirrors.ocf.berkeley.edu_access.log.1` instead of this was because the logs get rotated at around 6 AM, so if you check them once a day, unless you check very close to when they are rotated, you lose some data. For example, this is currently checked at midnight, so all the data from midnight to around 6 AM is currently lost since it is rotated before the next midnight.

tahabi:
I changed the cron.daily timing to be at 12:05am for precisely this reason.
Was trying to figure out how to puppet that for logrotate only.

On Jun 27, 2017 9:17 PM, "Jason Perrin" <notifications@github.com> wrote:

*@jvperrin* commented on this pull request.

In modules/ocf_mirrors/files/process-mirrors-logs
<https://github.com/ocf/puppet/pull/169#discussion_r124449152>:

> @@ -0,0 +1,82 @@
+#!/usr/bin/python
+import argparse
+import os
+from datetime import date
+
+import pymysql
+
+OCFSTATS_PWD = os.environ['OCFSTATS_PWD']
+MIRRORS_PATH = '/opt/mirrors/ftp'
+LOG_PATH = '/var/log/apache2'
+LOG_NAME = 'mirrors.ocf.berkeley.edu_access.log'

I think I forgot to mention this before, but the reason why I was using
mirrors.ocf.berkeley.edu_access.log.1 instead of this was because the logs
get rotated at around 6 AM, so if you check them once a day, unless you
check very close to when they are rotated, you lose some data. For example,
this is currently checked at midnight, so all the data from midnight to
around 6 AM is currently lost since it is rotated before the next midnight.

—
<https://github.com/ocf/puppet/pull/169#pullrequestreview-46736910>, or mute
the thread

modules/ocf_mirrors/files/process-mirrors-logs
```diff
@@ -0,0 +1,82 @@
...
+
+def process_file(fn):
+    dists = build_dists()
+    with open(fn, 'r') as f:
+        for line in f:
+            stats = line.split()
+
+            # extract dist name from request url
+            # '/debian/pool/main/h/hwdata/...' -> 'debian'
+            dist = stats[6]
```
jvperrin: I wrote this script as a bit of a hack to quickly see what traffic levels we had across dists, but ideally there would be a comment here explaining why these indices were chosen with maybe an example line:

The format comes from [here](https://github.com/ocf/puppet/blob/2e9e2cd12819653c7eb4c8ad362736b4f65122aa/modules/ocf_mirrors/manifests/init.pp#L56):

    %h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\" %I %O

which translates to something like

    8.8.8.8 - - [26/Jun/2017:00:05:13 -0700] "GET /centos/7.3.1611/path/here/blah HTTP/1.1" 200 3735 "-" "urlgrabber/3.10 yum/3.4.3" 147 3971

so that's why the 6th index is needed (`%r` corresponds to the path because there is an extra space in the timestamp), and the 8th index below gives the status code so that can be filtered for only successes. -1 (`%O`) and -2 (`%I`) below give the bytes downloaded and uploaded as seen by Apache respectively.

modules/ocf_mirrors/files/process-mirrors-logs
jvperrin: Another option I just thought of is to scan both log files and check if the date is within the current day, that might be the best if you truly want the numbers to be representative of the actual day boundaries instead of 6 AM - 6 AM the next day. (honestly I don't think it matters, but it wouldn't be too bad to scan both log files either)

jvperrin:
Meh, I think that's a bit much and would be surprising behavior honestly that one system has a different `cron.daily` time than the others for a script to collect mirror metrics. I think just scraping both log files and discarding those in the wrong date would be reasonably easy. The script only runs once a day, so if it takes a few more seconds it's not a big deal at all.

chriskuehl:
+1 for jvperrin's approach

PR #170, opened 2017-07-02
Add more details to munin alerts

chriskuehl:
I can no longer remember what every host at the OCF is :(

I think it's useful to include a little context of what the host does. Ideally we'd have the puppet classes, but they're a bit hard to read after we moved them into hiera, so that probably needs to wait for puppetdb.

This should make it easier to triage these at a glance, and also help newer staff.

### DIFF
--- a/modules/ocf_stats/files/munin/mail-munin-alert
+++ b/modules/ocf_stats/files/munin/mail-munin-alert
@@ -9,19 +9,19 @@ The expected call format from Munin is
 
 The list of warnings/criticals allows filtering of unhelpful messages such as
 those only containing OKs or UNKNOWNs. The list of fields must be in the same
-order as the list of fields.
+order as the list of alerts.
 """
 import argparse
 from sys import argv
 from sys import stdin
 from syslog import syslog
 
+from ocflib.infra import hosts
 from ocflib.misc.mail import MAIL_ROOT
 from ocflib.misc.mail import send_mail
 
 
 MAIL_FROM = 'Munin <munin@ocf.berkeley.edu>'
-MUNIN_HOSTS_URL = 'https://munin.ocf.berkeley.edu/ocf.berkeley.edu'
 
 
 # Filter out uninteresting or noisy warnings at the individual level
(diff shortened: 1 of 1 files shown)


modules/ocf_stats/files/munin/mail-munin-alert
```diff
@@ -33,14 +34,34 @@ IGNORED_WARNINGS = {
...
+    )) or '(none)'
+
+    subject = 'Munin alert for {} ({[type]})'.format(host, ldap_entry)
+    body = (
+        '{status}\n'
+        'See graph at https://munin.ocf.berkeley.edu/ocf.berkeley.edu/{host}/{plugin}.html\n'
+        '\n'
+        'Host details:\n'
+        '  - Type: {ldap_entry[type]}\n'
+        '  - Environment: {ldap_entry[environment][0]}\n'
```
chriskuehl: side note, we should probably change our ldap schema so that "environment" can have only one value

modules/ocf_stats/files/munin/mail-munin-alert
```diff
@@ -33,14 +34,34 @@ IGNORED_WARNINGS = {
...
-    body += '\nSee graph at {}/{}/{}.html'.format(MUNIN_HOSTS_URL, host, plugin)
+    ldap_entry, = hosts.hosts_by_filter('(cn={})'.format(host))
+    dns_names = ', '.join(sorted(
+        ldap_entry.get('dnsCname', []) + ldap_entry.get('dnsA', []),
+    )) or '(none)'
+
+    subject = 'Munin alert for {} ({[type]})'.format(host, ldap_entry)
+    body = (
+        '{status}\n'
+        'See graph at https://munin.ocf.berkeley.edu/ocf.berkeley.edu/{host}/{plugin}.html\n'
```
jvperrin: Why not use `MUNIN_HOSTS_URL` here? If not, probably should remove that constant since it doesn't seem to be used at all.

modules/ocf_stats/files/munin/mail-munin-alert
```diff
@@ -9,13 +9,14 @@ The expected call format from Munin is
 
 The list of warnings/criticals allows filtering of unhelpful messages such as
 those only containing OKs or UNKNOWNs. The list of fields must be in the same
-order as the list of fields.
+order as the list of alerts.
```
matthew-mcallister: Nice.

modules/ocf_stats/files/munin/mail-munin-alert
chriskuehl: oops yeah, i meant to delete that constant

modules/ocf_stats/files/munin/mail-munin-alert
jvperrin: Yeah, I definitely agree, changing the LDAP schema is tough though, however it might be easier now that the schema is stored directly in LDAP now?

PR #173, opened 2017-07-04
Update Puppet pre-commit hooks to 2.0.1

chriskuehl:
No real changes, just a rename from hooks.yaml => .pre-commit-hooks.yaml to get rid of the deprecation warning.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,6 +10,7 @@
         -   --no-documentation-check
         -   --no-puppet_url_without_modules-check
         -   --no-arrow_alignment-check
+        -   --no-arrow_on_right_operand_line-check
 -   repo: https://github.com/pre-commit/pre-commit-hooks.git
     sha: v0.9.1
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 -   repo: https://github.com/chriskuehl/puppet-pre-commit-hooks.git
-    sha: v2.0
+    sha: v2.0.1
     hooks:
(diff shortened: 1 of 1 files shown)


chriskuehl:
huh, apparently not a noop since puppet-lint has new rules:

```
modules/ocf/manifests/serial_getty.pp - WARNING: arrow should be on the right operand's line on line 7
modules/ocf_mesos/manifests/master/mesos.pp - WARNING: arrow should be on the right operand's line on line 58
modules/ocf_mesos/manifests/slave.pp - WARNING: arrow should be on the right operand's line on line 20
modules/ocf_admin/manifests/create/redis.pp - WARNING: arrow should be on the right operand's line on line 43
modules/ocf_admin/manifests/create/redis.pp - WARNING: arrow should be on the right operand's line on line 48
modules/ocf_admin/manifests/create/redis.pp - WARNING: arrow should be on the right operand's line on line 51
modules/ocf/manifests/extrapackages.pp - WARNING: arrow should be on the right operand's line on line 243
```

chriskuehl:
I'm not sure I like that rule but it appears to be the official styleguide: https://docs.puppet.com/puppet/latest/style_guide.html#chaining-arrow-syntax

will change our uses

chriskuehl:
on second thought, it looks really dumb when you write something like

```puppet
file {
  ...
}
-> file {
  ...
}
```

i'm just going to add that to our exclude list for now and see if anyone else has an opinion. i don't particularly like that having the arrow on the left causes the resources to look indented

jvperrin:
It does make the dependency more clear, but yeah I don't like the indentation much. Is it ok if the arrow is on its own line?

PR #175, opened 2017-07-08
Add check-executables-have-shebangs pre-commit hook

chriskuehl:
Pretty hilarious that we copy-pasted '#/bin/sh` everywhere and never noticed. I'm pretty sure it originated from copying from the Ubuntu wiki: https://wiki.ubuntu.com/Mirrors/Scripts#The_script

Somehow it works... but we should probably use a proper shebang anyway :)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,6 +21,7 @@
     -   id: check-added-large-files
     -   id: check-docstring-first
         language_version: python3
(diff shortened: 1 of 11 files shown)


chriskuehl:
more info about what this hook does: https://github.com/pre-commit/pre-commit-hooks/pull/211

jvperrin:
Lol that's hilarious

PR #179, opened 2017-08-17
Systemd anope and cleanup ocf_irc

jvperrin:
This should fix the issue where `anope` starts before `inspircd` and then `anope` has to be manually restarted to get services working. I welcome any systemd unit file critique!

I'd like to upstream the systemd changes for anope if possible, seems like it would be cool since the Debian package is still using the SysV init script.


modules/ocf_irc/files/anope.service
```diff
@@ -0,0 +1,17 @@
...
+After=network-online.target inspircd.service
+
+[Service]
+User=irc
+Group=irc
+Restart=always
+# Run ExecStartPre with root permissions to be able to create pid directory
+PermissionsStartOnly=true
+ExecStartPre=/bin/mkdir -p /var/run/anope/
+ExecStartPre=/bin/chown -R irc:irc /var/run/anope/
```
matthew-mcallister: `install -o irc -g irc -d /var/run/anope`?

modules/ocf_irc/files/anope.service
```diff
@@ -0,0 +1,17 @@
...
+Description=A set of IRC Services
+After=network-online.target inspircd.service
+
+[Service]
+User=irc
+Group=irc
+Restart=always
+# Run ExecStartPre with root permissions to be able to create pid directory
+PermissionsStartOnly=true
+ExecStartPre=/bin/mkdir -p /var/run/anope/
```
chriskuehl: probably `/run/anope`? `/var/run` is a symlink to `/run` on newer versions of debian/ubuntu

jvperrin: I mean that works too, it seems like `/var/run/anope/anope.pid` is where it tries to make the PID (with no option I could find to change the PID directory??), so that's why I chose this. If it's a symlink then that's fine too, although I'd prefer to keep it pointing to the actual path that anope will use for creating the PID file.

modules/ocf_irc/files/anope.service
jvperrin: Yeah, that seems reasonable, although I think it is more readable currently, especially if someone doesn't know what `install` does, which is pretty common... That being said it's a cleaner command to have instead of two separate ones to do the same thing.

modules/ocf_irc/files/anope.service
```diff
@@ -0,0 +1,17 @@
...
+ExecStartPre=/bin/chown -R irc:irc /var/run/anope/
+# Run anope with regular permissions (irc:irc)
+ExecStart=/usr/sbin/anope --debug --nofork --confdir=/etc/anope --dbdir=/var/lib/anope/db --logdir=/var/log/anope --modulesdir=/usr/lib/anope --localedir=/usr/share/anope/locale
```
jvperrin: I just copied this line directly from the packaged anope init script (`/etc/init.d/anope`). It just gives more output which can sometimes be useful, but I'm open to removing it.

modules/ocf_irc/files/anope.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=A set of IRC Services
+After=network-online.target inspircd.service
```
jvperrin: Yeah, good point, this essentially replaces that.

modules/ocf_irc/files/anope.service
chriskuehl: ah, if it's actually writing to /var/run/ then i'd agree we should keep that

jvperrin:
@kpengboy No, I'm not actually sure why the systemd drop-in doesn't work, perhaps because `anope` uses a SysV init script that has to be converted to systemd before running? Not sure if that would mess it up or something?

modules/ocf_irc/files/anope.service
matthew-mcallister: You can leave it as is, I only meant it as a suggestion.

jvperrin: If the original script does that then it seems more reasonable to copy, I'll do that.

PR #180, opened 2017-08-17
Staff ZNC server

matthew-mcallister:
The installation is very simple; all the config is living unmanaged in `/opt/znc` on flood and has to be manually managed by the web interface. Keeping it backed up should be sufficient in case we have to do a restore.

### DIFF
--- a/modules/ocf_irc/files/znc.service
+++ b/modules/ocf_irc/files/znc.service
@@ -1,0 +1,9 @@
+[Service]
+ExecStart=/usr/bin/znc --foreground --datadir /var/lib/znc
+User=ocfznc
+Group=ocfznc
+Restart=always
+SyslogIdentifier=znc
+
+[Install]
+WantedBy=multi-user.target
--- a/modules/ocf_irc/manifests/znc.pp
+++ b/modules/ocf_irc/manifests/znc.pp
@@ -1,0 +1,28 @@
+# A minimal ZNC installation
+# All extra configuration is managed by web interface and kept in backups
+class ocf_irc::znc {
+  package { 'znc': }
+
+  user { 'ocfznc':
+    comment => 'IRC Bouncer Server',
+    groups  => [ssl-cert],
+    home    => '/var/lib/znc',
+    shell   => '/bin/false',
+    system  => true,
+    require => Package['ssl-cert'],
+  }
+
+  file {
+    '/var/lib/znc':
+      ensure  => directory,
+      owner   => ocfznc,
+      group   => ocfznc,
+      mode    => '0700',
+      require => User['ocfznc'],
+  }
+
+  ocf::systemd::service { 'znc':
+    source  => 'puppet:///modules/ocf_irc/znc.service',
+    require => [Package['znc'], File['/var/lib/znc']],
+  }
+}
--- a/modules/ocf_backups/files/rsnapshot.conf
+++ b/modules/ocf_backups/files/rsnapshot.conf
@@ -73,4 +73,6 @@ backup	ocfbackups@rancid:/var/lib/rancid/	servers/rancid/
 
 backup	ocfbackups@ns:/etc/bind/keys/	servers/ns/
 
+backup	ocfbackups@irc:/var/lib/znc/	servers/irc/
+
 # vim: ts=16 sts=16 sw=16 noet
--- a/modules/ocf_irc/manifests/init.pp
+++ b/modules/ocf_irc/manifests/init.pp
@@ -3,6 +3,7 @@ class ocf_irc {
   include ocf_irc::ircd
   include ocf_irc::services
   include ocf_irc::nodejs::webirc
+  include ocf_irc::znc
 
   class { 'ocf_irc::nodejs::apt':
     stage => first,


matthew-mcallister:
Oh woops, time to rebase

modules/ocf_irc/files/znc.service
```diff
@@ -0,0 +1,11 @@
...
+ExecStart=/usr/bin/znc --foreground --datadir /opt/znc
+User=znc
+Group=znc
+Restart=always
+StandardOutput=syslog
+StandardError=syslog
+SyslogIdentifier=znc
+
+[Install]
+WantedBy=multi-user.target
```
jvperrin: Maybe include the reload command here like the one from the [docs](https://www.ocf.berkeley.edu/docs/staff/staffvm/znc/)? I don't think the syslog or stdout/stderr stuff is necessary either probably, but I also haven't tested it.

matthew-mcallister: Yeah, I was going to include a reload here, but I read this in the ZNC man page:
```
SIGNALS
       This section explains how znc reacts to different signals:

       SIGINT Exit ZNC. This is equivalent to /znc shutdown

       SIGHUP Reload znc.conf. This is equivalent to /znc rehash.  DO NOT do this
              very often, things can break badly!

       SIGUSR1
              Rewrite znc.conf. This is equivalent to /znc saveconfig
```
It sounds like sending a HUP reloads the config file _without_ saving changes from the web interface since the last write. Is that the right behavior, meaning won't you lose data on a reload that way?

Also, you're right, the stderr/stdout redirects were for testing.

matthew-mcallister:
Heads up, there is a merge conflict with https://github.com/ocf/puppet/pull/179, so whoever merges last will have to rebase

modules/ocf_irc/files/znc.service
jvperrin: Ah ok, that's good to know, probably good to change the docs for ZNC then, since it sounds like sending a SIGHUP isn't great.

matthew-mcallister: OK, I'll test if it does what I think it does and update the docs if it really is problematic.

modules/ocf_irc/files/znc.service
```diff
@@ -0,0 +1,11 @@
+[Service]
+ExecStart=/usr/bin/znc --foreground --datadir /opt/znc
```
chriskuehl: data probably belongs in `/var/lib/znc` or `/srv/znc` and not `/opt/znc`?

quoth `man hier`:

> `/opt   This directory should contain add-on packages that contain static files.`

> `/var/lib/<package>    State data for packages and subsystems (optional).`

> `/srv   This directory contains site-specific data that is served by this system.`

modules/ocf_irc/manifests/znc.pp
```diff
@@ -0,0 +1,28 @@
+# A minimal ZNC installation
+# All extra configuration is managed by web interface and kept in backups
+class ocf_irc::znc {
+  package { 'znc': }
+
+  user { 'znc':
```
chriskuehl: call it ocfznc? i think we should be naming all users we create starting with `ocf` when possible to avoid conflicts with ldap or debian packaging

modules/ocf_irc/files/znc.service
jvperrin: Wow, I had no idea `man hier` was a thing, cool

jvperrin:
I am planning on rebuilding flood soon to upgrade it to stretch since there is nothing blocking it now, so that might make this a little more difficult since I'll have to restore from the backup of the ZNC config.

modules/ocf_irc/manifests/znc.pp
jvperrin: Good idea, I think that helps a lot to make it clear where the user comes from.

matthew-mcallister: That sounds like a very good practice.

modules/ocf_irc/files/znc.service
matthew-mcallister: This is true. I wasn't sure whether to put under `/var` or `/opt`, but I saw we already use `/opt` for a few things, like stats, mirrors, and apt, so I just went with that. i'll take your suggestion, @chriskuehl.

matthew-mcallister:
If you want to test it, restoring from backups should be pretty easy since all you have to do is `rsync` it. Alternatively, we can just hold off on creating accounts until after the rebuild.

modules/ocf_irc/files/znc.service
matthew-mcallister: [The ZNC documentation](https://wiki.znc.in/FAQ#How_to_reload_ZNC_config_without_restart.3F) makes it sound like doing this is highly discouraged.

It might make sense to do a SIGUSR1 followed by SIGHUP, but I don't really know to what end. It seems like it writes to disk any time you save something in the web interface, so I don't know when the running config would differ from the written one. Perhaps if you connected to IRC as the admin user and did some commands, but I don't think we want to encourage that anyways.

jvperrin: Alright, well this works anyway, I imagine either way will work fine anyway, it's not like there are that many reloads anyway.

PR #181, opened 2017-08-20
ocf_mysql: Use new innodb format, large prefixes

matthew-mcallister:
Some software requires these features from newer MySQL versions. They're supported in MariaDB but have to be enabled in < 10.2. This won't update existing tables until they are rebuilt.


matthew-mcallister:
I don't really know how to test thoroughly, but it started and I could create tables.

jvperrin:
Seems fine, I mean if you could create tables, and existing tables worked fine then that should be fine.

Issue #182, opened 2017-08-22
Investigate Docker repo


matthew-mcallister:
It seems like apt.dockerproject.org hasn't updated since May 5 even though it was getting monthly updates before. We might consider switching to download.docker.com, [which is apparently the official source now](https://docs.docker.com/engine/installation/linux/docker-ce/debian/).

jvperrin:
The docker package changed name too, so we need to update that from `docker-engine` to `docker-ce` wherever it is referenced.

chriskuehl:
ffs, how many times are they going to rename the docker package?

`lxc-docker` => `docker.io` => `docker-engine` => `docker-ce`

jvperrin:
This is resolved by #183.

PR #183, opened 2017-08-23
Update Docker apt repo and cleaning rules

matthew-mcallister:
This addresses https://github.com/ocf/puppet/issues/182. It also has some improvements to the cleanup cronjobs which you can review while you're at it.

### DIFF
--- a/modules/ocf/manifests/packages/docker.pp
+++ b/modules/ocf/manifests/packages/docker.pp
@@ -15,11 +15,6 @@ class ocf::packages::docker($admin_group = undef,
     stage => first,
   }
 
-  package {
-    'docker.io':
-      ensure  => purged;
-  }
-
   if $::lsbdistcodename != 'jessie' {
     package {
       ['aufs-dkms', 'aufs-tools']:
@@ -29,9 +24,8 @@ class ocf::packages::docker($admin_group = undef,
 
   # Don't install AUFS stuff
   ocf::repackage {
-    'docker-engine':
+    'docker-ce':
       recommends => false,
-      require    => Package['docker.io'];
   }
 
   exec { 'docker-socket-update':
@@ -44,7 +38,7 @@ class ocf::packages::docker($admin_group = undef,
     ocf::systemd::override { 'set-docker-socket-group':
       unit    => 'docker.socket',
       content => "[Socket]\nSocketGroup=${admin_group}\n",
-      require => Package['docker-engine'],
+      require => Package['docker-ce'],
       notify  => Exec['docker-socket-update'];
     }
   }
--- a/modules/ocf/manifests/packages/docker.pp
+++ b/modules/ocf/manifests/packages/docker.pp
@@ -10,7 +10,9 @@
 #
 class ocf::packages::docker($admin_group = undef,
                             $autoclean = true,
-                            $image_max_age = '24h') {
+                            $image_max_age = '24h',
+                            $container_max_age = '24h',
+                            $prune_volumes = true) {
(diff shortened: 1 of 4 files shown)


modules/ocf/manifests/packages/docker.pp
```diff
@@ -10,16 +10,13 @@
...
+                            $container_max_age = '24h',
+                            $prune_volumes = true) {
   class { 'ocf::packages::docker::apt':
     stage => first,
   }
 
-  package {
-    'docker.io':
-      ensure  => purged;
-  }
```
matthew-mcallister: The `docker-ce` package `Replaces: docker-engine`, so we don't need the purge rule this time.

modules/ocf/manifests/packages/docker.pp
```diff
@@ -44,35 +40,36 @@
...
-        command => "docker ps -a --filter status=exited | grep -E 'Exited \\([0-9]+\\) [0-9]+ (days|weeks?|months?|years?) ago' | awk '{print \$1}' | chronic xargs -r docker rm",
+        command => "chronic docker container prune --filter until=${container_max_age} -f",
         hour    => 1,
         minute  => 3;
 
-      'clean-old-created-docker-containers':
-        # days is intentionally plural
-        command => "docker ps -a --filter status=created | grep -E '(days|weeks?|months?|years?) ago' | awk '{print \$1}' | chronic xargs -r docker rm",
-        hour    => 1,
-        minute  => 5;
```
matthew-mcallister: Using `docker container prune` already deletes containers with status `created`, so this is deleted.

modules/ocf/manifests/packages/docker/apt.pp
```diff
@@ -1,14 +1,14 @@
...
   apt::source { 'docker':
-    location    => 'http://apt.dockerproject.org/repo',
-    release     => "debian-${::lsbdistcodename}",
-    repos       => 'main',
+    location    => 'https://download.docker.com/linux/debian',
+    release     => $::lsbdistcodename,
+    repos       => 'stable',
     include   => {
       src => false
     },
```
jvperrin: Require `apt-transport-https` here?

modules/ocf/manifests/packages/docker.pp
jvperrin: Ah, nice, that definitely makes this nicer

modules/ocf/manifests/packages/docker.pp
```diff
@@ -44,35 +40,36 @@
...
       notify  => Exec['docker-socket-update'];
     }
   }
 
   if $autoclean {
     cron {
       'clean-old-docker-containers':
-        # days is intentionally plural
-        command => "docker ps -a --filter status=exited | grep -E 'Exited \\([0-9]+\\) [0-9]+ (days|weeks?|months?|years?) ago' | awk '{print \$1}' | chronic xargs -r docker rm",
+        command => "chronic docker container prune --filter until=${container_max_age} -f",
```
jvperrin: Much nicer, thanks!

modules/ocf/manifests/packages/docker/apt.pp
matthew-mcallister: Hm, it doesn't hurt, but it's currently already installed in stage `first`. Would it be good style?

jvperrin: Probably better style since maybe the stage could change or something? Given it's in stage `first` though, I don't think it matters really anyway.

PR #186, opened 2017-09-02
ocf_desktop: remove remmina

kpengboy:
It's not in stretch. We can add it back if it appears in a later Debian version.


kpengboy:
Should we add some sort of comment, or are we ambivalent about remmina going?

jvperrin:
Apparently it [will be available in `stretch-backports` sometime soon](https://www.remmina.org/wp/remmina-1-2-will-be-available-on-debian-9-stretch/), maybe add a comment about adding it back?

PR #187, opened 2017-09-02
ocf: switch to firefox-esr

kpengboy:


modules/ocf/manifests/packages/firefox.pp
```diff
@@ -1,27 +1,13 @@
 class ocf::packages::firefox {
   $browser_homepage = lookup('browser_homepage')
 
-  class { 'ocf::packages::firefox::apt':
-    stage => first,
-  }
-
-  # TODO: switch to mozilla.debian.net once they support stretch.
```
jvperrin: Do we still want to switch to this sometime? I suppose we'd rather use the upstream version

kpengboy: Well, they're not going to support it until at least buster.

kpengboy: I guess we can add A TODO for when buster comes out...

jvperrin: Hmm, alright

kpengboy:
Any other comments? If not I will merge

PR #190, opened 2017-09-08
Avert crisis

tahabi:
since this is causing more problems than it's currently worth until DNS gets fixed


jvperrin:
I'm don't think this would do it, since it would still be running Mesos, probably best to just run it on your own branch for now and shut off the Mesos agent on crisis manually? Assuming it's a short-term fix anyway, that seems easiest to me, otherwise you'd need to make sure it didn't have all the Mesos stuff any more through puppet, which would be a bit of a pain.

PR #191, opened 2017-09-09
ocf_mail: count each vhost mail address separately

kpengboy:


modules/ocf_mail/files/spam/logging/examine-mail-log
```diff
@@ -22,6 +23,12 @@ USER_EXCEPTIONS = {'root', 'apache', 'wordpress'}
 SYSTEM_USERS = {'ocfweb'}
 
 
+# vhost to user dictionary
+vhost_user_for_domain = {}
+for _vhost in get_mail_vhosts():
+    vhost_user_for_domain[_vhost.domain] = _vhost.user
```
chriskuehl: i'd rather not do this at the top level, imports should be as free as possible and this is slow

modules/ocf_mail/files/spam/logging/examine-mail-log
```diff
@@ -115,14 +115,30 @@ def sending_too_much_mail(entry):
...
 
     for mail in map(json.loads, sys.stdin):
+        from_name, from_addr = mail['from']
+        from_domain = extract_domain(from_addr)
+
         uid = mail.get('uid')
 
-        if uid:
+        if is_vhost_domain(from_domain):
+            vhost_mail_count[from_addr] += 1
```
chriskuehl: ideally it would only apply this to mail actually sent by vhost users, as-is doesn't this also apply to mail sent e.g. via PHP?

i'd be worried about spam that chooses a random from address on the vhost for each message which would never get flagged here

kpengboy:
Any more comments? If not I will merge

matthew-mcallister:
Judging from the diff and @chriskuehl's comments this looks fine, but I don't know if anyone realized you pushed more changes until now (I didn't).

kpengboy:
Strange, does GitHub not send more emails if you replace a commit by force push?

modules/ocf_mail/files/spam/logging/examine-mail-log
```diff
@@ -50,28 +50,12 @@ def extract_user(addr):
...
-    except (dns.resolver.NoAnswer, dns.resolver.NXDOMAIN):
-        return False
-
-
 def examine_mail(mail):
     """Examine a JSON dict of mail, yielding problem tuples."""
 
     from_name, from_addr = mail['from']
     from_domain = extract_domain(from_addr)
+    sender = mail.get('sender')
```
chriskuehl: someone is going to look at this a year from now and have no idea what the difference between "from" and "sender" are.

can we call this "authenticated_sender" or "smtp_user" or something less ambiguous?

modules/ocf_mail/files/spam/logging/examine-mail-log
```diff
@@ -87,10 +71,13 @@ def examine_mail(mail):
...
                 if from_user != user and from_user not in USER_EXCEPTIONS:
                     if not is_staff(user) or from_user != 'help':
-                        yield "From @ocf but user doesn't match", \
-                            'from {}, should be {} (uid {})'.format(from_user, user, uid)
+                        yield ("From @ocf but user doesn't match",
+                               'from {}, should be {} (uid {})'.format(from_user, user, uid))
             else:
                 yield "From @ocf but user doesn't make sense", 'from ' + from_addr
+        elif sender and sender != from_addr:
+            yield ("From address doesn't match sender address",
```
chriskuehl: again, not very clear. "From header doesn't match authenticated sender" maybe?

matthew-mcallister:
Sorry, my mistake. It does, I just didn't realize I got two emails and assumed you pushed to a fork.

matthew-mcallister:
For the record, I opine that it would generally be a better practice to comment or push a new commit when updating a PR. It's easy to miss when you get a near-duplicate email since you have to check the timestamp unless you already remember reading the original.

chriskuehl:
Another benefit of pushing a new commit after each revision (not rebasing) is that GitHub's pull request UI sucks and still can't show you the differences between revisions when rebasing is involved :(

I typically push new revisions as commits to keep it easy on the reviewers, and then optionally clean up (if necessary) before merging.

matthew-mcallister:
+1 to @chriskuehl. Rebasing too early is actually a bad habit of mine.

kpengboy:
On Sep 27, 2017 3:49 PM, "Chris Kuehl" <notifications@github.com> wrote:

Another benefit of pushing a new commit after each revision (not rebasing)
is that GitHub's pull request UI sucks and still can't show you the
differences between revisions when rebasing is involved :(

This is the one thing I like more about Gerrit.

I typically push new revisions as commits to keep it easy on the reviewers,
and then optionally clean up (if necessary) before merging.

That's a good suggestion, I will try next time.

PR #193, opened 2017-09-15
Convert rancid from cvs to git, remove dnssec logs

jvperrin:
These are the changes I've made to pestilence on stretch, it also backs up rancid config diffs to GitHub, which is pretty nice to have, especially since it strips the password (I didn't know it did that). `rancid-run` already attempts to push to a remote when it is run, so that part doesn't have to be added like it does in the `ocf_ldap` module for GitHub backups.

I removed the `/var/log/dnssec` log because `bind9` wouldn't start without it and it didn't even seem to be logging anything, so I figured keeping it didn't make sense.

### DIFF
--- a/modules/ocf_rancid/manifests/init.pp
+++ b/modules/ocf_rancid/manifests/init.pp
@@ -13,27 +13,18 @@ class ocf_rancid {
       unless  => 'grep -q ^FILTER_PWDS /etc/rancid/rancid.conf',
       require => Package['rancid'];
 
-    # idempotent command to update rancid cvs groups
-    'rancid-cvs-update':
-      command     => '/var/lib/rancid/bin/rancid-cvs',
-      user        => 'rancid',
-      refreshonly => true,
-      require     => Package['rancid'],
-      subscribe   => [
-        Package['rancid'],
-        Exec['rancid-add-ocf-group'],
-      ];
+    'rancid-set-vcs-git':
+      command => 'sed -i s/RCSSYS=cvs/RCSSYS=git/ /etc/rancid/rancid.conf',
+      unless  => 'grep -q ^RCSSYS=git /etc/rancid/rancid.conf',
+      require => Package['rancid'];
   }
 
   file {
     '/var/lib/rancid/ocf/router.db':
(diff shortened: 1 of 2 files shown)


modules/ocf_rancid/manifests/init.pp
```diff
@@ -49,4 +40,35 @@
...
     require => Package['rancid'],
   }
+
+  # Delete all logs older than 14 days.
+  # This would be better with logrotate, but the log files contain the date in
+  # their name, which logrotate doesn't handle very well. The files aren't large
+  # at all, so they don't need to be compressed, there's just a lot of them
+  # since they are created once for every hour.
+  cron { 'clean-rancid-logs':
+    command => 'find /var/log/rancid/* -mtime +14 > /dev/null',
```
chriskuehl: i like how i read this, thought "wtf use logrotate", then read your comment above and saw you pre-empted me :P

modules/ocf_rancid/manifests/init.pp
```diff
@@ -49,4 +40,35 @@
...
+    user    => 'rancid',
+    special => 'hourly',
+    require => Package['rancid'],
+  }
+
+  # Don't push to GitHub for dev-* hosts to prevent duplicate backups with
+  # different commit hashes
+  if $::hostname !~ /^dev-/ {
```
chriskuehl: we should really turn this into a global/hiera variable rather than repeating this dev- hostname junk everywhere

maybe a good starter ticket

modules/ocf_rancid/manifests/init.pp
jvperrin: Haha yeah, I tried logrotate first but couldn't find a good way to do that didn't involve adding a find command somewhere anyway. If we needed compression too I would have done that, but we just want a bit of cleanup here.

modules/ocf_rancid/manifests/init.pp
jvperrin: Good idea, I like the idea of making this a starter ticket because it wouldn't be too hard and would touch a few different parts of puppet, so that would be cool

PR #195, opened 2017-09-19
Enable PulseAudio SSH tunneling to TV

matthew-mcallister:
This enables connecting to Pulse via TCP on the desktops and tells Pulse
to listen for TCP connections on localhost

### DIFF
--- a/modules/ocf_tv/manifests/pulse.pp
+++ b/modules/ocf_tv/manifests/pulse.pp
@@ -1,0 +1,11 @@
+class ocf_tv::pulse {
+  include ocf::packages::pulse;
+
+  # Make pulse listen on localhost so desktops can SSH tunnel to it
+  $tcp_module = 'load-module module-native-protocol-tcp listen=127.0.0.1 port=4713 auth-anonymous=1'
+  exec { 'enable-pulse-tcp-listen':
+    command => "echo '${tcp_module}' >> /etc/pulse/default.pa",
+    unless  => "grep -q '^${tcp_module}$' /etc/pulse/default.pa",
+    require => Package['pulseaudio'],
+  }
+}
--- a/modules/ocf/manifests/packages/pulse.pp
+++ b/modules/ocf/manifests/packages/pulse.pp
@@ -1,5 +1,4 @@
-class ocf_desktop::pulse {
-
+class ocf::packages::pulse {
   ocf::repackage {
     # PulseAudio without recommends
     'pulseaudio':
@@ -10,7 +9,7 @@ class ocf_desktop::pulse {
       require    => Ocf::Repackage['pulseaudio'];
   }
 
-  # ALSA ultilities and PulseAudio plugin
+  # ALSA utilities and PulseAudio plugin
   package { [ 'alsa-utils', 'libasound2-plugins' ]: }
 
   # route ALSA sound through PulseAudio
@@ -18,5 +17,4 @@ class ocf_desktop::pulse {
     source  => 'puppet:///modules/ocf_desktop/asound.conf',
     require => Package['libasound2-plugins'],
   }
-
 }
--- a/modules/ocf_desktop/manifests/init.pp
+++ b/modules/ocf_desktop/manifests/init.pp
@@ -3,6 +3,7 @@ class ocf_desktop {
   include ocf::packages::chrome
   include ocf::packages::cups
   include ocf::packages::firefox
+  include ocf::packages::pulse
 
   include ocf_mesos::slave
 
@@ -12,7 +13,6 @@ class ocf_desktop {
   include ocf_desktop::grub
   include ocf_desktop::modprobe
   include ocf_desktop::packages
-  include ocf_desktop::pulse
   include ocf_desktop::sshfs
   include ocf_desktop::stats
   include ocf_desktop::steam
--- a/modules/ocf_tv/manifests/init.pp
+++ b/modules/ocf_tv/manifests/init.pp
@@ -1,4 +1,6 @@
 class ocf_tv {
+  include ocf_tv::pulse
+
   include ocf::packages::chrome
   include ocf::packages::firefox
   include ocf_desktop::drivers
@@ -11,8 +13,6 @@ class ocf_tv {
       'feh',
       'i3',
       'nodm',
-      'pavucontrol',
-      'pulseaudio',
       'vlc',
       'x11vnc',
       'xinit',


modules/ocf_tv/manifests/pulse.pp
```diff
@@ -0,0 +1,10 @@
+class ocf_tv::pulse {
+  include ocf::packages::pulse;
+
+  # Make pulse listen on 127.0.0.1:4713 desktops can SSH tunnel to it
+  exec { 'enable-pulse-tcp-listen':
+    command => 'echo "load-module module-native-protocol-tcp auth-ip-acl=127.0.0.1 auth-anonymous=1" >> /etc/pulse/default.pa',
```
tahabi: I'd probably just puppet the entire default.pa file as opposed to doing this, seems cleaner

modules/ocf_desktop/manifests/pulse.pp
```diff
@@ -1,22 +1,10 @@
...
-    # PulseAudio without recommends
-    'pulseaudio':
-      recommends => false;
-    # PulseAudio graphical interfaces
-    [ 'paprefs', 'pavucontrol' ]:
-      recommends => false,
-      require    => Ocf::Repackage['pulseaudio'];
+  # Let Pulse connect over TCP for tunneling audio to the TV
+  exec { 'enable-pulse-tcp':
+    command => 'echo "load-module module-native-protocol-tcp" >> /etc/pulse/default.pa',
```
tahabi: this is also probably better done in `ocf-tv` - we already load `module-zeroconf-discover` on the fly, may as well load this one as well.

matthew-mcallister: Sounds good, you can replace `zeroconf-discover` with `native-protocol-tcp`.

modules/ocf_tv/manifests/pulse.pp
matthew-mcallister: Have you seen the entire file? I don't think we should commit that whole thing to add just one line. We already do this elswere. We already use this pattern in a few other places, e.g.
https://github.com/ocf/puppet/blob/master/modules/ocf_desktop/manifests/sshfs.pp#L20-L24
https://github.com/ocf/puppet/blob/master/modules/ocf_rancid/manifests/init.pp#L5-L20

tahabi: is there a way to use augeas for this? e.g. https://github.com/tahabi/puppet/blob/15b16e5f80c0dc1a165ab8d1abcaa4f4116ecdf9/modules/ocf_desktop/manifests/pulse.pp#L24

matthew-mcallister: Yep, here's what it looks like:
```puppet
$tcp_module = 'module-native-protocol-tcp auth-ip-acl=127.0.0.1 auth-anonymous=1'
augeas { 'enable-pulse-tcp':
  incl    => '/etc/pulse/default.pa',
  lens    => 'Spacevars.lns',
  changes => "set load-module[. = '${tcp_module}'] '${tcp_module}'",
}
```
It took me about an hour to learn how to write this and the only thing that makes it look cleaner is the fact that I assigned the setting to a variable. For one-liners like this, I think augeas hurts readability more than it helps.

tahabi: Lol, nice. Apologies for making you go through the effort.

I guess I'll update `ocf-tv` to reflect the changes made here and hopefully that results in a more secure but equally reliable way to tunnel audio to the TV.

tahabi:
ok so getting this to work with `ocf-tv` is as simple as calling `pactl load-module module-native-protocol-tcp` and 'pactl load-module module-tunnel-sink-new server=tcp:tv:4713 sink_name=remote` but there is still no SSH or authentication necessary to start tunneling audio.

matthew-mcallister:
Presumably Pulse is still listening for external connections due to Avahi
or whatever was being used before.

tahabi:
I specifically unloaded the Avahi modules and restarted pulse to make sure they weren't loaded by using `pactl list modules | grep zeroconf`. I think loading `module-native-protocol-tcp` is all that's necessary to get pulse to listen on 4713.

matthew-mcallister:
Well yes, but that's why this commit sets it to listen on localhost only on
the TV.

tahabi:
What I'm saying is, I made these same changes to the TV, e.g. `pactl load-module module-native-protocol-tcp auth-ip-acl=127.0.0.1 auth-anonymous=1`, and that doesn't actually do what it seems like it should.

matthew-mcallister:
Sorry, but I can't replicate what you're describing. Set the desired line in `default.pa`...
```
load-module module-native-protocol-tcp port=<whatever> listen=<whatever> auth-ip-acl=<ditto> auth-anonymous=<0 or 1>
```
... restart pulse...
```
sudo -u ocftv pulseaudio -k; sudo -u ocftv pulseaudio --start --log-target=syslog
```
... and then `nmap -p<whatever> tv localhost` and I get just what I was expecting.

tahabi:
testing this over the last few days has been extremely tedious, but I have a version of `ocf-tv` that seems to work most of the time with the following line in `default.pa`:

    load-module module-native-protocol-tcp listen=127.0.0.1 port=4713 auth-anonymous=1

by doing `ssh -L <port>:localhost:4713 tv` before `pactl load-module module-tunnel-sink-new server=tcp:localhost:<port> sink_name=remote`

nmaping tornado with that line in `default.pa` and trying to create the tunnel sink without the SSH tunnel both fail, so the security issue should be resolved as well.

matthew-mcallister:
Nice, that sounds like exactly what we want. I'm not entirely sure if `pactl load-module` is idempotent, so perhaps that's why you experienced some inconsistencies?

Anyways, is there anything else that needs to be done here, or do you think this is ready to merge? Maybe disable Avahi?

tahabi:
`load-module` isn't idempotent, but the more annoying problem is that the Pulse daemon on the TV likes to crash randomly, and occasionally manages to restart with multiple daemons running at the same time, which breaks all audio and requires manual intervention to fix that isn't as simple as killing all but one of the daemons. @sahilhasan can regale you with stories of the most recent episode, which required a deep purge of pulse and all associated packages and config files before audio would work again. 

pactl on the local client also has several problems, in that occasionally it will appear to return success for a `load-module` call but actually fail behind the scenes, or when it explicitly fails, it does so with a worthless error message, and since pulse doesn't start with debug output by default, restarting the local pulse server with debugging output often causes its own set of problems that makes figuring out why the original `load-module` call failed incredibly tedious.

In any case, it should be ready. Zeroconf was uninstalled in the first of several purges by myself and sahil while we were trying to fix pulseaudio. Nice work in figuring out all of this.

matthew-mcallister:
Great, I removed the Pulse ACL (which was, as you noticed, unnecessary) and squashed. Also thanks for testing and updating `ocf-tv` in response to this.

PR #197, opened 2017-09-20
ocf_www: Disable Apache OPTIONS requests

matthew-mcallister:
We could potentially re-enable these requests once [CVE-2017-9798](https://security-tracker.debian.org/tracker/CVE-2017-9798) is patched, but I don't see a good reason to.

### DIFF
--- a/modules/ocf_www/manifests/mod/security.pp
+++ b/modules/ocf_www/manifests/mod/security.pp
@@ -1,0 +1,6 @@
+class ocf_www::mod::security {
+  class { '::apache::mod::security':
+    # Disable OPTIONS requests, originally due to CVE-2017-9798
+    allowed_methods => 'GET HEAD POST',
+  }
+}
--- a/modules/ocf_www/manifests/mod/security.pp
+++ b/modules/ocf_www/manifests/mod/security.pp
@@ -1,6 +1,6 @@
 class ocf_www::mod::security {
   class { '::apache::mod::security':
     # Disable OPTIONS requests, originally due to CVE-2017-9798
-    allowed_methods => 'GET HEAD POST',
+    allowed_methods => 'GET HEAD POST PUT DELETE PATCH',
   }
 }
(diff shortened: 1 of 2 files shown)


modules/ocf_www/manifests/mod/security.pp
```diff
@@ -0,0 +1,6 @@
+class ocf_www::mod::security {
+  class { '::apache::mod::security':
+    # Disable OPTIONS requests, originally due to CVE-2017-9798
+    allowed_methods => 'GET HEAD POST',
```
jvperrin: I think we should still allow `PUT`, `DELETE`, and `PATCH` at least, although I'm not sure how often they are used. I do know Rails uses them a fair amount anyway.

matthew-mcallister: Should we? I actually don't know what those do. All I did was remove "OPTIONS" from the list of defaults [here](https://github.com/puppetlabs/puppetlabs-apache#class-apachemodsecurity). Can you check if any rails websites are actually broken? If not, I'll just add them in anyways.

jvperrin: I don't know of any Rails websites not on apphosting, but this would probably break things on them, although I believe they have fallback methods too (just using POST with more details or something), but that's mostly to get around [browsers not supporting those methods](https://stackoverflow.com/a/904013) back when they actually didn't have proper support.

daradib:
Thanks! This will break [cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) (`Access-Control-Allow-Origin` header), i.e., off-site webpages generating cross-origin requests to the OCF. I don't see a reason to support that (web fonts hosted on the OCF?). So feature not a bug.

matthew-mcallister:
Oh, we actually don't want to break CORS, since groups on apphosting can host their static files on www.

PR #198, opened 2017-09-20
Use ServerKeychain cups option instead of outdated ServerKey

dkess:
In [rt #6438](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=6438) we discovered that whiteout was using self-signed SSL certificates for its web interface. This is because it was using certs defined in /etc/cups/ssl. The ServerKey option was supposed to configure cups to use /etc/ssl/private, but that option appears to not work. This PR uses the ServerKeychain option which should work.


dkess:
Yup, it's been tested on dev-whiteout.

kpengboy:
One remaining issue is that CUPS isn't serving the intermediate certs.

PR #199, opened 2017-09-26
Remove economode textbox from cups dialog

dkess:
Tested on dev-whiteout-- doesn't crash cups. Based on Dara's suggestion in [rt/6568](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=6568).


dkess:
This appears to have worked-- the option is gone from the print dialog, and printing still works.

tahabi:
+1 test in prod

PR #200, opened 2017-10-01
Use more DNS servers in case of outages

matthew-mcallister:
This adds two more university servers, srb-infoblox-nsm1.berkeley.edu and sut-infoblox-nsm2.berkeley.edu, for our nameserver to forward requests to, since they stayed up during the last outage. Further, this sets hosts to fall back to public DNS (Google) if either our or university DNS goes down.


modules/ocf/manifests/networking.pp
```diff
@@ -10,7 +10,7 @@
     $gateway6   = '2607:f140:8801::1',
 
     $domain      = 'ocf.berkeley.edu',
-    $nameservers = ['2607:f140:8801::1:22', '169.229.226.22', '169.229.2.171'],
+    $nameservers = ['2607:f140:8801::1:22', '169.229.226.22', '8.8.8.8'],
```
jvperrin: Might be good to mention here that this can only store 3 values at maximum, for anyone changing this in the future?

matthew-mcallister: I added an assertion to enforce this.

jvperrin: Nice, even better.

PR #202, opened 2017-10-04
pdf-open default

kkuehler:
As requested by @sahilhasan, uses a modified version of @chriskuehl's pdf-open script (https://github.com/ocf/utils/blob/master/desktop/pdf-open) to be the default for pdf, ps, and related file types. This is to reduce printing issues detailed in ticket 4839 on RT. I moved the script from utils to puppet because, as the default, there is no point in executing it in a terminal. The modifications to the original script address the issue where the filename and title take that of the temp file, which may confuse some people.

### DIFF
--- a/modules/ocf_desktop/files/xsession/pdf-open
+++ b/modules/ocf_desktop/files/xsession/pdf-open
@@ -1,22 +1,35 @@
-#!/bin/bash -e
-[ "$#" -ne 1 ] && { echo 'Bad arguments, trying evince...'; exec evince "$1"; }
+#!/bin/bash
+set -euo pipefail
+
+[ "$#" -ne 1 ] && { echo 'Bad arguments, trying evince...'; exec evince "$@"; }
 [ -f "$1" ] || { echo 'File does not exist, trying evince...'; exec evince "$1"; }
 
 size=$(du "$1") || { echo 'Cannot get size, trying evince...'; exec evince "$1"; }
 size=$(cut -f1 <<< "$size")
 [ "$size" -lt 5120 ] || { echo 'Too large, trying evince...'; exec evince "$1"; }
 
-echo 'Rasterizing PDF...'
-filename=$(basename $1)
-tmp="/tmp/ocf_$filename"
+len=$(pdftk "$1" dump_data | awk '/NumberOfPages/{print $2}')
+[ -z "$len" ] || [ "$len" -gt 75 ] && { echo 'Too long, trying evince...'; exec evince "$1"; }
 
-if ! convert -density 250 "$1" "$tmp" | zenity --progress --pulsate --auto-close; then
-    rm "$tmp"
+filename=$(basename "$1")
+tmpdir=$(mktemp -d)
+tmp="$tmpdir/$filename"
+
+convert -density 250 "$1" "$tmp" |
+zenity --progress \
+	--pulsate \
+	--auto-close \
+	--auto-kill \
+	--text="Rasterizing File..."
+
+if [ $? -ne 0 ]; then
+    rm -rf "$tmpdir"
     echo 'Convert failed, trying evince...'
     exec evince "$1"
 fi
 
 echo 'Done rasterizing, now opening with evince...'
-exiftool -tagsfromfile "$1" -title "$tmp"
+exiftool -tagsfromfile "$1" -title "$tmp" || true
 echo "(output file: $tmp)"
-exec evince "$tmp"
+evince "$tmp"
+rm -rf "$tmpdir"
--- a/modules/ocf_desktop/files/xsession/pdf-open
+++ b/modules/ocf_desktop/files/xsession/pdf-open
@@ -1,0 +1,23 @@
+#!/bin/bash -e
+[ "$#" -ne 1 ] && { echo 'Bad arguments, trying evince...'; exec evince "$@"; }
+[ -f "$1" ] || { echo 'File does not exist, trying evince...'; exec evince "$@"; }
+
+size=$(du "$1") || { echo 'Cannot get size, trying evince...'; exec evince "$@"; }
+size=$(cut -f1 <<< "$size")
+[ "$size" -lt 5120 ] || { echo 'Too large, trying evince...'; exec evince "$@"; }
+
+echo 'Rasterizing PDF...'
+tmp=$(mktemp --suffix='.pdf')
+
+if ! convert -density 250 "$1" "$tmp"; then
+    rm "$tmp"
+    echo 'Convert failed, trying evince...'
+    exec evince "$@"
+fi
+
+echo 'Done rasterizing, now overwriting unrasterized version...'
+exiftool -tagsfromfile "$@" -title "$tmp"
+mv "$tmp" "$@"
+
+echo 'Opening with evince...'
+exec evince "$@"
--- a/modules/ocf_desktop/files/xsession/pdf-open
+++ b/modules/ocf_desktop/files/xsession/pdf-open
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -euo pipefail
+set -uo pipefail
 
 [ "$#" -ne 1 ] && { echo 'Bad arguments, trying evince...'; exec evince "$@"; }
 [ -f "$1" ] || { echo 'File does not exist, trying evince...'; exec evince "$1"; }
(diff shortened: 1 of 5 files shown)


modules/ocf_desktop/manifests/defaults.pp
```diff
@@ -4,10 +4,23 @@
...
+  }
+
   # /etc/alternatives/
   if $::lsbdistcodename == 'jessie' {
     exec { 'update-alternatives --set x-terminal-emulator /usr/bin/xfce4-terminal.wrapper':
       unless => '/usr/bin/test $(readlink /etc/alternatives/x-terminal-emulator) == "/usr/bin/xfce4-terminal.wrapper"';
     }
   }
+
+  package { 'libimage-exiftool-perl':; }
```
tahabi: perhaps move this to ocf_desktop/manifests/packages.pp?

modules/ocf_desktop/manifests/defaults.pp
```diff
@@ -4,10 +4,23 @@
     source  => 'puppet:///modules/ocf_desktop/xsession/mimeapps.list';
   }
 
+  file { '/opt/share/puppet/pdf-open':
```
chriskuehl: should go in `/usr/local/bin` probably, i don't think we should encourage use of `/opt/share/puppet` (as much as possible), it's sort of a dumping ground

modules/ocf_desktop/files/xsession/pdf-open
```diff
@@ -0,0 +1,23 @@
...
+
+if ! convert -density 250 "$1" "$tmp"; then
+    rm "$tmp"
+    echo 'Convert failed, trying evince...'
+    exec evince "$@"
+fi
+
+echo 'Done rasterizing, now overwriting unrasterized version...'
+exiftool -tagsfromfile "$@" -title "$tmp"
+mv "$tmp" "$@"
```
chriskuehl: i don't think we should ever overwrite peoples' PDFs just because they opened it in (what they thought was) evince

we don't know what they're going to do with it afterwards -- printing is just one possible thing they might do. maybe they're just opening it before posting on a website or something. i'd be pretty annoyed if opening "evince" resulted in my files being overwritten with a rasterized version

modules/ocf_desktop/files/xsession/pdf-open
```diff
@@ -0,0 +1,23 @@
...
+tmp=$(mktemp --suffix='.pdf')
+
+if ! convert -density 250 "$1" "$tmp"; then
+    rm "$tmp"
+    echo 'Convert failed, trying evince...'
+    exec evince "$@"
+fi
+
+echo 'Done rasterizing, now overwriting unrasterized version...'
+exiftool -tagsfromfile "$@" -title "$tmp"
```
kpengboy: Check that this script only gets one argument while you're at it?

In any case you'd probably want `$1` here as `$@` is misleading.

modules/ocf_desktop/files/xsession/pdf-open
```diff
@@ -0,0 +1,23 @@
...
+[ "$#" -ne 1 ] && { echo 'Bad arguments, trying evince...'; exec evince "$@"; }
+[ -f "$1" ] || { echo 'File does not exist, trying evince...'; exec evince "$@"; }
+
+size=$(du "$1") || { echo 'Cannot get size, trying evince...'; exec evince "$@"; }
+size=$(cut -f1 <<< "$size")
+[ "$size" -lt 5120 ] || { echo 'Too large, trying evince...'; exec evince "$@"; }
+
+echo 'Rasterizing PDF...'
+tmp=$(mktemp --suffix='.pdf')
+
```
kpengboy: Since this script usually won't be run in a terminal, consider making a progress dialog using zenity. Otherwise people will wonder why their PDFs haven't opened yet and try to open it 10 times.

kkuehler:
@kpengboy, @chriskuehl changes made.

chriskuehl:
Is there any reason we don't just put the rasterization in the printing pipeline? This is going to break lots of things, e.g. make it impossible to copy-paste text out of PDFs, make it impossible to Ctrl-F to search in PDFs, break PDF forms, etc.

I wrote a branch at one point that did this: https://github.com/ocf/puppet/compare/rasterize-before-printing

kpengboy:
Similar solutions have been tried, but so far a PDF which rasterizes reasonably with pdf-open will produce a still unacceptably slow PDF when you try to rasterize it in the print pipeline.

With further experimentation it could maybe be made to work...

sahilhasan:
@chriskuehl 
So far any solutions like that as @kpengboy mentioned have been rather ridiculously slow. That being said,  imagemagick and pdftops both use ghostscript internally, so maybe in theory we could optimize by using ghostscript directly. Through various testing that @kkuehler and myself have been doing for the last few days, it seems as if rasterization can at best be done in something closely resembling polynomial time, so I'm not too confident in rasterizing within the printing pipeline itself. As it is that pipeline can feel bloated for users in the lab (given how many people I see loitering around the printers, and how many times people ask me why their stuff isn't printing when it prints 30 seconds later)

I think its also important to remember that people can still open evince WITHOUT rasterizing their pdf, and so it is not actually impossible to do most of the things that you mentioned, albeit more difficult and not as intuitive.  Additionally, two things:

1. This is not meant to be a catch-all permanent fix, but something that provides the OCF with some level of security against their printers being nuked until we can come up with something more optimized. Testing between @kkuehler and myself has shown that inserting this into the printing pipeline efficiently is not a trivial task. As it is I prefer a slightly dirty temporary bandage that can be improved on to having to reset our printers 7/8 times in one day (for the sake of clarity, this is the max number of times its happened, not the norm. On average it does happen once or twice a day though) and really would love to not have the opstaff perpetually attempt to police our printqueue to try to catch econ jobs. 

2. I think its far more reasonable to tell people that if they want to avoid rasterizing their pdf they have to right click on it and select document viewer rather than telling them they cannot print their econ pdfs or that they have to wait for a staff member to come help them. At worst, we slightly inconvenience the small subset of people who want to edit their pdfs at the lab or copy paste text. Personally, I think that tradeoff is worth it, because currently a single econ print job reaching the printer usually requires 2-3 refunds/free opstaff prints, a 2-5 minute delay on all print jobs, and in general a larger scale of congestion in the lab. 

I (and I'm sure both Kevins agree with me on this) have no intention of letting this be this be a final solution. We will improve on it over time, but I argue a temporary solution that can be improved upon is much better than no solution at all for the time being.

chriskuehl:
Is a better temporary fix just writing a cronjob that kills stuck jobs via the printer web UI (if the printer is stuck) or via the cups API (if it's stuck processing in CUPS)? Or do the printers get so stuck that we actually need to physically power cycle them?

sahilhasan:
AFAIK we have to physically power cycle them every time unless you actually go into the printer admin panel and cancel the stuck job there (i.e. not in CUPS), and even then you have to be in the admin panel rather promptly to prevent it. @kpengboy probably knows more.

chriskuehl:
Yeah, the printer admin panel is accessible via the web, right? I bet we could write something to send that request automatically pretty easy.

sahilhasan:
@chriskuehl but that still kind of avoids the problem of people printing econ pdfs, wouldn't a solution where people can still print their pdfs, albeit slowly, be better?

kpengboy:
When the printer gets a bad PDF the web interface becomes unusably slow.

On Oct 6, 2017 3:40 PM, "Chris Kuehl" <notifications@github.com> wrote:

Yeah, the printer admin panel is accessible via the web, right? I bet we
could write something to send that request automatically pretty easy.

—
You are receiving this because you were mentioned.

<https://github.com/ocf/puppet/pull/202#issuecomment-334886525>, or mute
the thread

chriskuehl:
:( that is unfortunate

sahilhasan:
I think this is probably the best temp fix we're going to have. Unless anyone has any issues with the code itself, I would say lets use it.

kpengboy:
I tried Chris's solution again and, with one page, it still takes the
printers ~3 min to process. pdf-open'ed files on the other hand are very
fast.

chriskuehl:
Three minutes doing what? Is pdf-open not the same as the branch I linked above, just happening at a different stage? Something is up with those numbers.

kpengboy:
I tested it on dev-werewolves with this patch, adapted from what you linked
above. Let me know if I rebased it wrong.

```
diff --git a/modules/ocf_printhost/files/ocfps
b/modules/ocf_printhost/files/ocfps
index 9eeccfc..f34c1d4 100755
--- a/modules/ocf_printhost/files/ocfps
+++ b/modules/ocf_printhost/files/ocfps
@@ -31,7 +31,7 @@ if [ "$#" -ne 5 ] && [ "$#" -ne 6 ]; then
 fi

 if [ "$#" -eq 5 ]; then
-    $PDFTOPS - - | $PSTOPS "$@"
+    $PDFTOPS - - | convert -density 300 -compress Group4 ps:- ps:- |
$PSTOPS "$@"
 else
-    $PDFTOPS "$6" - | $PSTOPS "${@:1:5}"
+    $PDFTOPS "$6" - | convert -density 300 -compress Group4 ps:- ps:- |
$PSTOPS "${@:1:5}"
 fi
```

After applying this, it takes the printers ~3min to process a single page
from one of those Econ slides, once the page is sent to the printers.

The key difference is that the PDF data that Evince sends to the printer is
not the same as the PDF data that Evince reads in.

kpengboy:
@kkuehler can you rebase this?

modules/ocf_desktop/files/xsession/pdf-open
```diff
@@ -0,0 +1,35 @@
...
+tmp="$tmpdir/$filename"
+
+convert -density 250 "$1" "$tmp" |
+zenity --progress \
+	--pulsate \
+	--auto-close \
+	--auto-kill \
+	--text="Rasterizing File..."
+
+if [ $? -ne 0 ]; then
```
kpengboy: With `set -e`, this check is superfluous because the script will bomb out if `convert` fails.

I now think that, while we should generally use `set -euo pipefail`, in this particular script the `e` part of it is counterproductive and we'd do better without it.

modules/ocf_desktop/files/xsession/pdf-open
```diff
@@ -0,0 +1,35 @@
...
+    rm -rf "$tmpdir"
+    echo 'Convert failed, trying evince...'
+    exec evince "$1"
+fi
+
+echo 'Done rasterizing, now opening with evince...'
+exiftool -tagsfromfile "$1" -title "$tmp" || true
```
kpengboy: As a side note, if you do choose to remove the `-e` flag, you won't need the `|| true` anymore.

kkuehler:
@kpengboy Changes made + rebased with master.

modules/ocf_desktop/manifests/defaults.pp
```diff
@@ -4,10 +4,22 @@
...
+    require => File['/usr/local/bin/pdf-open'];
+  }
+
   # /etc/alternatives/
   if $::lsbdistcodename == 'jessie' {
     exec { 'update-alternatives --set x-terminal-emulator /usr/bin/xfce4-terminal.wrapper':
       unless => '/usr/bin/test $(readlink /etc/alternatives/x-terminal-emulator) == "/usr/bin/xfce4-terminal.wrapper"';
     }
   }
+
```
kpengboy: Extra space

kkuehler: done...

PR #204, opened 2017-10-07
Make sure hooks are installed when people run `make`

kpengboy:
Thoughts?


kpengboy:
Actually what do you think about making a `.git/hooks/pre-commit` rule, and making `install-hooks` a shim that depends on that?

PR #205, opened 2017-10-10
Add Let's Encrypt interface and use for www/apps

matthew-mcallister:
Among minor bug fixes and superficial changes, this branch adds a script `ocf-lets-encrypt` which implements the interface proposed in [this doc](https://docs.google.com/document/d/1teNCynslk71N0UU-O_GiSbCMBLZkAvL8SEot4RN1U5E). It also configures `www`/`apphost` to defer acquiring certs to the script, so it will be put to use right away.

I'll provide the puppet interface for invoking the script in a subsequent PR. However, I've already used it to obtain a cert for meltdown.ocf.io with a new private key, and it was incredibly easy.

The script should be run as a cron job or by another script. In the case of `www`/`apphost`, we run it as the `ocfletsencrypt` user, but on other hosts we presumably want to run as `root:ssl-cert` and put the certs in `/etc/ssl/private`, which is the existing convention.

### DIFF
--- a/modules/ocf/files/ssl/ocf-lets-encrypt
+++ b/modules/ocf/files/ssl/ocf-lets-encrypt
@@ -1,23 +1,22 @@
 #!/usr/bin/env python3
 """Update or acquire Let's Encrypt certificates for a host.
 
-This script is triggered every time Puppet runs with the certs needed by
-the host in the arguments. It is also triggered by other scripts, e.g.
+This script is triggered every time Puppet runs with the host's domain
+name(s) in the arguments. It is also triggered by other scripts, e.g.
 the Let's Encrypt scripts for vhosts on www and apphost.
 
 The script takes the following basic steps to acquire a certificate:
 
-    1. Checks whether we have already acquired a certificate in the past
-       for this domain.
+    1. Checks if we have already acquired this certificate in the past.
 
     2. If we already have a cert, further checks the expiration date to
        see if it is expired or going to expire within a few weeks.
 
     3. If the answer to (1) or (2) is "no", we invoke acme-tiny to
-       acquire/renew a cert for the domain.
+       acquire/renew the cert.
 
 Sounds simple enough? Well, there's more to it. acme-tiny proves that we
-actually own the domain by writing a challenge file and expecting it to
+actually own the domains by writing a challenge file and expecting it to
 be served under http://<fqdn>/.well-known/acme-challenge/. This presents
 two problems: 1) on hosts that don't run a web server (e.g. mail), we
 need something to serve that file, and 2) on hosts that do run a web
@@ -51,19 +50,17 @@ The return code indicates whether the script did anything, so the caller
 can reload services as necessary:
 
     0       nothing changed
-    255     certs added/changed
+    255     cert renewed/acquired
     else    error
 """
 import argparse
 import functools
 import os
 import re
-import shutil
 import socket
 import subprocess
 import sys
 import tempfile
-from collections import namedtuple
 from contextlib import contextmanager
 from datetime import datetime
 from datetime import timedelta
@@ -71,16 +68,13 @@ from datetime import timezone
 from pathlib import Path
 
 import dateutil.parser
+from OpenSSL import crypto
 
 
 RENEW_WHEN_TIME_REMAINING = timedelta(days=45)
 
-DEFAULT_CERT_STORAGE_DIR = '/etc/ssl/private/'
 LE_ACCOUNT_KEY = Path('/etc/ssl/lets-encrypt/le-account.key')
 ACME_BASE_DIR = Path('/var/lib/lets-encrypt/')
-
-
-Request = namedtuple('Request', ('domain', 'cert_path', 'has_cert'))
 
 
 def debug(*args, **kwargs):
@@ -88,32 +82,48 @@ def debug(*args, **kwargs):
 
 
 def expiration_date(cert_path):
-    output = subprocess.check_output((
-        'openssl', 'x509', '-enddate', '-noout',
-        '-in', str(cert_path),
-    ))
-    m = re.match(b'notAfter=(.+)$', output)
-    assert m, output
-    return dateutil.parser.parse(m.group(1).decode('ascii'))
+    with cert_path.open('rb') as f:
+        cert = crypto.load_certificate(crypto.FILETYPE_PEM, f.read())
+    return dateutil.parser.parse(cert.get_notAfter())
+
+
+def cert_status(cert_path):
+    if cert_path.exists():
+        expires = expiration_date(cert_path)
+        time_remaining = expires - datetime.now(timezone.utc)
+        if time_remaining > RENEW_WHEN_TIME_REMAINING:
+            return False, time_remaining
+        else:
+            return True, time_remaining
+    else:
+        return True, None
 
 
 @contextmanager
-def make_csr(domain, key):
-    _, csr_path = tempfile.mkstemp()
-    try:
-        subprocess.check_call((
-            'openssl', 'req', '-new', '-sha256',
-            '-key', key,
-            '-subj', '/CN={}'.format(domain),
-            '-out', csr_path,
-        ))
-        yield csr_path
-    finally:
-        os.remove(csr_path)
+def make_csr(domains, key_path):
+    req = crypto.X509Req()
+    req.get_subject().CN = domains[0]
+
+    with open(key_path, 'rt') as f:
+        key = crypto.load_privatekey(crypto.FILETYPE_PEM, f.read())
+        req.set_pubkey(key)
+
+    req.add_extensions((crypto.X509Extension(
+        b'subjectAltName',
+        False,
+        bytes(', '.join('DNS:' + domain for domain in domains), encoding='ascii'),
+    ),))
+
+    req.sign(key, 'sha256')
+
+    with tempfile.NamedTemporaryFile() as csr:
+        csr.write(crypto.dump_certificate_request(crypto.FILETYPE_PEM, req))
+        csr.flush()
+        yield csr.name
 
 
-def acme_tiny(domain, key):
-    with make_csr(domain, key) as csr_path:
+def acme_tiny(domains, key_path):
+    with make_csr(domains, key_path) as csr_path:
         # TODO: use subprocess.run when we get python3.5
         proc = subprocess.Popen(
             (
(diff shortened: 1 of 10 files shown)


modules/ocf/files/ssl/ocf-lets-encrypt
```diff
@@ -0,0 +1,253 @@
...
+
+
+def expiration_date(cert_path):
+    output = subprocess.check_output((
+        'openssl', 'x509', '-enddate', '-noout',
+        '-in', str(cert_path),
+    ))
+    m = re.match(b'notAfter=(.+)$', output)
+    assert m, output
+    return dateutil.parser.parse(m.group(1).decode('ascii'))
```
kpengboy: Does it work to use `ssl.cert_time_to_seconds` here?

modules/ocf/files/ssl/ocf-lets-encrypt
```diff
@@ -0,0 +1,253 @@
...
+
+def write_cert(cert, cert_path):
+    """Write certificate file for a domain."""
+    _, temp_cert_path = tempfile.mkstemp(dir=str(cert_path.parent), prefix='.tmp')
+    try:
+        with open(temp_cert_path, 'wb') as f:
+            f.write(cert)
+        # atomic overwrite in case of errors
+        os.chmod(temp_cert_path, 0o644)
+        shutil.move(temp_cert_path, str(cert_path))
```
kpengboy: If `temp_cert_path` is guaranteed to be on the same filesystem as `cert_path` (which it must be in order for the overwrite to be actually atomic), then we should just use `os.rename` here.

modules/ocf/files/ssl/ocf-lets-encrypt
```diff
@@ -0,0 +1,253 @@
...
+            cwd=str(ACME_BASE_DIR),
+        )
+
+        # Wait for server to start up before proceeding
+        while socket.socket().connect_ex(('127.0.0.1', 80)) != 0:
+            pass
+
+        yield proc
+
+        proc.kill()
```
kpengboy: Harsh...I would use `SIGTERM` first, in case there was any cleanup the program should do...

modules/ocf/files/ssl/ocf-lets-encrypt
```diff
@@ -0,0 +1,253 @@
...
+                    request,
+                    key=args.private_key,
+                    dry_run=args.dry_run,
+                )
+
+    need_cert = {request for request in requests if not request.has_cert}
+    debug('Want cert: {}'.format(len(need_cert)))
+    debug('Already have cert: {}'.format(len(args.domains) - len(need_cert)))
+
+    return 255 if requests else 0
```
kpengboy: Uh, why's it an error if there are requests?

Also 255 tends to be a special exit code. Try picking another, like 3?

modules/ocf/files/ssl/ocf-lets-encrypt
```diff
@@ -0,0 +1,253 @@
...
+        '-in', str(cert_path),
+    ))
+    m = re.match(b'notAfter=(.+)$', output)
+    assert m, output
+    return dateutil.parser.parse(m.group(1).decode('ascii'))
+
+
+@contextmanager
+def make_csr(domain, key):
+    _, csr_path = tempfile.mkstemp()
```
kpengboy: Defensive programming practices suggest you should close the fd that `tempfile.mkstemp()` returns, or use `NamedTemporaryFile` instead. We never know when someone will copy this function's code into the code of some long-running process, and all of a sudden that long-running process is leaking fd's.

modules/ocf/files/ssl/ocf-lets-encrypt
```diff
@@ -0,0 +1,253 @@
...
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        stdout, stderr = proc.communicate()
+    return stdout, stderr, proc.returncode
+
+
+def write_cert(cert, cert_path):
+    """Write certificate file for a domain."""
+    _, temp_cert_path = tempfile.mkstemp(dir=str(cert_path.parent), prefix='.tmp')
```
kpengboy: Try using a `NamedTemporaryFile` with `delete=False` in the constructor instead of opening the temporary file twice.

modules/ocf/files/ssl/ocf-lets-encrypt
```diff
@@ -0,0 +1,253 @@
...
+@contextmanager
+def maybe_start_webserver(dry_run):
+    if not dry_run and socket.socket().connect_ex(('127.0.0.1', 80)) != 0:
+        proc = subprocess.Popen(
+            ('python3', '-m', 'http.server', '80'),
+            cwd=str(ACME_BASE_DIR),
+        )
+
+        # Wait for server to start up before proceeding
+        while socket.socket().connect_ex(('127.0.0.1', 80)) != 0:
```
kpengboy: Let's hope the server always manages to start up, or else this script will hang forever...

modules/ocf_www/files/build-vhosts
```diff
@@ -353,8 +349,12 @@ def main():
         report(config)
         return 0
 
-    changed = test_and_overwrite_config(site_cfg, config, args.target)
+    if args.target == 'web':
+        config_test_cmd = ('apachect', 'configtest')
```
kpengboy: s/apachect/apachectl/

modules/ocf_www/files/lets-encrypt-update
```diff
@@ -110,136 +81,6 @@ def eligible_domains(target):
...
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-        )
-        stdout, stderr = proc.communicate()
-    return stdout, stderr, proc.returncode
-
-
-def write_cert(cert, cert_path):
-    """Write certificate file for a domain."""
-    _, temp_cert_path = tempfile.mkstemp(dir=str(cert_path.parent), prefix='.tmp')
```
kpengboy: Oh ew....

modules/ocf_www/files/lets-encrypt-update
```diff
@@ -254,24 +95,27 @@ def main(argv=None):
...
-            maybe_acquire(domain, cert_path, dry_run=args.dry_run)
+    if args.target == 'web':
+        vhosts, target_domain = get_vhosts(), 'death.ocf.berkeley.edu'
+    else:
+        vhosts, target_domain = get_app_vhosts(), 'werewolves.ocf.berkeley.edu'
+
+    domains = all_domains(vhosts)
+    eligible = eligible_domains(domains, target_domain)
+
+    ret = subprocess.call(chain(
```
kpengboy: Somebody's been coding in Haskell a lot, I can tell :)

Why don't we try making use of PEP 448...

modules/ocf/files/ssl/ocf-lets-encrypt
matthew-mcallister: See https://docs.google.com/document/d/1teNCynslk71N0UU-O_GiSbCMBLZkAvL8SEot4RN1U5E

modules/ocf_www/files/lets-encrypt-update
matthew-mcallister: Yeah, most of your comments were on code that was already there.

modules/ocf_www/files/build-vhosts
jvperrin: Ah, that explains the error email I saw earlier

matthew-mcallister: That's I get for hasty refactoring...

modules/ocf/files/ssl/ocf-lets-encrypt
matthew-mcallister: Oh, thanks, it should error if the process dies.

modules/ocf/files/ssl/ocf-lets-encrypt
matthew-mcallister: Sure, why not. Although I'm pretty sure it uses no resources except those the OS would reclaiim anyways (memory and the socket).

modules/ocf_www/files/lets-encrypt-update
matthew-mcallister: Good suggestion, but this has to run on Python 3.4 because death is still on jessie. Also, this doesn't seem particularly haskell-esque to me.

kpengboy:
Oh right I forgot about that gg

modules/ocf/files/ssl/ocf-lets-encrypt
matthew-mcallister: It would be cool to make `lets-encrypt-update` trigger `build-vhosts` now that I think of it.

modules/ocf/files/ssl/ocf-lets-encrypt
matthew-mcallister: Generally it's not going to be on the same filesystem (tmpfs to NFS), but we could do two moves if that sounds valuable.

matthew-mcallister:
I've addressed comments if anyone wants to review all this again.

modules/ocf/files/ssl/ocf-lets-encrypt
```diff
@@ -0,0 +1,254 @@
...
+            assert proc.poll() is None, proc.returncode
+
+        yield proc
+
+        proc.terminate()
```
kpengboy: Well especially if we change it to `proc.terminate()`, then we may want to wait for the process to actually finish...

modules/ocf/files/ssl/ocf-lets-encrypt
```diff
@@ -0,0 +1,254 @@
...
+        delete=False,
+    ) as temp_cert:
+        try:
+            temp_cert.write(cert)
+            # atomic overwrite in case of errors
+            os.chmod(temp_cert.name, 0o644)
+            os.rename(temp_cert.name, str(cert_path))
+        finally:
+            try:
+                os.remove(temp_cert.name)
```
kpengboy: Eh, fwiw if it were me, I would only remove the file if the rename did not succeed. This would avert the small chance that some other process created a file with the same name as `temp_cert.name` between a successful rename and a remove. That having been said, it's a mostly theoretical race...

modules/ocf/files/ssl/ocf-lets-encrypt
matthew-mcallister: You mean like call `terminate()`, call `communicate()` with a timeout, and then call `kill()` if it still hasn't finished? If so, then, sure, it doesn't hurt, but I don't see the benefit.

modules/ocf/files/ssl/ocf-lets-encrypt
matthew-mcallister: Meh, somehow I think it's more likely for `os.rename` to mysteriously silently fail and not trigger the remove than for this race to happen.

matthew-mcallister:
Alright, seems to me like this is good to merge if no one comes up with any further comments.

PR #206, opened 2017-10-11
add buy spreadsheet shortlink

abizer:
going to RT to find this is super annoying + the link is shared only with officers@ so it should be safe


kpengboy:
I don't see too much of a problem for opening up spreadsheet access to regular staff fwiw. Every allocation needs to be voted on by BoD anyway...

matthew-mcallister:
What about adding this to `RESERVED_USERNAMES`?

I would just add a redirect to my `public_html` dir...

abizer:
I think baisang owns the spreadsheet and would have to change the sharing rules as I don't have privs on it.

and @matthew-mcallister this is hopefully useful for more people than just me which is why I proposed adding it here...

matthew-mcallister:
But anyone can visit a URL in your `public_html` dir, that's why it's called "public". Or `~staff/public_html` for that matter.

Baisang:
closing this due to new sheet layout.

new forms will be open to all

jvperrin:
(closed by #207)

PR #207, opened 2017-10-16
short urls for purchases

Baisang:


Baisang:
good catch @jvperrin , i closed it.

PR #211, opened 2017-10-21
Set NFS grace period using the environment file.

kpengboy:
It was a little tricky to find how to do this, but I ran `systemctl list-dependencies nfs-server`, found an `nfs-config` service which ran a script that read the files in `/etc/default` and built the environment file for `nfs-server.service`, and reading that script, found the correct option to add to `/etc/default/nfs-kernel-server`.


kpengboy:
Comments? If not I will merge

PR #212, opened 2017-10-23
Set up puppetdb

jvperrin:
I'd like to add a dashboard to use to display information from PuppetDB later, but I'll probably put it on Marathon. I tried out [puppetboard](https://github.com/voxpupuli/puppetboard) today and it was pretty nice, so I think going with that would be good.

### DIFF
--- a/modules/ocf/manifests/auth.pp
+++ b/modules/ocf/manifests/auth.pp
@@ -132,6 +132,14 @@ class ocf::auth($glogin = [], $ulogin = [[]], $gsudo = [], $usudo = [], $nopassw
     require => Augeas['sshd_config'],
   }
 
+  # Export SSH keys from every host, and use them to populate the global list
+  # in /etc/ssh/ssh_known_hosts
+  @@sshkey { $::hostname:
+    type => ecdsa-sha2-nistp256,
+    key  => $::sshecdsakey,
+  }
+  Sshkey <<| |>>
+
   # sudo user/group access controls
   package { 'sudo': }
   file {
--- a/modules/ocf_backups/files/rsnapshot.conf
+++ b/modules/ocf_backups/files/rsnapshot.conf
@@ -61,6 +61,8 @@ backup	ocfbackups@ldap:/var/backups/ldap/	servers/ldap/
 backup	ocfbackups@puppet:/etc/puppetlabs/	servers/puppet/
 backup	ocfbackups@puppet:/opt/puppetlabs/	servers/puppet/
 
+backup	ocfbackups@puppetdb:/etc/puppetlabs/puppetdb/ssl/	servers/puppetdb/
+
 backup	ocfbackups@stats:/opt/stats/var/	servers/stats/
 
 backup	ocfbackups@munin:/var/lib/munin/	servers/munin/
--- a/modules/ocf_backups/files/rsnapshot.conf
+++ b/modules/ocf_backups/files/rsnapshot.conf
@@ -61,7 +61,7 @@ backup	ocfbackups@ldap:/var/backups/ldap/	servers/ldap/
 backup	ocfbackups@puppet:/etc/puppetlabs/	servers/puppet/
 backup	ocfbackups@puppet:/opt/puppetlabs/	servers/puppet/
 
-backup	ocfbackups@puppetdb:/etc/puppetlabs/puppetdb/ssl/	servers/puppetdb/
+backup	ocfbackups@puppetdb:/etc/puppetlabs/puppet/ssl/	servers/puppetdb/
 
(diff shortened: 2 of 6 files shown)


modules/ocf/manifests/auth.pp
```diff
@@ -132,6 +132,14 @@
     require => Augeas['sshd_config'],
   }
 
+  # Export SSH keys from every host, and use them to populate the global list
+  # in /etc/ssh/ssh_known_hosts
+  @@sshkey { $::hostname:
+    type => ecdsa-sha2-nistp256,
+    key  => $::sshecdsakey,
+  }
+  Sshkey <<| |>>
```
jvperrin: I'm hoping this will remove the errors on regenerating a machine, but I don't think it will since each user stores this information in their local `~/.ssh` directory. It will make it so that `yes` doesn't have to be typed for the first connection though, so that's nice.

modules/ocf_backups/files/rsnapshot.conf
```diff
@@ -61,6 +61,8 @@ backup	ocfbackups@ldap:/var/backups/ldap/	servers/ldap/
 backup	ocfbackups@puppet:/etc/puppetlabs/	servers/puppet/
 backup	ocfbackups@puppet:/opt/puppetlabs/	servers/puppet/
 
+backup	ocfbackups@puppetdb:/etc/puppetlabs/puppetdb/ssl/	servers/puppetdb/
```
jvperrin: This is only backed up because this cert was manually generated to include the DNS SANs for `puppetdb`, `thunder`, and `puppetdb.ocf.berkeley.edu` alongside the default `thunder.ocf.berkeley.edu` (FQDN).

modules/ocf_puppet/manifests/puppetmaster.pp
```diff
@@ -3,9 +3,12 @@
     ['puppetserver', 'puppet-lint', 'augeas-tools']:;
   }
 
-  service { 'puppetserver':
-    enable  => true,
-    require => Package['puppetserver'],
+  # This defines Service['puppetserver'], so we can't do it ourselves.
+  class { 'puppetdb::master::config':
+    puppetdb_server => 'thunder.ocf.berkeley.edu',
```
jvperrin: This could probably be removed now that the SAN cert is generated, as the default here is `puppetdb`, so I'll go change that.

modules/ocf_puppetdb/manifests/init.pp
```diff
@@ -1,3 +1,31 @@
 class ocf_puppetdb {
-  package { ['puppetdb', 'puppetdb-termini']:; }
+  include ocf_ssl::default_bundle
```
jvperrin: I originally added this for [puppetboard](https://github.com/voxpupuli/puppetboard), but I think putting it on Marathon is cleaner, so I'll do that and remove this line and the certs.

jvperrin: Never mind, I'll probably just proxy this to puppetboard, like with the web IRC.

modules/ocf_puppet/manifests/puppetmaster.pp
jvperrin: I was wrong about the default, it defaults to the local host, but I've changed this to a more general `'puppetdb'` and that's better in case the hostname needs to change.

modules/ocf_puppetdb/manifests/init.pp
matthew-mcallister: So proxy it from the puppetdb server so you can visit it at puppetdb.ocf.berkeley.edu? Seems convenient.

I seem to recall you said that puppetboard required authentication, presumably because it lets you change things/view secrets. If so, would the idea be to put nginx here to do SSL and proxy to the LB, and then each puppetboard instance would have another nginx which does authentication?

jvperrin: I'm not exactly sure, but I do think that having it be available at `puppetdb.ocf.berkeley.edu` would be nice. Might even be nice to be available at `puppet.ocf.berkeley.edu`, but that might be weird if we implement something else that wants to run a web server on lightning, and then we wouldn't want it to be at both `puppet.` and `puppetdb.`, because that would be redundant and a bit confusing. Currently I'm leaning towards just putting it at `puppetdb.` because that's more indicative of what it does, but I'm open to suggestions on which of those would be better.

As for authentication, I think having the inner nginx do the authentication would be the most secure and make the most sense, like RT does it, except simpler because it doesn't have to pass through any credentials or anything and can use nginx. Just having this instance do SSL termination and proxying like the [web IRC proxy](https://github.com/ocf/puppet/blob/master/modules/ocf_irc/manifests/webirc.pp) does seems easy enough.

PR #214, opened 2017-10-24
Change ocf-nfs fact script to work on stretch.

### LINKED ISSUE #213: ocf_nfs fact is broken on stretch
The [check made for the `ocf_nfs` fact](https://github.com/ocf/puppet/blob/6d53dc9f18b926a0597b98d11b47f207c8567f3b/modules/ocf/facts.d/ocf-nfs#L2) doesn't work on stretch, since it only has one slash instead of two:

supernova (stretch):

    $ mount | grep home
    services:/home on /home type nfs4 [...]

tsunami (jessie):

    $ mount | grep home
    services://home on /home type nfs4 [...]

redplanks:
Resolves #213

### DIFF
--- a/modules/ocf/facts.d/ocf-nfs
+++ b/modules/ocf/facts.d/ocf-nfs
@@ -1,5 +1,5 @@
 #!/bin/bash -eu
-if mount | grep -qE '^[a-z]+://home\b'; then
+if mount | grep -qE '^[a-z]+://?home\b'; then
     echo 'ocf_nfs=true'
 else
     echo 'ocf_nfs=false'


chriskuehl:
Do we still need to support jessie?

jvperrin:
This works with jessie too, since jessie just has an extra slash, so it matches too

chriskuehl:
I don't believe you :P

jvperrin:
I thought it could just be changed to a single slash and it would match both, but I thought the format here was different. It would need to be more complicated, maybe something like `^[a-z]+:/{1,2}home\b`?

chriskuehl:
Just add a question mark after one of the slashes?

jvperrin:
Sure, that works too

PR #215, opened 2017-10-25
Print quota on panel

kkuehler:
Users can now more easily see their print quota stats on the desktop, as opposed to catching the notification as it displays, or god-forbid opening a terminal and typing 'paper'. 

I know not all staff will want to test this, so I attached two images of the modified panel.

https://i.fluffy.cc/97VK7V52jLkp9TC9HW3fZG5dCk856rZk.png
https://i.fluffy.cc/ZFz2Q2WlkG8Pt0bsdPthRbhTmrFtDj14.png


modules/ocf_desktop/files/xsession/notify
```diff
@@ -32,15 +31,10 @@ def main():
...
-        if quota != old_quota:
-            title, msg = quota.decode(encoding='UTF-8').split('\n', 1)
-            msg += '(as of {})'.format(datetime.now().strftime('%-I:%M%P'))
-            Notify.Notification.new(title, msg, NOTIFICATION_ICON).show()
-            old_quota = quota
-        time.sleep(60)
+    quota = subprocess.check_output(PAPER_PATH)
+    title, msg = quota.decode(encoding='UTF-8').split('\n', 1)
+    msg += '(as of {})'.format(datetime.now().strftime('%-I:%M%P'))
+    Notify.Notification.new(title, msg, NOTIFICATION_ICON).show()
```
chriskuehl: i still think this should use the ocflib functions rather than parsing the paper output

modules/ocf_desktop/files/xsession/paper-genmon
```diff
@@ -0,0 +1,39 @@
...
+def main(argv=None):
+    grammar_sux = {1: 'page'}
+    user = None
+    daily_quota = 'N/A'
+    semester_quota = 'N/A'
+
+    try:
+        user = current_user()
+    except os.error:
+        return 1
```
chriskuehl: when does this happen?

modules/ocf_desktop/manifests/xsession.pp
```diff
@@ -43,7 +47,8 @@
     '/opt/share/xsession/icons':
       source  => 'puppet:///modules/ocf_desktop/xsession/icons/',
       recurse => true,
-      purge   => true;
+      purge   => true,
+      before  => File['/usr/local/bin/paper-genmon'];
```
chriskuehl: imo it's usually better for things to specify their dependencies, rather than other things specifiying them as reverse dependencies

i.e. i'd write `require => File['/opt/share/xsession/icons']` on the script instead of this

modules/ocf_desktop/files/xsession/paper-genmon
```diff
@@ -0,0 +1,39 @@
...
+
+    if user_exists(user) and not user_is_group(user):
+        with get_connection() as c:
+            quota = get_quota(c, user)
+            daily_quota = quota.daily
+            semester_quota = quota.semesterly
+
+    print('<img>{}</img>'.format(IMAGE_PATH))
+    print('<txt> {} {} today</txt>'.format(daily_quota,
```
abizer: "pages left today" would be a little more clear

modules/ocf_desktop/files/xsession/paper-genmon
```diff
@@ -0,0 +1,39 @@
...
+                                           grammar_sux.get(daily_quota, 'pages')))
```
abizer: also why not just do something like `"page" if daily_quota == 1 else "pages"`

modules/ocf_desktop/files/xsession/notify
kkuehler: Understandable, but this is how notify was; I just took out the infinite loop. It makes more sense to do this in a separate PR.

modules/ocf_desktop/files/xsession/paper-genmon
kkuehler: Yeah, that works and is probably more clear.

modules/ocf_desktop/files/xsession/notify
chriskuehl: agreed

modules/ocf_desktop/files/xsession/paper-genmon
kkuehler: Heh, whoops. Yeah, ```getuid``` never fails. 👍

modules/ocf_desktop/files/xsession/paper-genmon
kkuehler: Yeah, but take a look at the linked pictures and let me know if you want that much text. Too much text on the bar is ugly, but I agree it is also important that users don't misinterpret this.

kkuehler:
@chriskuehl, @abizer Changes made (plus I switched to an in-house icon, courtesy to Raymond).

matthew-mcallister:
That icon is pretty faint. Will you even be able to see it against the background?

kkuehler:
@matthew-mcallister https://i.fluffy.cc/mv0jSSs1D7f40vQQtnspbjKfHsKgD3br.png

kpengboy:
Remind me, what's the motivation for keeping the notify script now? I don't see it being used except to display one notification on login...?

kkuehler:
@kpengboy it's nice to see staff on login, I think.

modules/ocf_desktop/files/xsession/paper-genmon
```diff
@@ -0,0 +1,33 @@
...
+    if user_exists(user) and not user_is_group(user):
+        with get_connection() as c:
+            quota = get_quota(c, user)
+            daily_quota = quota.daily
+            semester_quota = quota.semesterly
+
+    print('<img>{}</img>'.format(IMAGE_PATH))
+    print('<txt> {} {} left</txt>'.format(daily_quota,
+                                          'page' if daily_quota == 1 else 'pages'))
+    print('<tool>{} remaining this semester</tool>'.format(semester_quota))
```
kpengboy: Would be nice to have a message about group accounts not being able to print in the tooltip, for group accounts.

kpengboy:
@kkuehler eh...in that case it may not be necessary for notify to show any tooltip at all regarding pages. Do other people have other opinions?

kpengboy:
Thanks @kkuehler, this is a great addition!

PR #217, opened 2017-10-26
Don't run paper inside notify

kpengboy:
Since there's now a genmon which does this for us.

What do you all think? Should it stay or go?


chriskuehl:
imo stay, the paper notification is friendlier (ocf logo) and more obvious. i bet most people don't really look at the top bar, it's pretty crowded now.

kkuehler:
@kpengboy Forget what I just said. It needs to be fixed, but it's not exactly urgent and I want to get other staff members involved that are not as familiar with ocflib and puppet on this.

PR #218, opened 2017-10-28
Add global hiera veriable to track if a server is development.

### LINKED ISSUE #194: Add a fact to identify dev- hosts
As suggested by @chriskuehl in #193, it would be much nicer to have something to identify dev hosts rather than checking for `dev-` at the start of the hostname [every time](https://github.com/ocf/puppet/search?utf8=%E2%9C%93&q=%24%3A%3Ahostname+%21~+%2F%5Edev-%2F).

redplanks:
Should resolve #194

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -2,9 +2,9 @@
 moduledir 'vendor'
 
 
-# Third-party modules
-# Make sure to install any dependencies of packages here, dependency resolution
-# is not automatically done by r10k unfortunately
+# Third-party modules, make sure to check their dependencies and add them here
+# too, since r10k does not manage dependencies and you will be confused as to
+# why they are not working properly.
 
 # TODO: Upgrade puppet-nginx to 0.7.1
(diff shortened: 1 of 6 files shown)


matthew-mcallister:
@kpengboy Do you mean setting the value in hiera using a regex?

matthew-mcallister:
@redplanks Also, some classes (e.g. `dev_apphost`, `dev_printhost`) already have a `dev_config` variable. You shouldn't add another redundant variable but use that one instead. It would probably also be best to use that name everywhere rather than inconsistently using `dev_server` and `dev_config`.

jvperrin:
I do think having something to check the hostname as well would be good, but not sure how that would work with hiera, I think it would need to be done with a fact or something instead?

redplanks:
Right now, dev_config is manually injected into class blocks, so extending that solution further seems clunky, as someone would have to target the individual class directly and add a default false value to the class declaration. Would it be good to use dev_config, but change it to a global hiera variable?

PR #219, opened 2017-10-29
ocf_admin: Replace stunnel with hitch

matthew-mcallister:
stunnel has horrible problems with leaking memory and spamming up the
logs; hitch is nice and simple and spawns way fewer worker processes

### DIFF
--- a/modules/ocf_admin/manifests/create/redis.pp
+++ b/modules/ocf_admin/manifests/create/redis.pp
@@ -1,5 +1,5 @@
 class ocf_admin::create::redis {
-  package { ['redis-server']: }
+  package { ['redis-server', 'hitch']: }
 
   service { 'redis-server':
     require => Package['redis-server'],
@@ -21,10 +21,8 @@ class ocf_admin::create::redis {
     lens    => 'Spacevars.simple_lns',
     incl    => '/etc/redis/redis.conf',
     changes =>  [
-      'set port 0',
+      'set port 6379',
       "set requirepass ${redis_password}",
-      'set unixsocket /var/run/redis/redis.sock',
-      'set unixsocketperm 666',
 
       # Redis offers two data stores: RDB and AOF.
       # Supposedly AOF is more durable than RDB, so that's what we'll use.
@@ -40,14 +38,19 @@ class ocf_admin::create::redis {
     notify  => Service['redis-server'];
   }
 
-  ocf::repackage { 'stunnel4': backport_on => jessie } ->
-  augeas { '/etc/default/stunnel4':
-    lens    => 'Shellvars.lns',
-    incl    => '/etc/default/stunnel4',
-    changes =>  ['set ENABLED 1'],
-  } ~>
-  file { '/etc/stunnel/redis.conf':
-    content => template('ocf_admin/stunnel/redis.conf.erb'),
-  } ~>
-  service { 'stunnel4': }
+  # We already have an OCF member with the username "hitch", so dpkg
+  # chooses "_hitch" as a fallback username.
+  user { '_hitch':
+    groups  => 'ssl-cert',
+    notify  => Service['hitch'],
+    require => Package['hitch'],
+  }
+
+  file { '/etc/hitch/hitch.conf':
+    content => template('ocf_admin/hitch.conf.erb'),
+    notify  => Service['hitch'],
+    require => Package['hitch'],
+  }
+
+  service { 'hitch': }
 }
--- a/modules/ocf_admin/templates/hitch.conf.erb
+++ b/modules/ocf_admin/templates/hitch.conf.erb
@@ -1,0 +1,10 @@
+frontend = {
+	host = "*"
+	port = "6378"
+}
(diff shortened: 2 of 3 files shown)


modules/ocf_admin/manifests/create/redis.pp
```diff
@@ -40,14 +38,21 @@
...
-    incl    => '/etc/default/stunnel4',
-    changes =>  ['set ENABLED 1'],
-  } ~>
-  file { '/etc/stunnel/redis.conf':
-    content => template('ocf_admin/stunnel/redis.conf.erb'),
-  } ~>
-  service { 'stunnel4': }
+  package { 'hitch': }
+
+  # Yes, the underscore is intentional. We already have a user named "hitch".
```
jvperrin: Can't you add the `ssl-cert` group to the existing user instead of creating a new user?

matthew-mcallister: `dpkg` picked the username, not me. Look:
```
mattmcal@supernova ~ $ check hitch
hitch:*:25740:1000:******:/home/h/hi/hitch:/bin/bash
Created on: 2012-10-28
CalNet UID number: ******
```
I'll edit the comment to clarify.

modules/ocf_admin/manifests/create/redis.pp
```diff
@@ -40,14 +38,21 @@
...
+  user { '_hitch':
+    groups  => 'ssl-cert',
+    notify  => Service['hitch'],
+    require => Package['hitch'],
+  }
+
+  $pem_file = "/etc/ssl/private/${::fqdn}.pem"
```
jvperrin: Why not just put this in the template itself? Seems like that would be pretty reasonable and then you wouldn't need this line:

    pem-file = "/etc/ssl/private/<%= @fqdn -%>.pem"

modules/ocf_admin/manifests/create/redis.pp
jvperrin: Ah, that makes sense, I thought you meant the package already created a user named hitch, not that the OCF already had one because of a user, that makes more sense.

modules/ocf_admin/templates/hitch.conf.erb
```diff
@@ -0,0 +1,9 @@
+frontend = {
+	host = "*"
+	port = "6378"
+}
+backend = "[127.0.0.1]:6379"
+workers = <%= @processors['count'] %>
+keepalive = 32767
```
jvperrin: Maybe add a comment about this being the maximum keepalive so nobody tries to increase it?

modules/ocf_admin/manifests/create/redis.pp
```diff
@@ -40,14 +38,21 @@
...
-  augeas { '/etc/default/stunnel4':
-    lens    => 'Shellvars.lns',
-    incl    => '/etc/default/stunnel4',
-    changes =>  ['set ENABLED 1'],
-  } ~>
-  file { '/etc/stunnel/redis.conf':
-    content => template('ocf_admin/stunnel/redis.conf.erb'),
-  } ~>
-  service { 'stunnel4': }
+  package { 'hitch': }
```
jvperrin: I think just putting this in the package array at the top of the file would be easier and a bit cleaner, but keeping it here does make some sense since it's grouped with the other hitch stuff then. I'd say it's up to you either way, but personally I'd put it at the top of the file with the `redis-server` package.

modules/ocf_admin/manifests/create/redis.pp
matthew-mcallister: That's a reasonable question. At first it was a parameter to the class, but I changed my mind about generalizing the class if we're not actually planning to use redis anywhere else at the moment. I'll take your advice since it seems silly as-is.

modules/ocf_admin/templates/hitch.conf.erb
matthew-mcallister: Oh yeah, I forgot about that. Excellent idea.

modules/ocf_admin/manifests/create/redis.pp
matthew-mcallister: Also a good suggestion; that's a leftover from the `->` chaining in the original code.

modules/ocf_admin/manifests/create/redis.pp
```diff
@@ -40,14 +38,21 @@
...
+  }
+
+  $pem_file = "/etc/ssl/private/${::fqdn}.pem"
+  file { '/etc/hitch/hitch.conf':
+    content => template('ocf_admin/hitch.conf.erb'),
+    notify  => Service['hitch'],
+    require => Package['hitch'],
+  }
+
+  service { 'hitch': }
```
kpengboy: Would it be of any concern that this doesn't depend on `Package['hitch']`?

matthew-mcallister: At the moment it already does because the preceding resources `notify` it. Unless someone deletes those resources, it should be fine.

PR #221, opened 2017-10-31
Added ocf survey shortlink

vaibhavjayaraman:


matthew-mcallister:
[You will need to reserve the username in ocflib](https://github.com/ocf/ocflib/blob/master/ocflib/account/validators.py#L10).

dkess:
You also need to add this to the list of reserved usernames, but this commit is good.

PR #223, opened 2017-10-31
Make corruption a hypervisor.

redplanks:


matthew-mcallister:
You'll also need `ocf::networking::bridge: true` for the VMs to have networking.\

Even after we merge this, though, you still can't run any VMs because we've already partitioned the whole disk on corruption, so you won't be able to create LVM volumes. It will have to wait until after we add more drives and set them up with RAID.

redplanks:
Can the current partition be resized?

matthew-mcallister:
Maybe, if you PXE booted it. Or you could just reinstall. There's 256G total on the disk, so there would be room to create a couple VMs for testing. Or you could mount NFS, but it's slow.

Staffers may be using corruption at the moment, so we'd have to tell everyone to gtfo first.

abizer:
We'd have to reprovision corruption when we get new disks anyways, I think. Honestly I don't think it's worthwhile to do this right now anyways, what are you planning on using the VMs for?

redplanks:
The consensus on IRC was to use VMs right now in order to divide up HPC for multiple users at a time, and later use the VMs as nodes for the SLURM job scheduler.

abizer:
We don't even have multiple users for HPC at this time, and the requests we've received probably won't come through until next semester. Furthermore the overhead of running VMs to multiplex HPC jobs as opposed to using a straight scheduler seems to me like it would result in worse performance, especially since the kinds of jobs we are interested in running are probably optimized for multiple cores anyways. IMHO setting up KVM right now would be most interesting for figuring out to what degree we can support PCI passthrough, SR-IOV, and IOMMU, and then, once we get more RAM, disks, a second processor for the empty socket in corruption, and more GPUs, we can use VMs to multiplex jobs and pin the VMs to physical cores and exclusively assign GPUs to them to minimize the VM overhead as much as possible.

matthew-mcallister:
@abizer By your logic, it sounds like we _should_ be doing this now, before we have any users, so we have time to develop and experiment. It's not like anyone is using corruption for any serious work right now.

abizer:
Sure, my wording was pretty terrible. At present, setting up VMs to multiplex jobs seems like it would incur more overhead than it gives us benefits, but once we have more resources in corruption, that balance may switch around.

One of the other benefits of the XCF hypervisors is they can be used as SLURM nodes as well, so you may want to experiment with that as well.

redplanks:
Closing; SLURM has support for multiple jobs on a single node, so making corruption a hypervisor would just be unnecessary overhead.

PR #224, opened 2017-10-31
Add ocfweb to dkessler staffvm

dkess:


abizer:
You may also need to install libcrack2-dev if it isn't already as some of the python crypto libraries won't compile without it

PR #225, opened 2017-10-31
add libcrack2-dev to ocfweb::dev_config

abizer:
we ran into this problem recently with @tanx16's staffvm, where running `make venv` would fail when compiling the python crypto libraries because of a missing header file. Making it an explicit dependency for the dev config should fix that issue.

### DIFF
--- a/modules/ocf_ocfweb/manifests/dev_config.pp
+++ b/modules/ocf_ocfweb/manifests/dev_config.pp
@@ -11,6 +11,11 @@ class ocf_ocfweb::dev_config($group = 'ocfstaff') {
   $broker = "redis://:${redis_password}@admin.ocf.berkeley.edu:6378"
   $backend = $broker
 
+
+  # installed in extrapackages but this is an dependency
+  # of the crypto libraries used in ocfweb as well
+  package { 'libcrack2-dev':; }
+
   file {
     '/etc/ocfweb':
       ensure    => directory;
--- a/modules/ocf_ocfweb/manifests/dev_config.pp
+++ b/modules/ocf_ocfweb/manifests/dev_config.pp
@@ -12,9 +12,11 @@ class ocf_ocfweb::dev_config($group = 'ocfstaff') {
   $backend = $broker
 
 
-  # installed in extrapackages but this is an dependency
-  # of the crypto libraries used in ocfweb as well
-  package { 'libcrack2-dev':; }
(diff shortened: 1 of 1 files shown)


modules/ocf_ocfweb/manifests/dev_config.pp
```diff
@@ -11,6 +11,11 @@
   $broker = "redis://:${redis_password}@admin.ocf.berkeley.edu:6378"
   $backend = $broker
 
+
+  # installed in extrapackages but this is an dependency
+  # of the crypto libraries used in ocfweb as well
+  package { 'libcrack2-dev':; }
```
chriskuehl: this is already installed by ocf::extrapackages, this will conflict if you have both of these manifests in use on the same host (which we do in some cases)

abizer: :(

I guess I'll add a note to the docs on ocfweb instead and delete the resource declaration while keeping the comment so anyone who happens to read this in the future knows that's something they need to do.

chriskuehl: you can just use ensure_packages instead: https://github.com/puppetlabs/puppetlabs-stdlib#ensure_packages

abizer: oh interesting

modules/ocf_ocfweb/manifests/dev_config.pp
```diff
@@ -12,9 +12,11 @@
...
 
 
-  # installed in extrapackages but this is an dependency
-  # of the crypto libraries used in ocfweb as well
-  package { 'libcrack2-dev':; }
+  # libcrack2-dev is a dependency of the crypto libraries that
+  # ocfweb depends on, but we can't just declare the package
+  # because it conflicts with extrapackages, which is declared in
+  # ocf_admin along with this manifest (for supernova)
+  ensure_packages(['libcrack2-dev'], { 'ensure' => 'present'})
```
abizer: @chriskuehl ?

chriskuehl: you just need the first argument

chriskuehl:
eventually we may want to split out a manifest that contains all packages required to build ocflib (and then include it here), but this is fine for now

PR #226, opened 2017-11-03
Add dev_config as a global fact and set it using the hostname in facts.d

### LINKED ISSUE #194: Add a fact to identify dev- hosts

redplanks:
Should resolve #194

### DIFF
--- a/modules/ocf/facts.d/dev-config
+++ b/modules/ocf/facts.d/dev-config
@@ -1,0 +1,6 @@
+#!/bin/bash -eu
+if hostname | grep -qE '^dev-'; then
+    echo 'dev_config=true'
+else
+    echo 'dev_config=false'
+fi
--- a/modules/ocf/facts.d/dev-config
+++ b/modules/ocf/facts.d/dev-config
@@ -1,6 +1,0 @@
-#!/bin/bash -eu
-if hostname | grep -qE '^dev-'; then
-    echo 'dev_config=true'
-else
-    echo 'dev_config=false'
-fi
--- a/modules/ocf/facts.d/host-env
+++ b/modules/ocf/facts.d/host-env
@@ -1,0 +1,6 @@
+#!/bin/bash -euo pipefail
+if hostname | grep -qE '^dev-'; then
+    echo 'host_env=dev'
+else
+    echo 'host_env=prod'
+fi
--- a/modules/ocf_apphost/manifests/lets_encrypt.pp
+++ b/modules/ocf_apphost/manifests/lets_encrypt.pp
@@ -14,7 +14,7 @@ class ocf_apphost::lets_encrypt {
       mode      => '0400';
   }
 
-  if $::hostname !~ /^dev-/ {
+  if !$::dev_config {
     cron { 'lets-encrypt-update':
       command     => 'chronic /usr/local/bin/lets-encrypt-update -v app',
       user        => ocfletsencrypt,
--- a/modules/ocf_apphost/manifests/lets_encrypt.pp
+++ b/modules/ocf_apphost/manifests/lets_encrypt.pp
@@ -14,7 +14,7 @@ class ocf_apphost::lets_encrypt {
       mode      => '0400';
   }
 
-  if !$::dev_config {
+  if $::host_env == 'prod' {
     cron { 'lets-encrypt-update':
       command     => 'chronic /usr/local/bin/lets-encrypt-update -v app',
       user        => ocfletsencrypt,
--- a/modules/ocf_irc/manifests/webirc.pp
+++ b/modules/ocf_irc/manifests/webirc.pp
@@ -1,8 +1,8 @@
 class ocf_irc::webirc {
 
-  $webirc_fqdn = $::dev_config ? {
-    true  => 'dev-irc.ocf.berkeley.edu'
-    false => 'irc.ocf.berkeley.edu'
+  $webirc_fqdn = $::host_env ? {
+    'dev'  => 'dev-irc.ocf.berkeley.edu',
+    'prod' => 'irc.ocf.berkeley.edu',
   }
 
   # Nginx is used to proxy to Marathon and to supply a HTTP -> HTTPS redirect
--- a/modules/ocf_irc/manifests/webirc.pp
+++ b/modules/ocf_irc/manifests/webirc.pp
@@ -1,7 +1,8 @@
 class ocf_irc::webirc {
-  $webirc_fqdn = $::hostname ? {
-    /^dev-/ => 'dev-irc.ocf.berkeley.edu',
-    default => 'irc.ocf.berkeley.edu',
+
+  $webirc_fqdn = $::dev_config ? {
+    true  => 'dev-irc.ocf.berkeley.edu'
+    false => 'irc.ocf.berkeley.edu'
   }
 
   # Nginx is used to proxy to Marathon and to supply a HTTP -> HTTPS redirect
--- a/modules/ocf_www/manifests/site/ocfweb_redirects.pp
+++ b/modules/ocf_www/manifests/site/ocfweb_redirects.pp
@@ -2,9 +2,9 @@
 # redirect those sites to the appropriate pages on ocfweb.
 class ocf_www::site::ocfweb_redirects {
   # accounts
-  $accounts_canonical_url = $::hostname ? {
-    /^dev-/ => 'https://dev-accounts.ocf.berkeley.edu/',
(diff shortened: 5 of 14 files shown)


kpengboy:
LGTM, do other TMs have comment?

modules/ocf/facts.d/dev-config
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -eu
+if hostname | grep -qE '^dev-'; then
```
matthew-mcallister: Not important but I'm pretty sure the `-E` doesn't do anything here, unless you know somewhere we don't use GNU grep.

redplanks:
Ah thanks; I was confused by the syntax of the other regexes--thought they were some kind of arcane variable assignment.

modules/ocf/facts.d/dev-config
redplanks: ocf-nfs also uses grep -qE, so I thought it best to follow convention.

modules/ocf_www/manifests/site/ocfweb_redirects.pp
```diff
@@ -3,7 +3,7 @@
 class ocf_www::site::ocfweb_redirects {
   # accounts
   $accounts_canonical_url = $::hostname ? {
-    /^dev-/ => 'https://dev-accounts.ocf.berkeley.edu/',
+    $::dev_config => 'https://dev-accounts.ocf.berkeley.edu/',
```
matthew-mcallister: [This arcane syntax](https://puppet.com/docs/puppet/4.9/lang_conditional.html#selectors) does not mean what you think it means. All these matches against the host name should be replaced with simple if-else statements.

redplanks:
Changes made.

modules/ocf_irc/manifests/webirc.pp
```diff
@@ -1,7 +1,9 @@
 class ocf_irc::webirc {
-  $webirc_fqdn = $::hostname ? {
-    /^dev-/ => 'dev-irc.ocf.berkeley.edu',
-    default => 'irc.ocf.berkeley.edu',
+  if $::dev_config {
+    $webirc_fqdn = 'dev-irc.ocf.berkeley.edu'
+  }
+  else {
```
matthew-mcallister: >else on a new line

Why though

modules/ocf_www/manifests/site/ocfweb_redirects.pp
```diff
@@ -2,9 +2,11 @@
...
 class ocf_www::site::ocfweb_redirects {
   # accounts
-  $accounts_canonical_url = $::hostname ? {
-    /^dev-/ => 'https://dev-accounts.ocf.berkeley.edu/',
-    default => 'https://accounts.ocf.berkeley.edu/',
+  if $::dev_config {
+    $accounts_canonical_url = 'https://dev-accounts.ocf.berkeley.edu/'
+  }
+  else {
+    $accounts_canonical_url = 'https://accounts.ocf.berkeley.edu/'
```
jvperrin: I think keeping these as selectors makes the most sense, or at least keeps it cleaner since you don't have to duplicate the variable name:

```puppet
$accounts_cannonical_url = $::dev_config ? {
  true    => 'https://dev-accounts.ocf.berkeley.edu/',
  default => 'https://accounts.ocf.berkeley.edu/',
}
```

matthew-mcallister: Huh, I've never thought to do that before, kinda neat. Alternatively you could do
```puppet
$accounts_cannonical_url = if $::dev_config {
  'https://dev-accounts.ocf.berkeley.edu/'
} else {
  'https://accounts.ocf.berkeley.edu/'
}
```
No definite opinion on which looks better.

jvperrin: Yeah, the if statement version is ok, I think the indentation makes it a little less clear, but I don't have a strong opinion on one over the other either.

modules/ocf/facts.d/dev-config
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -eu
```
chriskuehl: needs `-o pipefail`, there's a pipe in this script

modules/ocf_apphost/manifests/lets_encrypt.pp
```diff
@@ -13,7 +13,7 @@
       mode      => '0400';
   }
 
-  if $::hostname !~ /^dev-/ {
+  if !$::dev_config {
```
chriskuehl: did you confirm this is actually getting treated as a boolean? in the past, we always had to use `str2bool` for external facts like this, not sure if things have changed.

modules/ocf/facts.d/dev-config
```diff
@@ -0,0 +1,6 @@
...
+if hostname | grep -qE '^dev-'; then
+    echo 'dev_config=true'
```
chriskuehl: `dev_host` might be a better name. i might even change this to not be a boolean, e.g. `environment=prod` or `environment=dev`

modules/ocf_irc/manifests/webirc.pp
```diff
@@ -1,7 +1,9 @@
...
+    $webirc_fqdn = 'irc.ocf.berkeley.edu'
```
chriskuehl: a slightly more canonical way to write this in puppet is:

```puppet
$webirc_fqdn = $::dev_config ? {
  true  => 'dev-irc.ocf.berkeley.edu',
  false => 'irc.ocf.berkeley.edu',
}
```

this makes it clear that the variable will *always* have a value, and makes it easier to avoid mistakes when refactoring (e.g. accidentally not setting the variable in one of the conditions)

modules/ocf_www/manifests/site/ocfweb_redirects.pp
```diff
@@ -2,9 +2,11 @@
...
   }
```
chriskuehl: same comment

modules/ocf_www/manifests/site/ocfweb_redirects.pp
```diff
@@ -43,9 +45,11 @@
...
   # wiki
-  $wiki_canonical_url = $::hostname ? {
-    /^dev-/ => 'https://dev-wiki.ocf.berkeley.edu/',
-    default => 'https://wiki.ocf.berkeley.edu/',
+  if $::dev_config {
+    $wiki_canonical_url = 'https://dev-wiki.ocf.berkeley.edu/'
+  }
+  else {
+    $wiki_canonical_url = 'https://wiki.ocf.berkeley.edu/'
   }
```
chriskuehl: same comment -- gonna stop commenting on these

modules/ocf_irc/manifests/webirc.pp
jvperrin: I think giving a `default` value here (instead of `false`) is even better to make sure that it always has a value, so that if it happens to not be boolean for some reason that it still gets a value.

chriskuehl: i think it's actually better to avoid default when you expect it to always be one of the two values, because it will error (and make it clear that something is wrong with your inputs) if one of the arms doesn't match:

```puppet
ckuehl@ithuriel:~$ cat test.pp  
$foo = 'blah'
$x = $foo ? {
  true => 1,
  false => 2,
}
```
```bash
ckuehl@ithuriel:~$ puppet apply test.pp
Error: Evaluation Error: No matching entry for selector parameter with value 'blah' at /home/ckuehl/test.pp:2:6 on node ithuriel.home.ckuehl.me

```

redplanks:
Made requested changes.

kpengboy:
Thanks for doing this!

modules/ocf/facts.d/host-env
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -euo pipefail
```
kpengboy: Unfortunately, this doesn't work, because Linux treats everything after the command as one argument. So the actual command called will be:

    /bin/bash "-euo pipefail" /var/cache/puppet/facts.d/host-env

redplanks: Ah, I see. Thank you for fixing it on master!

PR #227, opened 2017-11-03
Small TODO resolutions.

redplanks:


modules/ocf/files/puppet-trigger
```diff
@@ -32,8 +32,7 @@ def switch_to_environment(env):
         environment=env,
     )
 
-    # TODO: use subprocess.run on python3.5; we don't actually care about stdout here
-    subprocess.check_output(
+    subprocess.run(
```
jvperrin: Jessie still has [Python 3.4](https://packages.debian.org/jessie/python3) and not 3.5, so I don't think this can be done until we fully upgrade to stretch.

modules/ocf_www/files/lets-encrypt-update
```diff
@@ -137,8 +137,7 @@ def make_csr(domain):
 
 def acme_tiny(domain):
     with make_csr(domain) as csr_path:
-        # TODO: use subprocess.run when we get python3.5
```
jvperrin: Same here, since `ocf_www` is not on stretch yet :(

modules/ocf_desktop/manifests/packages.pp
```diff
@@ -62,7 +62,8 @@
   # rstudio updates to libgstreamer1.0-0 and libgstreamer-plugins-base1.0-0.
   package { 'rstudio':; }
 
-  # TODO: install remmina from backports when that becomes available
+  # Install remmina from backports.
+  package { 'remmina':; }
```
jvperrin: Nice, that should work, I'd probably include it in the list of packages above instead of keeping it separate (because although it's coming from backports, that's not especially important), but looks good.

redplanks:
Reverted the python changes, and moved remmina over.

jvperrin:
Might want to squash commits, but I don't think it matters too much either way, up to you

redplanks:
I was thinking once everything's on stretch, we can go back and revert the revert.

matthew-mcallister:
I feel like opening an issue or a PR would be a better solution...

jvperrin:
Yeah, I'd agree, I think it'll be a while before we're fully on stretch too, especially `ocf_www`, as we don't really have a solution to change that yet.

kpengboy:
@redplanks can you rebase and squash this?

PR #228, opened 2017-11-03
Firewall Output Rules

kkuehler:
Only OUTPUT rules are in this PR, and the rules are outlined in this design doc: https://docs.google.com/document/d/1xTK9PXBP6c-64_8mytBAEIClWB0hkvGuISYwRBocS2k/edit

### DIFF
--- a/modules/ocf/manifests/firewall/firewall46.pp
+++ b/modules/ocf/manifests/firewall/firewall46.pp
@@ -1,9 +1,9 @@
 define ocf::firewall::firewall46($opts,){
-	firewall{"${title} (IPv4)":
-		* => $opts,
-	}
-	firewall{"${title} (IPv6)":
-	   	       * => $opts,
-		provider => 'ip6tables',
-	}
+  firewall{"${title} (IPv4)":
+    * => $opts,
+  }
+  firewall{"${title} (IPv6)":
+              * => $opts,
+    provider    => 'ip6tables',
+  }
 }
--- a/modules/ocf/manifests/firewall/firewall46.pp
+++ b/modules/ocf/manifests/firewall/firewall46.pp
@@ -1,9 +1,18 @@
-define ocf::firewall::firewall46($opts,){
-  firewall{"${title} (IPv4)":
-    * => $opts,
+define ocf::firewall::firewall46($opts,) {
+  firewall { "${title} (IPv4)":
+    require   => $require,
+    subscribe => $subscribe,
+    before    => $before,
+    notify    => $notify,
+    *         => $opts,
   }
-  firewall{"${title} (IPv6)":
-              * => $opts,
-    provider    => 'ip6tables',
+
+  firewall { "${title} (IPv6)":
+    provider  => 'ip6tables',
+    require   => $require,
+    subscribe => $subscribe,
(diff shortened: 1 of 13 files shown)


modules/ocf_desktop/manifests/firewall_output.pp
```diff
@@ -0,0 +1,24 @@
+#allow desktops to send packets to papercut, pagefault, radiation
```
abizer: why radiation?

jvperrin: That's a good question, because I don't believe they ever have to access radiation directly, only laptops or phones should need access to it.

kpengboy:
This rule concerns access to the access point's configuration interface. I
think it's useful for staff to be able to access that from staff computers.

modules/ocf_desktop/manifests/firewall_output.pp
```diff
@@ -0,0 +1,24 @@
+# allow desktops to send packets to papercut, pagefault, radiation
```
jvperrin: Why to radiation? Do the desktops ever need to communicate with it?

modules/ocf/manifests/firewall/post.pp
```diff
@@ -0,0 +1,71 @@
...
+        proto  => 'tcp',
+        dport  => 'smtp',
+        action => 'drop',
+      },
+      before => undef,
+  }
+
+
+  # Special devices we want to protect from most hosts
+  $devices = ['corruption-mgmt','hal-mgmt', 'jaws-mgmt', 'pagefault', 'pandemic-mgmt',
```
jvperrin: Maybe move these arrays down closer to where they are used so that it is clearer what they are doing?

modules/ocf/manifests/firewall/firewall46.pp
```diff
@@ -0,0 +1,18 @@
+define ocf::firewall::firewall46($opts,) {
```
jvperrin: Why not just add an option here to set `$before` to `undef` by default so that you don't have to keep specifying it everywhere in the same way? Same with the output chain, seems like it would be easier to have a separate class for that so that there's less repetition in `$opts`.

modules/ocf_desktop/manifests/firewall_output.pp
jvperrin: (I know this was asked before, but I don't believe it was answered)

jvperrin: Never mind, it was answered, but not in the same thread. I guess the access point's configuration interface is useful to have access to, although I've never actually used it, but I'm fine with allowing it to be accessible.

modules/ocf/manifests/firewall/firewall46.pp
kpengboy: We wouldn't want to set `$before` to `undef` by default, because we purposely set the default otherwise in `site.pp`:

    Firewall {
      before => Class['ocf::firewall::post'],
    }

As for your other comment, it could be looked into, although I'm not sure it adds much value over writing another `chain => 'PUPPET-OUTPUT'`.

PR #231, opened 2017-11-04
Add dev-rt Marathon service

kpengboy:
There have been several occasions on which I wish I had one to
test with.

This service isn't expected to always be running.


modules/ocf_mesos/manifests/master/load_balancer.pp
```diff
@@ -82,6 +82,14 @@
     ssl_key        => '/opt/share/docker/secrets/ocfweb/static.ocf.berkeley.edu.key',
   }
 
+  ocf_mesos::master::load_balancer::http_vhost { 'dev-rt':
+    server_name  => 'dev-rt.ocf.berkeley.edu',
```
jvperrin: Oh also, add an alias for `dev-rt`, since then this will mean that just typing `dev-rt` when in the lab will actually work, like RT.

kpengboy:
I've decided not to add this for the time being.

PR #233, opened 2017-11-08
OCF listed first in Hexchat network list

kkuehler:
Just a convenience feature. The Hexchat default server list will still be there, only the OCF network will be listed first; set up to connect over SSL.


modules/ocf_desktop/files/skel/config/hexchat/servlist.conf
```diff
@@ -0,0 +1,569 @@
+v=2.12.4
+
+N=OCF
+E=UTF-8 (Unicode)
+F=23
+D=0
+S=irc.ocf.berkeley.edu/6697
```
jvperrin: I think having `J=#rebuild` would be nice here too so they join the `#rebuild` channel by default too.

chriskuehl:
+1 for having a small list. I'd suggest maybe OCF, Freenode, OFTC (that's what Debian uses), Mozilla (used for mozilla projects, rust, etc)?

chriskuehl:
Also, I don't know if the default hexchat config has it, but I bet we can update it so all of the servers we include in the list use SSL.

matthew-mcallister:
You know what? There used to be a web page that documented the `servlist.conf` file format, but I can't find it any more. Anyways, here's what a typical entry should look like:
```
N=freenode
E=UTF-8 (Unicode)
F=22
D=0
S=chat.freenode.net
J=#rebuild
```
You can still see what the flags mean from the [source headers](https://github.com/hexchat/hexchat/blob/15600f405f2d5bda6ccf0dd73957395716e0d4d3/src/common/servlist.h#L58-L81).
```
22 = use global login info + use SSL + use proxy
```

kkuehler:
If we are doing a short list then it'll be easy to use SSL by default for all the provided networks. To add to @chriskuehl's list, EFNet, IRCNet, QuakeNet, Undernet, and Rizon are pretty popular.

chriskuehl:
If you've ever used it, I'd say that's enough to justify adding it. I don't think we need to bikeshed over which to include -- the main thing is just not shipping the _entire_ list Debian ships.

chriskuehl:
(in case my comment wasn't clear, adding those sounds fine to me, and i'm also fine with adding any other servers people feel are useful)

kkuehler:
@jvperrin Okay, changes are made. The connection defaults to SSL on all networks that support it. Unfortunately, QuakeNet and UnderNet don't seem to support SSL, and IRCNet's SSL is experimental and wasn't working for me, so I left it out. [Interestingly, QuakeNet has some strong views on the issue] (https://www.quakenet.org/articles/99-trust-is-not-transitive-or-why-irc-over-ssl-is-pointless)

jvperrin:
Yeah, if they don't support SSL then that's fine, just leave it as non-SSL and that's the best we can do.

Wow, if I didn't know better, I'd think that QuakeNet blog post was trolling. Especially as they say stuff like:
> If your connection to your SSH server gets MITM'ed, you are the only one to lose out as your server/shell gets rooted.

I mean obviously not if it's a shared server, or anyone keeps any other data on it. Their reasoning seems to hint more that IRC clients shouldn't allow invalid SSL certs than anything.

Well, thanks for linking to that, it's pretty interesting.

PR #237, opened 2017-11-11
Add base class to fix excess classes in motds

jvperrin:
This isn't an issue as much any more since staff VMs have a single class now that is shown instead pertaining to docker groups, and they were mostly the hosts that had this problem before, but I think this would make things a bit cleaner organization-wise. This is [rt#5848](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=5848).

### DIFF
--- a/modules/ocf/manifests/base.pp
+++ b/modules/ocf/manifests/base.pp
@@ -1,0 +1,19 @@
+class ocf::base {
+  include ocf::apt
+  include ocf::auth
+  include ocf::autologout
+  include ocf::groups
+  include ocf::kerberos
+  include ocf::ldap
+  include ocf::locale
+  include ocf::motd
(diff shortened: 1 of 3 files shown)


modules/ocf/manifests/base.pp
```diff
@@ -0,0 +1,19 @@
+class ocf::base {
```
chriskuehl: can we just go back to having this be `class ocf` (in the `init.pp` file)? that's the pattern we use elsewhere and what we used to do here. in retrospect i think it was a mistake for us to move this list into hiera, glad we're moving it back

jvperrin: Yeah, that's fair, probably a better naming scheme than this.

PR #238, opened 2017-11-11
Add aliases for SSH known hosts

jvperrin:
This should make it so that connecting to hosts that have been recently re-puppeted does not cause SSH host key errors.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
         -   --no-puppet_url_without_modules-check
         -   --no-arrow_alignment-check
         -   --no-arrow_on_right_operand_line-check
+        -   --no-variable_is_lowercase-check
 -   repo: https://github.com/pre-commit/pre-commit-hooks.git
     sha: v0.9.5
     hooks:
--- a/modules/ocf/manifests/auth.pp
+++ b/modules/ocf/manifests/auth.pp
@@ -132,11 +132,23 @@ class ocf::auth($glogin = [], $ulogin = [[]], $gsudo = [], $usudo = [], $nopassw
     require => Augeas['sshd_config'],
   }
 
+  # Get all DNS names, FQDNs, and IPs for a host to include in SSH keys
+  $ssh_aliases = delete(concat(
+    suffix(delete(any2array($::dnsA), ''), ".${::domain}"),
+    $::dnsA,
+    suffix(delete(any2array($::dnsCname), ''), ".${::domain}"),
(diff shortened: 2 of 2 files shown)


modules/ocf/manifests/auth.pp
```diff
@@ -132,11 +132,23 @@
...
+  # Get all DNS names, FQDNs, and IPs for a host to include in SSH keys
+  $ssh_aliases = delete(concat(
+    suffix(delete(any2array($::dnsA), ''), ".${::domain}"),
+    $::dnsA,
+    suffix(delete(any2array($::dnsCname), ''), ".${::domain}"),
+    $::dnsCname,
+    $::fqdn,
+    $::ipHostNumber,
+    $::ipaddress6,
+  ), '')
```
jvperrin: I think there should be a nicer way to work with the puppet stdlib package to do this, but I couldn't find a way since `$::dnsA` and `$::dnsCname` can be empty, have a single element, or be an array, so there doesn't seem like an easy way to work with them except converting them both to arrays, and then removing any empty strings that still exist.

modules/ocf/manifests/auth.pp
```diff
@@ -132,11 +132,23 @@
...
+
   # Export SSH keys from every host, and use them to populate the global list
   # in /etc/ssh/ssh_known_hosts
```
chriskuehl: so uh... does this mean we can kill kerberos now?

.pre-commit-config.yaml
```diff
@@ -12,6 +12,7 @@ repos:
         -   --no-puppet_url_without_modules-check
         -   --no-arrow_alignment-check
         -   --no-arrow_on_right_operand_line-check
+        -   --no-variable_is_lowercase-check
```
jvperrin: Unfortunately since `$::dnsA`, `$::dnsCname`, and `$::ipHostNumber` come from LDAP, there's not much we can do here except maybe modify the [ldap-enc script](https://github.com/ocf/puppet/blob/master/modules/ocf_puppet/files/ldap-enc) to lowercase them or something.

modules/ocf/manifests/auth.pp
jvperrin: Maybe the ldap-enc script could be modified to handle this better, so that we tell it which values are array-type and which are single element only?

modules/ocf/manifests/auth.pp
jvperrin: Hopefully? :P

PR #241, opened 2017-11-15
enforcer: correctly deduct pages when printing multiple copies on double

tmochida:
Fixing an enforcer bug where printing multiple copies of a document double sided will only deduct 1 copy worth of pages.

**Repro steps:**
1. Make a 2 page document
1. Print 10 copies on single: 20 pages deducted
1. Print 10 copies on double: 2 pages deducted

When you look at the generated PostScript files, page_count() called on the single version returns 20 vs. PS output from double results in 2 (probably because they use different PostScript writers aka pdftops vs ps2write, which write different values on "Pages:" header).
In both cases, the TEACOPIES constant is set to the user-defined number of copies (10).

As a workaround, we can multiply TEACOPIES with the number returned by page_count() only when printing on double.

### DIFF
--- a/modules/ocf_printhost/files/enforcer
+++ b/modules/ocf_printhost/files/enforcer
@@ -112,8 +112,10 @@ def read_config():
     return mysql_user, mysql_passwd
 
 
-def page_count(path):
-    return int(subprocess.check_output(('/usr/local/bin/enforcer-pc', path), timeout=30))
+def page_count(env):
+    path = env['TEADATAFILE']
+    num_copy = 1 if env['CLASS'] == 'single' else int(env['TEACOPIES'])
+    return num_copy * int(subprocess.check_output(('/usr/local/bin/enforcer-pc', path), timeout=30))
 
 
 def create_job(env):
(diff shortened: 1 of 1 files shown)


tmochida:
The alternative is to parse the "Requirements: numcopies(N)" header in the PS file. There's also a RBINumCopies line but this seems to be a comment. Open to suggestions on which approach would be best.

matthew-mcallister:
OK, this makes enough sense to me. Apparently 8530b1fa4a88769a0ad6ec1ae1601dccf6b87e1f only affected single-sided documents; I didn't guess that it would mysteriously be different for single and double.

matthew-mcallister:
Does printing multiple copies double-sided actually work now? In the past we had trouble getting it to because it wouldn't do things like collate copies and insert a blank page at the end if the page count was odd.

tmochida:
@matthew-mcallister I haven't tried collating but multi-copy double sided printing worked for me last week.

abizer:
I forgot to ask earlier but doesn't this mean our jobs tracking of double-sided prints is incorrect? And that our pages-printed numbers are therefore artificially low because of this bug?

tmochida:
Technically yes, but the deviation should be relatively small as I doubt many users print many copies of a double sided document, especially considering the daily quota.

PR #242, opened 2017-11-16
Shorturls to OCF code on Github

ElDynamite:
This will allow links to code on Github in the format ocf.io/gh/p/... to automatically redirect to blob/master if not already pointing there. Links to pulls and issues will not be redirected, but all other "special pages" like /pulse will redirect to project code.

I have not tested this on a staff VM, but I have tested it with .htaccess. Some examples:

ocf.io/nikitnainwal/gh/p/modules/ocf_tv/manifests/pulse.pp : redirects
ocf.io/nikitnainwal/gh/p/blob/master/modules/ocf_tv/manifests/pulse.pp : falls through
ocf.io/nikitnainwal/gh/p/modules/ocf_tv/ : redirects
ocf.io/nikitnainwal/gh/p : falls through
ocf.io/nikitnainwal/gh/l/issues : falls through

### DIFF
--- a/modules/ocf_www/manifests/site/shorturl.pp
+++ b/modules/ocf_www/manifests/site/shorturl.pp
@@ -42,6 +42,7 @@ class ocf_www::site::shorturl {
       {rewrite_rule => '^/email-update$ http://status.ocf.berkeley.edu/2014/06/email-discontinuation-update-forward.html [R]'},
       {rewrite_rule => '^/facebook$ https://goo.gl/forms/dEzJmyRMwAPWCDAY2 [R]'},
       {rewrite_rule => '^/faq$ https://www.ocf.berkeley.edu/docs/faq/ [R]'},
+      {rewrite_rule => '^/gh/([^/]*)(/(?!blob)(?!tree)(?!issue)(?!pull).+)$ https://ocf.io/gh/$1/blob/master$2 [R]'},
       {rewrite_rule => '^/gh/l(/.*)?$ https://github.com/ocf/ocflib$1 [R]'},
       {rewrite_rule => '^/gh/p(/.*)?$ https://github.com/ocf/puppet$1 [R]'},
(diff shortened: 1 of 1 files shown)


abizer:
/gh/p etc. should all still work as that's existing and often-used behavior

kpengboy:
One thing that this does break is `git clone https://ocf.io/gh/p`, although that's probably a minor consideration.

modules/ocf_www/manifests/site/shorturl.pp
```diff
@@ -42,6 +42,7 @@
       {rewrite_rule => '^/email-update$ http://status.ocf.berkeley.edu/2014/06/email-discontinuation-update-forward.html [R]'},
       {rewrite_rule => '^/facebook$ https://goo.gl/forms/dEzJmyRMwAPWCDAY2 [R]'},
       {rewrite_rule => '^/faq$ https://www.ocf.berkeley.edu/docs/faq/ [R]'},
+      {rewrite_rule => '^/gh/([^/]*)(/(?!blob)(?!tree)(?!issue)(?!pull).+)$ https://ocf.io/gh/$1/blob/master$2 [R]'},
```
kpengboy: Also, it would be better to use `(/(?!blob/)(?!tree/)(?!issue/)(?!pull/).+)` or something along those lines, so we can still access a hypothetical file called `blobs/blob1`.

kpengboy:
Anyway, this is pretty useful. Thanks for doing this!

ElDynamite:
Not sure why it breaks with cloning without the info blacklist, but I've added it.

I added the slashes for the categories except issue and pull, since issue and issues are both valid as special pages (same with pull).

kpengboy:
> I added the slashes for the categories except issue and pull, since issue and issues are both valid as special pages (same with pull).

`issues?/` and `pulls?/` would have probably made better regexes fwiw

PR #245, opened 2017-11-17
Update broken link to the Puppet wiki.

redplanks:
In addition, Puppet seems to have retconned references to "Simple Text Patterns" or using sed, so the sed link was updated to include only the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ future staffers what options you actually changed.
 
 Instead of overwriting an entire config file just to change one value, try to
 [use augeas][augeas] ([example][augeas-example]) or [sed][sed]
-([example][sed-example]) to change just the necessary values.
+to change just the necessary values.
 
 ## Future improvements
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -165,7 +165,6 @@ Instead of overwriting an entire config file just to change one value, try to
 [jenkins]: https://jenkins.ocf.berkeley.edu/view/puppet-deploy/
 [r10k]: https://github.com/puppetlabs/r10k
 [puppet-forge]: https://forge.puppet.com/
-[augeas]: http://projects.puppetlabs.com/projects/1/wiki/puppet_augeas
+[augeas]: https://puppet.com/docs/puppet/4.8/types/augeas.html
```
kpengboy: Ideally this wouldn't be pinned on one version of the puppet docs, but pinning on latest wouldn't be a great solution either...

jvperrin: Yeah, I think leaving it like this is fine, latest seems a bit more likely to break anyway, so I think this is slightly better.

PR #246, opened 2017-11-22
Use a smarter autocleaning job

kpengboy:
This job detects when the `/var/lib/docker` filesystem is starting to
get full (as opposed to running at a fixed time indiscriminate of
FS usage) and incrementally cleans Docker stuff until below the
limit.


modules/ocf_jenkins/files/autoclean-docker.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Try progressively cleaning things from oldest to youngest until disk usage is
+# below acceptable threshold. Start at 24 hours and work our way down by 1 hour
+# steps.
+# This cleans containers by creation time rather than exit time, but whatever.
+for ((max_age=24; max_age>=0; max_age--)); do
+    if [ "$(fs_usage)" -le 75 ]; then
+        break
+    fi
+    echo "Cleaning Docker stuff older than ${max_age}h:"
```
kpengboy: Do you people think I should also print this message to syslog (or even redirect the entire output of this script to syslog), or is syslog overrated

modules/ocf_jenkins/files/autoclean-docker
```diff
@@ -0,0 +1,21 @@
...
+
+# Try progressively cleaning things from oldest to youngest until disk usage is
+# below acceptable threshold. Start at 24 hours and work our way down by 1 hour
+# steps.
+# This cleans containers by creation time rather than exit time, but whatever.
+for ((max_age=24; max_age>=0; max_age--)); do
+    if [ "$(fs_usage)" -le 75 ]; then
+        break
+    fi
+    echo "Cleaning Docker stuff older than ${max_age}h:"
```
kpengboy: Do you people think I should also print this message to syslog (or even redirect the entire output of this script to syslog), or is syslog overrated

jvperrin: I think more logging would probably be good, wouldn't the message just be absorbed by chronic and discarded otherwise?

modules/ocf_jenkins/files/autoclean-docker
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+# Deletes old Docker junk to keep /var/lib/docker at a resonable size.
+set -euo pipefail
+
+fs_usage() {
+    df --output=pcent /var/lib/docker | sed -r 's/^ *([0-9]+)%$/\1/; t; d'
```
jvperrin: I think doing something like `df --output=pcent /var/lib/docker | tail -n 1 | sed -r 's/^ *([0-9]+)%$/\1/'` would be clearer than using `t` and `d` from within `sed`, even though it's more commands. I just don't think anyone will understand what those `sed` commands do, whereas `tail` is pretty easy to understand.

kpengboy:
Do you think it's more useful to have an external `syslog-tee` script or just implement a `log` function that echoes the output to stdout as well as syslog?

chriskuehl:
I think an external script would be neat (I'd use it on basically ever cronjob we ever write, similar to chronic), but not essential here. This is definitely a little out-of-scope, just something I was thinking of. (At Yelp we have a cron wrapper which is basically chronic + syslog-tee + a sensu alert combined, it is pretty useful for things like this.)

chriskuehl:
But also like I sorta mentioned, if we ever move to systemd timers/units for scheduled jobs, it's unnecessary since that output goes to the journal by default and can be filtered by unit. Long-term I think that'd be pretty cool.

kpengboy:
Too bad if you want to get output from systemd timer-jobs emailed to you, you have to hack up your own script...

kpengboy:
I'll stick with an internal logging command for now. A `syslog-tee` script is tricky to get right, especially when it gets to outputting the right PID to `logger`.

PR #248, opened 2017-11-27
Create ocf::nginx_proxy for common proxy settings

jvperrin:
I found myself wanting to repeat the same proxy setup many times and thought `ocf_mesos::master::load_balancer::http_vhost` was pretty useful, but it could only really be used for Marathon proxies unless it was made more generic. Anyway, so I made a more general-purpose proxy that makes it easy to create new SSL proxies. It should also make it easier to change any OCF-wide SSL settings as well, for instance.

This is still a bit of a WIP, since some of the `ocf_mesos` stuff could be migrated to use this, and my original intention was to add a puppetboard proxy at https://puppet.ocf.berkeley.edu, but that hasn't been done yet either.

### DIFF
--- a/modules/ocf/manifests/nginx_proxy.pp
+++ b/modules/ocf/manifests/nginx_proxy.pp
@@ -17,18 +17,20 @@ define ocf::nginx_proxy(
   $nginx_options = {},
 ) {
 
+  $base_headers = [
+    'Host $host',
+    'X-Forwarded-For $proxy_add_x_forwarded_for',
+    'X-Forwarded-Protocol $scheme',
+    'X-Real-IP $remote_addr',
+  ]
+
   if $ssl {
     nginx::resource::server {
       $title:
         server_name      => [$server_name],
         proxy            => $proxy,
         proxy_redirect   => $proxy_redirect,
-        proxy_set_header => concat([
-          'Host $host',
-          'X-Forwarded-For $proxy_add_x_forwarded_for',
-          'X-Forwarded-Protocol $scheme',
-          'X-Real-IP $remote_addr',
-        ], $proxy_set_header),
+        proxy_set_header => concat($base_headers, $proxy_set_header),
 
         listen_port => 443,
         ssl         => true,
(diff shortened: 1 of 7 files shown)


modules/ocf_mesos/manifests/master/load_balancer.pp
```diff
@@ -53,59 +53,52 @@
...
-  ocf_mesos::master::load_balancer::http_vhost { 'thelounge':
-    server_name    => 'thelounge.ocf.berkeley.edu',
-    server_aliases => ['thelounge'],
-    service_port   => 10008,
-    ssl            => true,
-    ssl_cert       => '/opt/share/docker/secrets/thelounge/thelounge.ocf.berkeley.edu.crt',
-    ssl_key        => '/opt/share/docker/secrets/thelounge/thelounge.ocf.berkeley.edu.key',
-  }
+  # Port 10008 is used by thelounge, it is proxied to by ocf_irc
+  # Port 10009 is used by puppetboard, it is proxied to by ocf_puppet
```
jvperrin: Well, it doesn't currently, but it should do soon

jvperrin: Ok, now it does :P

modules/ocf/manifests/nginx_proxy.pp
```diff
@@ -0,0 +1,68 @@
...
+        add_header => $headers,
+
+        * => $nginx_options;
+    }
+  } else {
+    nginx::resource::server { $title:
+      server_name => concat([$server_name], $server_aliases),
+      proxy       => $proxy,
+      add_header  => $headers,
+      *           => $nginx_options,
```
matthew-mcallister: Hm, no `proxy_set_header` here?

jvperrin: Ah, yeah, that would be a good idea to have, thanks

modules/ocf/manifests/nginx_proxy.pp
```diff
@@ -0,0 +1,75 @@
...
+    nginx::resource::server { $title:
+      server_name      => concat([$server_name], $server_aliases),
+      proxy            => $proxy,
+      proxy_redirect   => $proxy_redirect,
+      proxy_set_header => concat([
+        'Host $host',
+        'X-Forwarded-For $proxy_add_x_forwarded_for',
+        'X-Forwarded-Protocol $scheme',
+        'X-Real-IP $remote_addr',
+      ], $proxy_set_header),
```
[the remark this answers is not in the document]
jvperrin: That's also probably better, fair enough. I was thinking of doing this, like making a default options variable, but it collided with the `*` attribute, but making the default headers into a variable is good.

PR #249, opened 2017-11-28
Firewall input rules

kpengboy:


abizer:
I kind of feel like all the rules should live in a firewall.pp instead of being added to init.pp

kkuehler:
I agree with abizer. But we should only do this for modules that have more rules than a couple includes.

kpengboy:
I can agree on that logic for modules which are already split over several manifest files (e.g. `ocf_desktop`), but I'm not convinced it should apply to all other modules. We don't insist on putting every package declaration inside a `packages.pp` file, or every file declaration inside a `files.pp` file.

modules/ocf/manifests/firewall.pp
```diff
@@ -1,11 +1,15 @@
...
+  ) {
   # Install prerequisite packages (that is, netfilter-persistent)
   include firewall
 
   include ocf::firewall::chains
 
   class {
-    ['ocf::firewall::post']:
+    ['ocf::firewall::pre', 'ocf::firewall::post']:
       require => Class['ocf::firewall::chains'],
```
matthew-mcallister: Would you consider doing `require ocf::firewall::chains` above instead?

modules/ocf/manifests/firewall/allow_http.pp
```diff
@@ -0,0 +1,20 @@
+# firewall input rule to allow http and https
+class ocf::firewall::allow_http {
+  ocf::firewall::firewall46{
```
matthew-mcallister: Seems like you're missing a space here.

chriskuehl:
+1 for what @kpengboy said -- organize by logical component, not by resource type

even ocf_desktop i would expect to probably split rules out at least somewhat, e.g. the rule for allowing incoming mesos traffic might be added by `ocf_mesos::slave` or whatever.

modules/ocf/manifests/firewall/allow_http.pp
```diff
@@ -0,0 +1,20 @@
...
+    '101 allow https':
+      opts => {
+        chain  => 'PUPPET-INPUT',
+        proto  => 'tcp',
+        dport  => 'https',
+        action => 'accept',
+      };
+
+    '102 allow http':
```
matthew-mcallister: What is this sequentially numbered naming convention?

modules/ocf/manifests/firewall/post.pp
```diff
@@ -68,4 +70,20 @@
       before      => undef,
     }
   }
+
+
+  # Drop packets on the primary network inteface that are not whitelisted
+  # TODO: eliminate this if statement once testing is complete
+  if $ocf::firewall::drop_other_input {
```
matthew-mcallister: This is supposed to be non-optional in the end, right? I don't think it matters at this point, but I would have called it `allow_other_traffic` or something to indicate that the case is exceptional..

modules/ocf_admin/manifests/init.pp
```diff
@@ -35,4 +35,27 @@
       group  => ocfstaff,
       mode   => '0640';
   }
+
+  # Firewall Rules #
```
matthew-mcallister: It looks like this header comment is being used somewhat inconsistently. It seems unnecessary to me, but, as long as it's there, I would like it to show up in every `init.pp` with firewall rules and for it to have the same appearance everywhere (e.g. the trailing `#`, the before-and-after spacing).

modules/ocf_kvm/manifests/init.pp
```diff
@@ -51,4 +51,15 @@
       owner     => 'root',
       show_diff => false;
   }
+
+  # firewall input rules, allow all ssh, allow ntp
+  ocf::firewall::firewall46 {
+    '101 accept all ssh':
```
matthew-mcallister: Isn't this the same as `ocf::firewall::allow_ssh`? Also I don't see ntp here as described in the above comment.

modules/ocf_puppetdb/manifests/init.pp
```diff
@@ -28,4 +28,15 @@
     manage_firewall    => true,
     ssl_set_cert_paths => true,
   }
+
+  # firewall input, allow access to ports 8081
+  ocf::firewall::firewall46 {
+    '101 allow access to port 8081':
```
matthew-mcallister: Maybe comment what the port is for instead of just writing the number?

modules/ocf/manifests/firewall/allow_http.pp
chriskuehl: i don't think we need to sequentially number except in cases where order matters -- these can both just be 100, can't they?

matthew-mcallister: Do they need a numeric prefix if order doesn't matter?

chriskuehl: yes because there are rules at the start/end (often global ones) which need to be there

there are groups like these though where you can flip the order, they might as well just have the same prefix

modules/ocf/manifests/firewall.pp
kpengboy: That's probably a better idea

modules/ocf/manifests/firewall/post.pp
kpengboy: Sure, let's change it to that. We can rewrite the commit history if it gets too ugly anyway.

matthew-mcallister: If it's only test code that's ultimately going away, then I say it's your call whether to rename it.

modules/ocf/manifests/firewall.pp
```diff
@@ -1,13 +1,15 @@
...
+  # host-by-host basis.
+  $allow_other_traffic = true,
+  ) {
   # Install prerequisite packages (that is, netfilter-persistent)
   include firewall
+  include ocf::firewall::pre
+  include ocf::firewall::post
 
-  include ocf::firewall::chains
+  require ocf::firewall::chains
```
matthew-mcallister: I'm not sure if this works the way it's supposed to. That is, I don't think it's necessarily the case that the `require` statement here creates a before/after relationship with the `include` statements above. What would certainly work is to maintain the `class` statement below and just remove the `require` attribute. This is the case because `class` statements behave like resources, which are indeed subject to the before/after relationship on the containing class imposed by the `require` statement.

modules/ocf/manifests/packages/ntp.pp
```diff
@@ -8,6 +8,17 @@
       content => template('ocf/ntp.conf.erb'),
       require => Package['ntp'],
     }
+
+    #firewall input rule, allow ntp (123 udp/tcp)
+    ocf::firewall::firewall46 {
+      '101 accept all ntp':
+        opts => {
+          chain  => 'PUPPET-INPUT',
+          proto  => ['tcp', 'udp'],
```
kpengboy: Unfortunately you can't pass an array into here, you need to add these rules individually or by using `each`.

Ditto for every other place this pattern occurs.

modules/ocf_dhcp/manifests/init.pp
```diff
@@ -41,4 +41,45 @@
       minute  => '*/15',
       require => File['/usr/local/bin/lab-wakeup'];
   }
+
+  # Firewall Rules #
+
+  # Allow DNS
+  ocf::firewall::firewall46 {
+    '101 allow domain':
```
kpengboy: This rule doesn't belong in `ocf_dhcp`

modules/ocf_dhcp/manifests/init.pp
```diff
@@ -41,4 +41,45 @@
...
+  firewall { '101 allow bootps':
+    chain  => 'PUPPET-INPUT',
+    proto  => 'udp',
+    dport  => '67',
+    action => 'accept',
+  }
+
+  # Allow TFTP
+  ocf::firewall::firewall46 {
+    '101 allow tftp':
```
kpengboy: This rule should be in `netboot.pp`.

modules/ocf_ns/manifests/init.pp
```diff
@@ -31,4 +31,31 @@
...
+  ocf::firewall::firewall46 {
+    '101 allow domain':
+      opts => {
+        chain  => 'PUPPET-INPUT',
+        proto  => ['tcp', 'udp'],
+        dport  => '53',
+        action => 'accept',
+      };
+
+    '102 allow bootps':
```
kpengboy: These rule and the next are already defined in `ocf_dhcp`

modules/ocf/manifests/firewall/post.pp
```diff
@@ -20,7 +22,7 @@
       opts => {
```
kpengboy: @jvperrin @abizer @chriskuehl @matthew-mcallister I've been wondering if it's a better idea to simply drop packets from DMZed servers (i.e. tsunami, werewolves, corruption) instead. It would make the mesos rules much simpler and other rules unnecessary.

modules/ocf/manifests/firewall/post.pp
```diff
@@ -68,4 +70,20 @@
       before      => undef,
     }
   }
+
+
+  # Drop packets on the primary network inteface that are not whitelisted
+  # TODO: eliminate this if statement once testing is complete
+  if !$ocf::firewall::allow_other_traffic {
+    ocf::firewall::firewall46 {
+      '999 drop unrecognized input packets on primary interface':
```
kpengboy: @jvperrin @abizer @chriskuehl @matthew-mcallister I've been wondering if it's a better idea to simply drop packets from DMZed servers (i.e. tsunami, werewolves, corruption) instead. It would make the mesos rules much simpler and some other rules unnecessary. What do you all think?

matthew-mcallister: In general I'd say it's better to whitelist than to blacklist. It's more work, yes, but it's possible to forget to blacklist a new public machine. Also, we would not benefit from limiting traffic from desktops, staff VMs, and DHCP addresses.

matthew-mcallister:
Why switch to using numeric ports?

modules/ocf/manifests/firewall/allow_mosh.pp
```diff
@@ -0,0 +1,12 @@
+# firewall input rule to allow mosh
+class ocf::firewall::allow_mosh {
```
kpengboy: I'm not sure if this is the right abstraction, since it doesn't make sense to include `ocf::firewall::allow_mosh` without also including `ocf::firewall::allow_ssh`, as mosh creates the initial connection using ssh. I would instead create an `ocf::firewall::allow_ssh_and_mosh` class that includes `ocf::firewall::allow_ssh` and defines the additional rules for mosh. Then `ocf_ssh`, `ocf_apphost`, `ocf_admin`, and `ocf_kvm` can include `ocf::firewall::allow_ssh_and_mosh` instead.

matthew-mcallister: Or we could just move `allow_mosh` into `allow_ssh` without using a new name.

If we for some reason want to allow ssh without allowing mosh, we should simply add `require allow_ssh` to `allow_mosh`.

modules/ocf_admin/manifests/init.pp
kkuehler: This was my doing. I'm going to get rid of them. We already leave a comment before every rule which should be enough to indicate that what follows are firewall rules.

modules/ocf/manifests/firewall/allow_mosh.pp
kpengboy: Drawback of that though is it's not obvious from the name that `allow_mosh` also allows ssh

matthew-mcallister: I think it will be obvious to someone basically informed of what mosh is, which someone touching this code obviously should be.

kpengboy: No, I meant the forward proposition.

I on the contrary don't think it's a given that anyone reading these rules will already know what mosh is. They might be confused that hosts with the rule are allowing SSH, even though there's apparently no `allow_ssh` rule.

Having said that, we are getting into the bikeshed zone...

matthew-mcallister: > They might be confused that hosts with the rule are allowing SSH, even though there's apparently no allow_ssh rule.

That's possible, although I was envisioning a scenario in which one class would `include allow_ssh` and another class would include it and add `include allow_mosh` to that. How about we just merge `allow_mosh` into `allow_ssh` and relieve ourselves of this question completely?

modules/ocf/manifests/packages/ntp.pp
chriskuehl: imo this is something our firewall type should do for us

kpengboy: Hence my comment about [firewall_multi](https://forge.puppet.com/alexharvey/firewall_multi). I tried using that module, but unfortunately it doesn't forward dependencies, so it's not usable inside pre.pp or post.pp.

modules/ocf/manifests/firewall/nfs.pp
```diff
@@ -0,0 +1,69 @@
...
+      lens    => 'Shellvars.lns',
+      incl    => '/etc/default/quota',
+      changes => "set RPCRQUOTADOPTS '\"-p 32769\"'",
+      notify  => Service['quotarpc'];
+
+    'Set nfs-kernel-server port':
+      lens    => 'Shellvars.lns',
+      incl    => '/etc/default/nfs-kernel-server',
+      changes => "set RPCMOUNTDOPTS '\"-p 32767\"'",
+      notify  => Service['nfs-server'];
```
jvperrin: These also need to be enabled for `ocf_www`, since death runs a NFS server to share apache logs, and also needs the quota port to be open, otherwise `quota` hangs until it finally times out.

jvperrin:
One problem I've noticed so far is that the `PUPPET-INPUT` chain drops all IPv6 traffic at the end, but allows IPv4 traffic:

    DROP       all      anywhere             anywhere             /* 999 drop unrecognized input packets on primary interface (IPv6) */

(this line is only present in `ip6tables -L` output, not in `iptables -L` output). I noticed this when trying to SSH to syslog (currently on @kpengboy's environment) from my staff VM, which worked, but only after IPv6 timed out in connecting.

Going along with that, I think SSH connections to other servers should be allowed from staff VMs too. Honestly since we have unified logins across all our servers (except overheat and maybe some extra accounts here and there), and some servers are open to the world to connect, I think opening up SSH pretty much everywhere has almost no impact on security and significantly helps make staff less frustrated when connecting to things if we don't have to connect back to the same bastion host every time.

modules/ocf/manifests/firewall/nfs.pp
kpengboy: I believe this has been done in `ocf_www/manifests/logging.pp`?

jvperrin: It appears so, but something is still definitely blocking quota lookups from `www` in the firewall, so maybe something on @kkuehler's puppet environment?

kkuehler: To my knowledge, my environment is only running on mirrors and is up to date with master + some systemd timer stuff.

jvperrin: Ok, so I went on `death` and found that the `statd` service wasn't listening on the correct ports, so I restarted it and now running `quota` does not hang, although it still gets a connection refused only on tsunami (not on supernova). I couldn't see an instance of `rquotad` running, so I'm not sure why that is, but I think that is pretty normal, there shouldn't be quotas on the apache logs on death anyway.

modules/ocf/manifests/firewall/post.pp
kpengboy: I think that, in the rush to find all of the firewall rules we need, we may have lost sight of the original intent of this project.

My original intention with this firewall project was to place non-internal servers (i.e. servers running arbitrary user code, the DHCP range, and staff VMs) inside a DMZ. After all, we don't want malicious users on those servers attacking our internal servers, and intra-network firewalling isn't a function our existing hardware firewall can provide.

How to implement such a DMZ then? The most obvious answer is to start with one rule: internal hosts drop traffic from DMZ'ed hosts.

It doesn't take long to see that that one rule alone won't do. DMZed hosts still have to access many services on those internal hosts, such as LDAP, Kerberos, Puppet, etc. Nor would we want to prevent tsunami from accessing sites on www.ocf.berkeley.edu. Clearly, we need to whitelist those services.

Which brings us to attempt 2: on each host, we'll whitelist the ports used by actual services and drop all other traffic. Where that other traffic comes from is not considered. In this model, there is no DMZ; all hosts (except for the explicitly whitelisted supernova + hypervisors) are treated as untrusted. (This is the model we're using at this moment.)

This seemed to work for a while, but eventually problems started to show. For one, some of the rules turned out to be incredibly complicated. Take `ocf_mesos` for instance. The current list of planned rules for that module is a nightmare. There are 12 rules we need so far, and there might be even more rules needed that we haven't discovered yet. Worse, Mesos/Marathon/etc. ports are highly dynamic, which makes it difficult to find out exactly which port ranges are being used. This is a fragile, confusing, and not easily maintainable solution.

A resolution to this problem comes in the observation that not all of these ports are actually supposed to be accessible by all hosts. In particular, non-internal machines only need to access the http and https ports. If, instead of dropping traffic from all hosts by default, the internal firewalls only drop traffic from non-internal hosts by default (i.e. if non-internal hosts are placed into a DMZ), we can simplify the entire set of `ocf_mesos` firewall rules to a simple `import ocf::firewall::allow_http`. How much simpler that is! We would also similarly be able to get rid of a few other rules in other modules too.

Now, you might point out that, instead of changing our drop-traffic rule, we could instead solve our problems with `ocf_mesos` by allowing unrestricted traffic from every Mesos host + supernova + maybe desktops to every other Mesos host. That is true; however, introducing such a rule would create a discontinuity between the Mesos hosts and all other hosts. Why should Mesos hosts allow unrestricted traffic from all other Mesos hosts, when our model for every other host is to only open up a certain number of ports?

There are also conceptual considerations to take into account. Our current model is difficult to explain conceptually -- hosts firewall off exactly the set of ports that no service is listening on, so what's the point of having internal firewalls? It's also not obvious either why the internal firewalls don't render the external firewall redundant. (Eliminating the external firewall is a non-goal of this project.) On the contrary, the DMZ trust model is easy to explain: if it's not an internal server, it's in the DMZ and considered untrusted; otherwise it's trusted. The DMZ model also lends itself to a natural explanation of how this project relates to the external firewall: the external firewall is responsible for filtering traffic from outside the network, while the internal firewalls are responsible for filtering traffic only from within the network, and there is no overlap of responsibility.

It's thus evident that the DMZ model is far superior to the current model. Migrating to the DMZ model would not be difficult, as most of the rules would actually stay the same. It's in the differences, however, where the DMZ model shines.

My current proposal of implementing this DMZ involves changing the set of hosts from which traffic is dropped by default. I would use both whitelisting and blacklisting. It would be better in theory to use only whitelisting, but unfortunately our DMZ'ed hosts don't form a contiguous IP range. My current plan is roughly "blacklist `{dev-,}{tsunami,death,werewolves}`, then whitelist `169.229.226.2 - 169.229.226.139`, then blacklist every other IP in our IP block". I would appreciate discussion on how best to implement this.

kpengboy:
I've never trusted staff VMs to be too secure. It's too easy for a random staffer to leave a vulnerable service running on their staff VM. There isn't any firewalling on staff VMs either.

I would thus put staff VMs in the DMZ and put the same restrictions on them as we do for tsunami and werewolves. (We already refuse SSH connections from tsunami and werewolves to other servers in `access.conf`; on a side note, I would replace that with a firewall rule.) What alternative SSH policy are you exactly proposing?

modules/ocf/manifests/firewall/post.pp
kpengboy: @jvperrin @abizer @chriskuehl @matthew-mcallister any more thoughts on this?

jvperrin: Thanks for the long write-up on this, it definitely helps explain some of the motivations and the current state of this project.

I agree having a DMZ would work decently well and might be a good starting point, but I don't know if having just having "trusted" and "untrusted" is enough for our purposes really, because that's essentially what the current external firewall does too (our subnet is trusted, everything external is not), and this system helps with that system, but it seems to have a lot of the same shortcomings if we just have trusted and untrusted pools of hosts and some open service ports. For instance, I'd say tsunami and desktops should be untrusted in general since everyone has access to them, however something like supernova or staff VMs, although staffers can have arbitrary services, should be somewhat more trusted than tsunami (allowing SSH to other servers for instance, but maybe nothing else for staff VMs?), so I feel like multiple levels of trust would make this system more flexible now and in the future. Also, I think having all internal servers open to each other makes that perhaps too open, although I do like the plan of making the rules simpler than they currently are if possible. I agree that splitting things up more granularly like this makes things more complicated, but I'm worried otherwise we won't be able to express the full complexity of our setup of different hosts if we just separate hosts into two groups.

modules/ocf/manifests/firewall/post.pp
```diff
@@ -68,4 +69,59 @@
...
+        chain  => 'PUPPET-INPUT',
+        proto  => ['tcp', 'udp'],
+        action => 'drop',
+        source => $s
+      }
+    }
+  }
+
+  firewall {
+    '998 allow from internal zone (IPv4)':
```
kpengboy: You'll also want to allow IPv6 from the internal zone too

kpengboy: You'll want to write an IPv6 version of this rule too

modules/ocf/manifests/firewall/post.pp
```diff
@@ -68,4 +69,59 @@
...
+      chain     => 'PUPPET-INPUT',
+      src_range => '169.229.226.5-169.229.226.90',
+      proto     => ['tcp', 'udp'],
+      action    => 'accept';
+
+    '998 allow ssh from desktops (IPv6)':
```
kpengboy: Consider modifying `ocf::firewall::allow_ssh` to take address ranges and using that instead.

modules/ocf/manifests/firewall/pre.pp
```diff
@@ -0,0 +1,73 @@
+class ocf::firewall::pre {
+  $firewall_defaults = {
+    require => undef,
+  }
+
+  firewall {
+    default:
+      * => $firewall_defaults;
```
kpengboy: Nitpick: there should be an extra blank line after here

modules/ocf_desktop/manifests/init.pp
```diff
@@ -22,4 +22,27 @@
...
+  # Allow HTTP and HTTPS
+  include ocf::firewall::allow_http
+
+  # Allow Steam login and Steam content
+  ocf::firewall::firewall46 {
+    '101 allow steam (tcp)':
+      opts => {
+        chain  => 'PUPPET-INPUT',
+        proto  => 'tcp',
+        dport  => ['27015-27030', 27036, 27037],
```
kpengboy: Nitpick: I would opt for consistence - if one of these has to be a string, then make these all strings. Same with below.

modules/ocf_dhcp/manifests/init.pp
```diff
@@ -53,4 +53,21 @@
...
+    dport  => 67,
+    action => 'accept',
+  }
+
+  # Allow DHCP Server (IPv6 only)
+  firewall { '101 allow dhcpv6-server':
+    provider => 'ip6tables',
+    chain    => 'PUPPET-INPUT',
+    proto    => 'udp',
+    dport    => 'dhcpv6-server',
```
kpengboy: This should be changed to 547 in line with our convention

modules/ocf_kerberos/manifests/init.pp
```diff
@@ -61,4 +61,35 @@
...
+  ocf::firewall::firewall46 {
+    '101 allow kpasswd':
+      opts => {
+        chain  => 'PUPPET-INPUT',
+        proto  => 'tcp',
+        dport  => 464,
+        action => 'accept',
+      };
+  }
+  # Allow Kerberos Admin
```
kpengboy: Perhaps we can limit this rule to desktops, I see no reason why people on tsunami/werewolves/death should be able to use `kadmin`

modules/ocf_kvm/manifests/init.pp
```diff
@@ -1,8 +1,10 @@
 class ocf_kvm($group = 'root') {
+  include ocf::firewall::allow_ssh
   include ocf::firewall::output_all
   include ocf::ipmi
   include ocf::tmpfs
 
+
```
kpengboy: Why

modules/ocf_kvm/manifests/init.pp
```diff
@@ -51,4 +53,5 @@
       owner     => 'root',
       show_diff => false;
   }
+
```
kpengboy: Same

modules/ocf_mail/manifests/firewall_input.pp
```diff
@@ -0,0 +1,20 @@
+# firewall input rules for ocf_mail, allow submission (587), allow smtp (25)
```
kpengboy: Not sure this needs to be a separate file - it's not a separate functional module (unlike `spam`, `site_vhost`, ...). Consider placing these rules in `init.pp`?

modules/ocf_syslog/manifests/init.pp
```diff
@@ -8,6 +8,24 @@
...
+  ocf::firewall::firewall46 {
+    '100 allow syslog on TCP':
+      opts => {
+        chain  => 'PUPPET-INPUT',
+        proto  => 'tcp',
+        dport  => 514,
+        action => 'accept',
+      };
+
+    '101 allow syslog on UDP':
```
kpengboy: I think we can combine this rule with the previous one using `proto => ['tcp', 'udp']`

modules/ocf/manifests/firewall/pre.pp
```diff
@@ -0,0 +1,73 @@
...
+      };
+  }
+
+  firewall {
+    default:
+      * => $firewall_defaults;
+
+    '002 allow ssh from desktops (IPv4)':
+      chain     => 'PUPPET-INPUT',
+      src_range => '169.229.226.100-169.229.226.139',
```
kpengboy: We might want to make these ranges hiera variables, that way it's easier to change all of them at once

modules/ocf/manifests/firewall/pre.pp
```diff
@@ -0,0 +1,73 @@
...
+        'state'  => ['RELATED', 'ESTABLISHED'],
+        'action' => 'accept',
+      };
+  }
+
+  firewall {
+    default:
+      * => $firewall_defaults;
+
+    '002 allow ssh from desktops (IPv4)':
```
kpengboy: One thing we can consider is modifying `ocf::firewall::allow_ssh` to do this for us.

modules/ocf/manifests/firewall/nfs.pp
```diff
@@ -0,0 +1,69 @@
...
+        action => 'accept',
+      };
+  }
+
+  # Allow SunRPC
+  ocf::firewall::firewall46 {
+    '101 allow sunrpc':
+      opts => {
+        chain  => 'PUPPET-INPUT',
+        proto  => ['tcp', 'udp'],
```
kpengboy: Also this doesn't work, we need to specify these individually

hieradata/common.yaml
```diff
@@ -21,3 +21,11 @@ mesos_slaves:
...
 
 browser_homepage: https://www.ocf.berkeley.edu/about/lab/open-source
+
+# IPv4 and IPv6 src ranges for types of servers
+internal_zone_range_4: 169.229.226.5-169.229.226.90
+desktop_src_range_4: 169.229.226.100-169.229.226.139
+staffvm_src_range_4: 169.229.226.200-169.229.226.252
+internal_zone_range_6: 2607:f140:8801::1:5-2607:f140:8801::1:90
+desktop_src_range_6: 2607:f140:8801::1:100-2607:f140:8801::1:139
+staffvm_src_range_6: 2607:f140:8801::1:200-2607:f140:8801::1:252
```
jvperrin: This bothers me a bit because the IPv6 ranges actually have gaps in them because of the unused hex digits (a-f) that we don't use to make the IPs map more easily to IPv4, but I don't know of a good way to represent that here.

modules/ocf/manifests/firewall/nfs.pp
kpengboy: Now it does

modules/ocf/manifests/packages/ntp.pp
kpengboy: Resolved

hieradata/common.yaml
kpengboy: True. I don't know of a good way either...

kpengboy:
This PR is now in the testing phase, so it's about time for it to get some scrutiny again.

kpengboy:
Alright, this branch is getting ready to be merged in. Review would be appreciated.

modules/ocf/manifests/firewall/pre.pp
```diff
@@ -0,0 +1,80 @@
...
+      dport     => 22,
+      action    => 'accept';
+
+    '003 allow ssh from staff VMs (IPv6)':
+      provider  => 'ip6tables',
+      chain     => 'PUPPET-INPUT',
+      src_range => $staffvm_src_range_6,
+      proto     => 'tcp',
+      dport     => 22,
+      action    => 'accept';
```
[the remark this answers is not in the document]
kpengboy: I think that it actually makes more sense to restrict SSH from certain hosts through firewalls rather than `access.conf`.

Although we could also consider blanket-allowing SSH, even from tsunami etc. Non-staff would still be prevented from logging in to internal servers due to `access.conf`. Does anyone else have thoughts?

hieradata/nodes/biohazard.yaml
```diff
@@ -1,2 +1,4 @@
 classes:
     - ocf_docker
+
+ocf::firewall::allow_other_traffic: false
```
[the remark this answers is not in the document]
kpengboy: Some of these rules have been tested on dev- hosts rather than production hosts...

modules/ocf/manifests/firewall/post.pp
```diff
@@ -23,47 +25,103 @@
...
-      action      => 'drop',
+      action      => 'reject',
       destination => $devices,
     },
     before => undef,
   }
+
+  # reject from hosts in internal zone range but not actually internal
+  # TODO: eliminate this if statement once testing is complete
+  if !$ocf::firewall::allow_other_traffic {
```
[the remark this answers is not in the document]
kpengboy: Well noted, though since this is meant to be temporary anyway, I guess it doesn't matter that much...

modules/ocf/manifests/firewall/pre.pp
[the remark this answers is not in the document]
kpengboy: As of now I would still rather restrict SSH access from the DMZ in line with what we do with the external firewall. Jason does have a valid argument though. Would anyone else like to voice their opinion?

kpengboy:
Any more comments? If not I will merge in 10 days

kkuehler:
@kpengboy right. fixed and force pushed.

kpengboy:
Any more comments? If not I will merge in 2 days. For real this time.

modules/ocf/manifests/firewall.pp
```diff
@@ -1,13 +1,14 @@
-class ocf::firewall {
+class ocf::firewall(
+  $reject_unrecognized_input = true,
```
kpengboy: What do people think of this name? +? -? Δ?

PR #250, opened 2017-11-28
quickfix enforcer to work with new mysql api

abizer:
this is completely untested


sahilhasan:
@abizer never admit you are mortal :P

PR #252, opened 2017-11-29
session tracking -> ocfweb

abizer:
WIP

I'm going to let this continue testing for a few days but feedback is appreciated

### DIFF
--- a/modules/ocf_desktop/files/stats/update.sh
+++ b/modules/ocf_desktop/files/stats/update.sh
@@ -7,9 +7,6 @@ CUR_USER=$(who | awk '$NF == "(:0)" { print $1 }')
 DATA="state=inactive"
 
 if [ -n "$CUR_USER" ]; then
-	DATA="state=active&user=$CUR_USER"
+    curl -H "Content-Type: application/json" -X POST -d "{\"user\": \"$CUR_USER\"}" \
+         https://ocf.berkeley.edu/api/session/log 2>/dev/null
 fi
-
-curl --data "$DATA" \
-    https://labstats.ocf.berkeley.edu:444/update.cgi \
-	2>/dev/null
--- a/modules/ocf_ocfweb/manifests/dev_config.pp
+++ b/modules/ocf_ocfweb/manifests/dev_config.pp
@@ -3,9 +3,11 @@
 # staff VMs, etc.
 class ocf_ocfweb::dev_config($group = 'ocfstaff') {
   include ocf::firewall::output_printers
+
   # TODO: stop copy-pasting this password validation everywhere
   $redis_password = assert_type(Pattern[/^[a-zA-Z0-9]*$/], hiera('create::redis::password'))
   $ocfmail_password = assert_type(Pattern[/^[a-zA-Z0-9]*$/], hiera('ocfmail::mysql::dev_password'))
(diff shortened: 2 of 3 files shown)


modules/ocf_desktop/files/stats/update.sh
```diff
@@ -4,12 +4,8 @@
 [ "$(imvirt 2>/dev/null)" == "Physical" ] || exit 0 # only count hours on desktops
 
 CUR_USER=$(who | awk '$5 == "(:0)" { print $1 }')
-DATA="state=inactive"
 
 if [ -n "$CUR_USER" ]; then
-	DATA="state=active&user=$CUR_USER"
+    curl -H "Content-Type: application/json" -X POST -d "{\"user\": \"$CUR_USER\"}" \
+         https://ocf.berkeley.edu/api/session/log 2>/dev/null
```
chriskuehl: www

modules/ocf_desktop/files/stats/update.sh
```diff
@@ -4,12 +4,8 @@
 [ "$(imvirt 2>/dev/null)" == "Physical" ] || exit 0 # only count hours on desktops
 
 CUR_USER=$(who | awk '$5 == "(:0)" { print $1 }')
-DATA="state=inactive"
```
jvperrin: Doesn't removing this mean that it will rely on the 3 minute timeout to close all sessions now? That seems like something that's not great compared to sending an update saying that nobody is logged in when the session has ended. This also means that some sessions might overlap if someone logs out and then someone quickly logs in, right?

jvperrin: I noticed that when a new session is made, the old one is closed, so there won't be overlaps, but I believe the 3 minute thing is still a problem when nobody logs in after a user logs out.

modules/ocf_ocfweb/manifests/dev_config.pp
```diff
@@ -7,7 +7,10 @@
   $redis_password = hiera('create::redis::password')
   validate_re($redis_password, '^[a-zA-Z0-9]*$', 'Bad Redis password')
   $ocfmail_password = hiera('ocfmail::mysql::dev_password')
-  validate_re($ocfmail_password, '^[a-zA-Z0-9]*$', 'Bad ocfmail password')
+  validate_re($ocfmail_password, '^[a-zA-Z0-9]*$', 'Bad ocfmail dev password')
+
```
kpengboy: Why this extra blank line?

modules/ocf_desktop/files/stats/update.sh
kpengboy: I agree it would be nicer to eagerly clean up sessions when they are known to have ended.

abizer: I think shichenh would like eagerly cleaned up sessions as well.

modules/ocf_ocfweb/manifests/dev_config.pp
```diff
@@ -7,7 +7,10 @@
...
+  $ocfstats_password = hiera('ocfstats::mysql::dev_password')
+  validate_re($ocfstats_password, '^[a-zA-Z0-9]*$', 'Bad ocfstats dev password')
```
jvperrin: Unfortunately these all should follow the new format (https://github.com/ocf/puppet/commit/33b89c0159c3019d4fac0d06165618d12a089340) now.

PR #254, opened 2017-12-01
add support for fish shell

Baisang:
seems to work

### DIFF
--- a/modules/ocf/files/shell/config.fish
+++ b/modules/ocf/files/shell/config.fish
@@ -1,0 +1,38 @@
+# Quit if not interactive
+if not status --is-interactive
+    exit
+end
+
+# OCF environment variables
+set -x PATH /opt/share/utils/bin \
+    /opt/share/utils/sbin \
+    /usr/local/bin \
+    /usr/local/sbin \
+    /bin \
+    /usr/bin \
+    /usr/sbin \
+    /usr/lib \
+    /sbin \
+    /usr/games
+
+if test -e /opt/puppetlabs/bin
+    set -x PATH $PATH /opt/puppetlabs/bin
+end
+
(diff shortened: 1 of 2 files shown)


modules/ocf/files/shell/config.fish
```diff
@@ -0,0 +1,38 @@
+# Quit if not interactive
+if not status --is-interactive
+    exit
+end
```
chriskuehl: i don't know anything about fish, but if exit means the same thing as it does in sh, i wonder if this causes the entire shell to exit? lol

Baisang: oh shit you're right lmao

Baisang: oh wait https://fishshell.com/docs/current/commands.html#exit

`If exit is called while sourcing a file (using the source builtin) the rest of the file will be skipped, but the shell itself will not exit.`

Does that seem to be intended behavior?

chriskuehl: yeah, that sounds right (different than the sh behavior, but that's ok)

modules/ocf/manifests/packages/shell.pp
```diff
@@ -1,5 +1,5 @@
 class ocf::packages::shell {
-  package { ['bash', 'tcsh', 'zsh']: }
+  package { ['bash', 'tcsh', 'zsh', 'fish']: }
```
abizer: fish

matthew-mcallister:
Duplicate this in [ocflib](https://github.com/ocf/ocflib/blob/master/ocflib/misc/validators.py#L45-L54)?

redplanks:
So long and thanks for all the zsh

PR #259, opened 2017-12-07
Mirror KDE releases and application data.

harrisonkuo:


modules/ocf_mirrors/manifests/kde.pp
```diff
@@ -0,0 +1,18 @@
...
+      group   => mirrors,
+      mode    => '0755',
+      recurse => true;
+  }
+
+  cron {
+    'kde':
+      command => '/opt/mirrors/project/kde/sync-archive > /dev/null',
+      user    => 'mirrors',
+      hour    => '*/2',
```
jvperrin: I think putting `minute` as `0` explicitly here is a good idea, just to make it clear.

modules/ocf_mirrors/files/project/kde/sync-archive
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh -eu
+/usr/local/bin/rsync-no-vanished -az --delete --safe-links \
+	rsync.kde.org::kdeftp /opt/mirrors/ftp/kde
+/usr/local/bin/rsync-no-vanished -az --delete --safe-links \
+	rsync.kde.org::applicationdata /opt/mirrors/ftp/kde-applicationdata
```
jvperrin: I think it might make sense to mirror from `rsync://rsync.kde.org/alldata` instead, so that you then mirror both in one command essentially. Then, when you sign up as a mirror, the `kde` stuff is under `ftp/`, and `applicationdata` is under `files/`, which will clean up the mirror root directory too. Also, you wouldn't have to resync, since you'd just move the directories to the correct place and sync once again. You can do `rsync --list-only rsync://rsync.kde.org/alldata/` to list the files available to sync.

modules/ocf_mirrors/files/rsyncd.conf
```diff
@@ -57,6 +57,14 @@ dont compress = *.gz *.tgz *.zip *.z *.rpm *.deb *.iso *.bz2 *.tbz
 	comment = kali-images mirror
 	path = /opt/mirrors/ftp/kali-images
 
+[kde]
+	comment = kde mirror
+	path = /opt/mirrors/ftp/kde
+
+[kde-applicationdata]
+	comment = kde-applicationdata mirror
+	path = /opt/mirrors/ftp/kde-applicationdata
```
jvperrin: If you combine, you'd want to combine here too, although I don't know if they'd want to have these two separate or not, might be best to ask when you send an email asking to join their mirror list.

PR #260, opened 2017-12-07
mirrors systemd timers

kkuehler:
Looking for feedback.

I opted to create the service and timer together in a resource file since the unit file is activated by the timer of the same name, so this should ultimately lead to less code in the future.

I'm not sure if it's worth having the user pass in the template files to the timer, but for now I followed the form of the systemd::service.

### DIFF
--- a/modules/ocf/manifests/systemd/service.pp
+++ b/modules/ocf/manifests/systemd/service.pp
@@ -18,10 +18,10 @@ define ocf::systemd::service(
     require => Package['systemd-sysv'],
   } ~>
   service { $title:
-    ensure    => $ensure,
-    enable    => $enable,
-    provider  => systemd,
-    require   => [
+    ensure   => $ensure,
+    enable   => $enable,
+    provider => systemd,
+    require  => [
       Package['systemd-sysv'],
       File["/etc/systemd/system/${title}.service"],
       Exec['systemd-reload'],
--- a/modules/ocf/manifests/systemd/service.pp
+++ b/modules/ocf/manifests/systemd/service.pp
@@ -4,18 +4,23 @@ define ocf::systemd::service(
   $ensure = 'running',
   $enable = true,
 ) {
+  if $content and $source {
+    fail('You may not supply both content and source parameters to ocf::systemd::service')
+  }
+  elsif $content == undef and $source == undef {
+    fail('You must supply either the content or source parameter to ocf::systemd::service')
+  }
+
   file { "/etc/systemd/system/${title}.service":
     source  => $source,
     content => $content,
     notify  => Exec['systemd-reload'],
     require => Package['systemd-sysv'],
-  }
-
+  } ~>
   service { $title:
     ensure    => $ensure,
     enable    => $enable,
     provider  => systemd,
-    subscribe => File["/etc/systemd/system/${title}.service"],
     require   => [
       Package['systemd-sysv'],
       File["/etc/systemd/system/${title}.service"],
--- a/modules/ocf/manifests/systemd/timer.pp
+++ b/modules/ocf/manifests/systemd/timer.pp
@@ -1,0 +1,51 @@
+# Creates two systemd units: a timer and service bearing the same
+# name, whereby the service will be activated once the timer elapses.
+define ocf::systemd::timer(
+  $timer_content = undef,
+  $service_content = undef,
+  $timer_source = undef,
+  $service_source = undef,
+  $ensure = 'running',
+  $enable = true,
+) {
+  # ocf::systemd::service will handle multiple timer sources
+  if $service_content and $service_source {
+    fail('You may not supply both content and source parameters to ocf::systemd::timer')
+  }
+  elsif $service_content == undef and $service_source == undef {
+    fail('You must supply either the content or source parameter to ocf::systemd::timer')
+  }
+  elsif $timer_content and $timer_source {
+    fail('You may not supply both content and source parameters to ocf::systemd::timer')
+  }
+  elsif $timer_content == undef and $timer_source == undef {
+    fail('You must supply either the content or source parameter to ocf::systemd::timer')
+  }
+
+  file {
+    "/etc/systemd/system/${title}.timer":
+      source  => $timer_source,
+      content => $timer_content,
+      notify  => Exec['systemd-reload'],
+      require => Package['systemd-sysv'];
+
+    "/etc/systemd/system/${title}.service":
+      source  => $service_source,
+      content => $service_content,
+      notify  => Exec['systemd-reload'],
+      require => [
+        Package['systemd-sysv'],
+        File["/etc/systemd/system/${title}.timer"],
+      ];
+  } ~>
+  service { "${title}.timer":
+    ensure   => $ensure,
+    enable   => $enable,
+    provider => systemd,
+    require  => [
+      Package['systemd-sysv'],
+      File["/etc/systemd/system/${title}.timer"],
+      Exec['systemd-reload'],
+    ],
+  }
+}
--- a/modules/ocf_mirrors/manifests/centos.pp
+++ b/modules/ocf_mirrors/manifests/centos.pp
@@ -15,11 +15,10 @@ class ocf_mirrors::centos {
       ts_path       => 'TIME';
   }
 
-  cron {
+  ocf_mirrors::timer {
     'centos':
-      command => '/opt/mirrors/project/centos/sync-archive > /dev/null',
-      user    => 'mirrors',
-      hour    => '*/3',
-      minute  => '22';
+      exec_start => '/opt/mirrors/project/centos/sync-archive',
+      hour       => '0/3',
+      minute     => '22';
   }
 }
--- a/modules/ocf_mirrors/manifests/centos_altarch.pp
+++ b/modules/ocf_mirrors/manifests/centos_altarch.pp
@@ -20,6 +20,6 @@ class ocf_mirrors::centos_altarch {
     'centos-altarch':
       exec_start => '/opt/mirrors/project/centos-altarch/sync-archive',
       hour       => '0/3',
-      minute     => '44';
+      minute     => '44',
   }
 }
--- a/modules/ocf_mirrors/manifests/centos_altarch.pp
+++ b/modules/ocf_mirrors/manifests/centos_altarch.pp
@@ -16,11 +16,10 @@ class ocf_mirrors::centos_altarch {
       ts_path       => 'TIME';
   }
 
-  cron {
+  ocf_mirrors::timer {
     'centos-altarch':
-      command => '/opt/mirrors/project/centos-altarch/sync-archive > /dev/null',
-      user    => 'mirrors',
-      hour    => '*/3',
-      minute  => '44';
+      exec_start => '/opt/mirrors/project/centos-altarch/sync-archive',
+      hour       => '0/3',
+      minute     => '44';
   }
 }
--- a/modules/ocf_mirrors/manifests/ftpsync.pp
+++ b/modules/ocf_mirrors/manifests/ftpsync.pp
@@ -45,14 +45,13 @@ define ocf_mirrors::ftpsync(
 
   if $use_systemd {
     ocf_mirrors::timer { "ftpsync-${title}":
-      exec_start   => "${project_path}/bin/ftpsync > /dev/null 2>&1",
+      exec_start   => "${project_path}/bin/ftpsync",
       environments => { 'BASEDIR' => $project_path },
       hour         => $cron_hour,
       minute       => $cron_minute,
       require      => File["${project_path}/bin", "${project_path}/etc/ftpsync.conf"],
     }
-  }
-  else {
+  } else {
     cron { "ftpsync-${title}":
       command => "BASEDIR=${project_path} ${project_path}/bin/ftpsync > /dev/null 2>&1",
       user    => 'mirrors',
--- a/modules/ocf_mirrors/manifests/ftpsync.pp
+++ b/modules/ocf_mirrors/manifests/ftpsync.pp
@@ -11,7 +11,7 @@ define ocf_mirrors::ftpsync(
     $mirror_name = 'mirrors.ocf.berkeley.edu',
     $mirror_path = "/opt/mirrors/ftp/${title}",
     $project_path = "/opt/mirrors/project/${title}",
-    $use_systemd = false,
+    $use_systemd = true,
   ) {
 
   file {
--- a/modules/ocf_mirrors/manifests/ftpsync.pp
+++ b/modules/ocf_mirrors/manifests/ftpsync.pp
@@ -11,6 +11,7 @@ define ocf_mirrors::ftpsync(
     $mirror_name = 'mirrors.ocf.berkeley.edu',
     $mirror_path = "/opt/mirrors/ftp/${title}",
     $project_path = "/opt/mirrors/project/${title}",
+    $use_systemd = false,
   ) {
 
   file {
@@ -42,11 +43,22 @@ define ocf_mirrors::ftpsync(
     require => File[$project_path];
   }
 
-  cron { "ftpsync-${title}":
-    command => "BASEDIR=${project_path} ${project_path}/bin/ftpsync > /dev/null 2>&1",
-    user    => 'mirrors',
-    minute  => $cron_minute,
-    hour    => $cron_hour,
-    require => File["${project_path}/bin", "${project_path}/etc/ftpsync.conf"];
+  if $use_systemd {
+    ocf_mirrors::timer { "ftpsync-${title}":
+      exec_start   => "${project_path}/bin/ftpsync > /dev/null 2>&1",
+      environments => { 'BASEDIR' => $project_path },
+      hour         => $cron_hour,
+      minute       => $cron_minute,
+      require      => File["${project_path}/bin", "${project_path}/etc/ftpsync.conf"],
+    }
+  }
+  else {
+    cron { "ftpsync-${title}":
+      command => "BASEDIR=${project_path} ${project_path}/bin/ftpsync > /dev/null 2>&1",
+      user    => 'mirrors',
+      minute  => $cron_minute,
+      hour    => $cron_hour,
+      require => File["${project_path}/bin", "${project_path}/etc/ftpsync.conf"];
+    }
   }
 }
--- a/modules/ocf_mirrors/templates/systemd/sync.service.erb
+++ b/modules/ocf_mirrors/templates/systemd/sync.service.erb
@@ -1,0 +1,9 @@
+[Unit]
+Description=Sync service for <%= @title %>
+
+[Service]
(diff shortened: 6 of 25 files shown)


kkuehler:
With some minor tweaking to ocf::systemd::service (that is allowing a suffix and setting the default to be .service), the *.timer logic can be moved there.

modules/ocf_mirrors/templates/sync.timer.erb
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Timer to sync %i mirrors
+After=network.target
+Wants=network.target
+
+[Timer]
+OnCalendar=<%= @on_calendar %>
+#Persistent=true
```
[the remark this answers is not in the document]
kkuehler: Yeah. Forgot about that. We definitely don't want persistent.

modules/ocf/manifests/systemd/timer.pp
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  service { "${title}.timer":
+    ensure    => running,
+    enable    => true,
+    provider  => systemd,
+    subscribe => [
+      File["/etc/systemd/system/${title}.timer"],
+      File["/etc/systemd/system/${title}.service"],
+    ],
```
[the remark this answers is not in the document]
kkuehler: Yep. All of the timer logic could also be handled by passing a template to ocf::systemd::service. Thoughts on doing this? The only difference is we would only be able to subscribe to the *.timer file, but I'm not sure if that even matters.

modules/ocf/manifests/systemd/timer.pp
```diff
@@ -0,0 +1,48 @@
+define ocf::systemd::timer(
+    $minute = '*',
+    $hour   = '*',
+    $day    = '*',
+    $month  = '*',
+    $year   = '*',
+    # For reasonable second precision,
+    # AccuracySec will need to be set
```
kpengboy: Nitpick: this is wrapped weirdly

modules/ocf/manifests/systemd/timer.pp
```diff
@@ -0,0 +1,48 @@
...
+    default:
+      owner => root,
+      group => root,
+      mode  => '0644';
+
+    "/etc/systemd/system/${title}.timer":
+      content => template($timer_template),
+      source  => $timer_source;
+
+    "/etc/systemd/system/${title}.service":
```
kpengboy: I would suggest using `ocf::systemd::service` here instead.

modules/ocf_mirrors/templates/sync.service.erb
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Sync releases for the %i project
```
kpengboy: I think `%i` should say something like `<%= @title %>` instead? systemd templates are not puppet templates.

modules/ocf_mirrors/templates/sync.timer.erb
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Timer to sync %i mirrors
```
kpengboy: ibid

modules/ocf_mirrors/manifests/ftpsync.pp
```diff
@@ -11,6 +11,7 @@
     $mirror_name = 'mirrors.ocf.berkeley.edu',
     $mirror_path = "/opt/mirrors/ftp/${title}",
     $project_path = "/opt/mirrors/project/${title}",
+    $nocron = false,
```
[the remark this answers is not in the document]
kkuehler: This is just temporary anyway.

modules/ocf_mirrors/templates/sync.service.erb
kkuehler: Yes, you are correct.

modules/ocf/manifests/systemd/timer.pp
kkuehler: Yeah, I was thinking the same thing. Declare the service file first and ocf::systemd::service will create the timer file and start the service.

kpengboy: ? I was saying that instead of creating a raw service file here, you can make use of the existing `ocf::systemd::service`.

kkuehler: No. We don't want to use that because ocf::systemd::service starts a service for us, which we do not want. The timer should start the service for us. From the systemd.timer manpage:

>For each timer file, a matching unit file must exist, describing the unit to activate when
>the timer elapses. By default, a service by the same name as the timer (except for the
>suffix) is activated.

kkuehler: Ah, I see. Are you suggesting to set it up like this?

```enable => false,
    ensure => 'stopped',
```
If so, I'm on board with this idea.

kpengboy: > No. We don't want to use that because ocf::systemd::service starts a service for us, which we do not want.

Ah right, never mind then. Nah, we don't want to ensure stopped, that would stop the service even when we wanted it to run.

kkuehler:
@chriskuehl @kpengboy changed to a newer design that i'm much happier with.

kkuehler:
I also wonder if both ocf::systemd::service and ocf::systemd::timer should have if->fail statements that verify *both* the template and source are not undef.

modules/ocf_mirrors/templates/systemd/sync.service.erb
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Sync service for <%= @title %>
+
+[Service]
+User=<%= @exec_user %>
+Type=simple
+ExecStart=<%= @exec_start %>
```
kpengboy: You might want to set up email so that we continue to get emails when systemd timer-controlled jobs fail

modules/ocf/manifests/systemd/service.pp
```diff
@@ -3,22 +3,20 @@
...
   $ensure = 'running',
   $enable = true,
+  $suffix = 'service',
 ) {
-  file { "/etc/systemd/system/${title}.service":
+  file { "/etc/systemd/system/${title}.${suffix}":
     source   => $source,
     content  => $content,
     notify   => Exec['systemd-reload'],
-    require  => Package['systemd-sysv'],
```
kpengboy: Why is this line being removed?

modules/ocf_mirrors/templates/systemd/sync.service.erb
chriskuehl: should be optional imo, and maybe not even needed for any of the mirrors cronjobs

a while back we decided that we just can't deal with cronjob spam from mirrors, they are just too inherently flaky (we're talking to ~a dozen servers that we don't control, they're gonna have downtime)

instead we took the approach of implementing healthchecks so we only get notified when it has failed consistently for a while, i think this is a much better approach for mirrors

(some other cronjobs make sense for emails though)

modules/ocf/manifests/systemd/service.pp
kkuehler: We talked about it in the lab, just putting it here in case future reference is necessary. Exec['systemd-reload'] already handles this dependency.

kkuehler: Given that you pointed out this is not immediately obvious, I'm not against keeping it in.

kkuehler:
@kpengboy, @chriskuehl any issues with this? if not, i'd like to start moving the other mirrors over to systemd timers.

kkuehler:
Here's an alternative to the timer.pp I currently have. It works exactly the same but doesn't use ocf::systemd::service to create and start the timer. 

https://gist.github.com/kkuehler/f0664592519900ff4fd5df93508d86bd

I personally prefer the design in place.

modules/ocf_mirrors/manifests/ftpsync.pp
```diff
@@ -42,11 +43,22 @@
...
-  cron { "ftpsync-${title}":
-    command => "BASEDIR=${project_path} ${project_path}/bin/ftpsync > /dev/null 2>&1",
-    user    => 'mirrors',
-    minute  => $cron_minute,
-    hour    => $cron_hour,
-    require => File["${project_path}/bin", "${project_path}/etc/ftpsync.conf"];
+  # TODO: Remove. Testing Only!
+  if $use_systemd {
+    ocf_mirrors::timer { "ftpsync-${title}":
+      exec_start => "/bin/bash -c 'BASEDIR=${project_path} ${project_path}/bin/ftpsync > /dev/null 2>&1'",
```
jvperrin: Does this have to execute `bash` with a command like this? Can we just have the environment variable put into the systemd service file and then run it directly instead? I know that's supported with the `Environment` key in the unit file at least but I haven't tried that with timers specifically, but if it works it seems like it would be a bit cleaner.

modules/ocf_mirrors/manifests/centos.pp
```diff
@@ -8,17 +8,17 @@
...
 
-  cron {
+  ocf_mirrors::timer {
     'centos':
-      command => '/opt/mirrors/project/centos/sync-archive > /dev/null',
-      user    => 'mirrors',
-      hour    => '*/3',
-      minute  => '22';
+      exec_start => '/opt/mirrors/project/centos/sync-archive',
+      stdout     => 'null',
```
jvperrin: I think we'd actually want to see the stdout get logged, given that this won't email us now. We'd also need to have some kind of health checks on all our mirrors (not just the ftpsync ones) so that we can tell when one has been failing for a while.

modules/ocf_mirrors/templates/systemd/sync.service.erb
jvperrin: I think for this we'd want to not have output as @chriskuehl said, since they can be pretty spammy if the upstream server is down. However, we'd definitely need health checks on all mirrors first, because otherwise we'd have no way of knowing if one of the non-ftpsync ones are way behind.

modules/ocf/manifests/systemd/service.pp
```diff
@@ -3,22 +3,27 @@
...
   $enable = true,
+  $suffix = 'service',
 ) {
-  file { "/etc/systemd/system/${title}.service":
+  if $content and $source {
+    fail('You may not supply both content and source parameters to ocf::systemd::service')
+  }
+  elsif $content == undef and $source == undef {
+    fail('You must supply either the content or source parameter to ocf::systemd::service')
+  }
```
jvperrin: Nice, this is quite good to have, and honestly we should have more checks like this on any of our custom definitions (like `ocf::repackage` for example).

modules/ocf/manifests/systemd/timer.pp
```diff
@@ -0,0 +1,27 @@
...
+  $service_source = undef,
+) {
+  # ocf::systemd::service will handle multiple timer sources
+  if $service_content and $service_source {
+    fail('You may not supply both content and source parameters to ocf::systemd::timer')
+  }
+  elsif $service_content == undef and $service_source == undef {
+    fail('You must supply either the content or source parameter to ocf::systemd::timer')
+  }
+  file { "/etc/systemd/system/${title}.service":
```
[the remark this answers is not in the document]
kkuehler: Nope, I modified ocf::systemd::service to take a custom suffix. The timer is created there and the service is started; once that elapses the service file with the same name is activated. 

It essentially does the same thing as this https://gist.github.com/kkuehler/f0664592519900ff4fd5df93508d86bd.

I've actually been debating between using the current design and the one I linked above. Opinions are welcome.

modules/ocf/manifests/systemd/service.pp
```diff
@@ -3,22 +3,27 @@
...
-    require  => Package['systemd-sysv'],
-  }
-
-  service { $title:
+  }~>
+  service { "${title}.${suffix}":
     ensure    => $ensure,
     enable    => $enable,
     provider  => systemd,
-    subscribe => File["/etc/systemd/system/${title}.service"],
```
jvperrin: When the service's unit file is changed, doesn't removing this mean that it won't restart the service at all any more?

modules/ocf/manifests/systemd/timer.pp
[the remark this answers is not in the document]
jvperrin: Ah, good point, I saw the suffix but forgot about that, I think that works fine then.

modules/ocf_mirrors/manifests/ftpsync.pp
kkuehler: The ```exec_start``` is put in a service file. Part of the timer abstraction is that it creates both a timer and service file, and activates the service once the timer elapses. So, in short, yes we can do this-it just needs to be added to the template.

jvperrin: Ok, yeah I think having an environment parameter that is passed to `ocf::systemd::service` would be good and could be used in other services. It would have to be an array or something, because you can specify multiple environment variables, but I don't think that would be too bad, it would just have to iterate through it in the template (erb is fun lol).

modules/ocf/manifests/systemd/service.pp
kkuehler: I changed that to a notify arrow, ```~>```, because it defines that relationship (I think) in a much cleaner manner.

jvperrin: Ah yeah, so you did. I'm not very used to using that format, this looks good then.

modules/ocf/manifests/systemd/timer.pp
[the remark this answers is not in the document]
kpengboy: I still don't know if it's the best idea to use `ocf::systemd::service` to manage timers. I'm worried it could lead to confusion because in this case we use `ocf::systemd::service` to manage the timer unit and not the actual service unit. I'm not sure that generalizing `ocf::systemd::service` is thus the right way to go here. Does anyone else have thoughts?

modules/ocf/manifests/systemd/timer.pp
```diff
@@ -0,0 +1,27 @@
...
+  $timer_source = undef,
+  $service_source = undef,
+) {
+  # ocf::systemd::service will handle multiple timer sources
+  if $service_content and $service_source {
+    fail('You may not supply both content and source parameters to ocf::systemd::timer')
+  }
+  elsif $service_content == undef and $service_source == undef {
+    fail('You must supply either the content or source parameter to ocf::systemd::timer')
+  }
```
kpengboy: Add the same checks for `$timer_content` and `$timer_source`?

kkuehler:
@jvperrin @chriskuehl @kpengboy I've finally revisited this PR and tried to address the changes ya'll suggested.

modules/ocf/manifests/systemd/timer.pp
```diff
@@ -0,0 +1,51 @@
...
+
+    "/etc/systemd/system/${title}.service":
+      source  => $service_source,
+      content => $service_content,
+      notify  => Exec['systemd-reload'],
+      require => [
+        Package['systemd-sysv'],
+        File["/etc/systemd/system/${title}.timer"],
+      ];
+  }~>
```
jvperrin: nit: add a space between `}` and `~>`

modules/ocf_mirrors/manifests/ftpsync.pp
```diff
@@ -42,11 +43,23 @@
...
+  if $use_systemd {
+    ocf_mirrors::timer { "ftpsync-${title}":
+      exec_start   => "${project_path}/bin/ftpsync > /dev/null 2>&1",
+      environments => { 'BASEDIR' => $project_path },
+      hour         => $cron_hour,
+      minute       => $cron_minute,
+      require      => File["${project_path}/bin", "${project_path}/etc/ftpsync.conf"],
+    }
+  }
+  else {
```
jvperrin: nit: Put the previous `}`and this `else {` on the same line (like `} else {`), mostly just to follow the convention in this repo so far.

modules/ocf_mirrors/manifests/ftpsync.pp
```diff
@@ -11,6 +11,7 @@
     $mirror_name = 'mirrors.ocf.berkeley.edu',
     $mirror_path = "/opt/mirrors/ftp/${title}",
     $project_path = "/opt/mirrors/project/${title}",
+    $use_systemd = false,
```
jvperrin: I'd change the default here to be `true` since you're changing it elsewhere to be passed in as such. If we have problems, then this'll provide an easy way to change back, but having it set to a default of true seems like it would make more sense for any future additions.

modules/ocf_mirrors/templates/systemd/sync.service.erb
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Sync service for <%= @title %>
+
+[Service]
+User=<%= @exec_user %>
+Type=simple
+Environment=<%= @environments.map{ | env, val | "\""+env+"="+val+"\"" }.join(' ') %>
```
jvperrin: You can (and should, by best practive) have multiple `Environment` options in a systemd service file if you have multiple environment variables to set. I'd change this to be something like this instead, seems much simpler and then if you don't have any environment variables to set, there's no empty `Environment` option either:

```
<% @environments.each do |env, val| -%>
Environment="<%= env %>=<%= val %>"
<% end -%>
```
(I still do like Ruby)

kkuehler:
@jvperrin Since stdout of the units is not going to null it gets sucked into systemd journal. Easiest way to view logs is with, for example, `journalctl -eu archlinux`.

modules/ocf_mirrors/manifests/centos_altarch.pp
```diff
@@ -16,11 +16,10 @@
...
   }
 
-  cron {
+  ocf_mirrors::timer {
     'centos-altarch':
-      command => '/opt/mirrors/project/centos-altarch/sync-archive > /dev/null',
-      user    => 'mirrors',
-      hour    => '*/3',
-      minute  => '44';
+      exec_start => '/opt/mirrors/project/centos-altarch/sync-altarch',
```
jvperrin: This should be `sync-archive`, it's failing right now.

modules/ocf_mirrors/manifests/ftpsync.pp
```diff
@@ -42,11 +43,22 @@
     require => File[$project_path];
   }
 
-  cron { "ftpsync-${title}":
-    command => "BASEDIR=${project_path} ${project_path}/bin/ftpsync > /dev/null 2>&1",
-    user    => 'mirrors',
-    minute  => $cron_minute,
-    hour    => $cron_hour,
-    require => File["${project_path}/bin", "${project_path}/etc/ftpsync.conf"];
+  # TODO: Remove once systemd timers have been tested for reliability
```
jvperrin: I'd say you could remove this now if you want given this has been tested for a while.

kkuehler:
Phew! Spent some time squashing and reordering these commits into a much more readable sequence.

Gonna merge this once CI finishes.

PR #261, opened 2017-12-14
ocf: extrapackages: add nodeenv

kpengboy:


kpengboy:
Any reason it's not installed?

kpengboy:
Merging as no one appears to have objections

PR #262, opened 2017-12-15
Only try to resolve IPv6 addresses when IPv6 is active

kpengboy:
puppetlabs-firewall uses the ruby Resolv class to resolve hostnames. Resolv [only resolves IPv6 addresses when there is a public IPv6 address][resolv]. So, only try to apply IPv6 firewall rules with host names when the local host has an IPv6 address configured.

[resolv]: https://github.com/ruby/ruby/blob/a1a20cfaa2cf73ae8daaf2123d7c5b513427162e/lib/resolv.rb#L412

### DIFF
--- a/modules/ocf/manifests/firewall/post.pp
+++ b/modules/ocf/manifests/firewall/post.pp
@@ -4,7 +4,7 @@ class ocf::firewall::post {
   ['root', 'postfix'].each |$username| {
     ocf::firewall::firewall46 {
       "996 allow ${username} to send on SMTP port":
-        opts => {
+        opts   => {
           chain  => 'PUPPET-OUTPUT',
           proto  => 'tcp',
           dport  => 25,
@@ -17,7 +17,7 @@ class ocf::firewall::post {
 
   ocf::firewall::firewall46 {
     '997 forbid other users from sending on SMTP port':
-      opts => {
+      opts   => {
         chain  => 'PUPPET-OUTPUT',
         proto  => 'tcp',
         dport  => 25,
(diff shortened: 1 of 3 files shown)


modules/ocf/manifests/firewall/post.pp
```diff
@@ -28,10 +28,9 @@
 
 
   # Special devices we want to protect from most hosts
-  $devices = ['corruption-mgmt','hal-mgmt', 'jaws-mgmt', 'pagefault', 'pandemic-mgmt',
-              'papercut', 'radiation', 'riptide-mgmt']
-
-  $devices_ipv6 = ['radiation']
+  $devices_ipv4_only = ['corruption-mgmt','hal-mgmt', 'jaws-mgmt', 'pagefault',
+                        'pandemic-mgmt', 'papercut', 'riptide-mgmt']
+  $devices = ['radiation']
```
chriskuehl: this is data, not code -- would this fit better in hiera?

jvperrin: I'd agree, hiera does seem like a better fit for this, I guess it would go in `common.yaml` currently, but we might need to split that up into different files if it gets too large or confusing. Currently it seems fine since it is short, but splitting it up later seems good, since it has a mixture of Mesos stuff and other common host settings that don't exactly belong together.

kpengboy: Trouble is that rules in `ocf_kvm`, `ocf_desktop` and other places overriding these rules have hostnames hardcoded into the puppet file. Moving this data to hiera would break parallelism.

If we could have hiera rules applying to specific classes though, that would make a move to hiera much more desirable.

PR #264, opened 2017-12-19
Simplify ftpsync health checks

kkuehler:
Use one script that takes command line args for health checks as opposed to
creating a unique script for every mirror using a template. This way we take
advantage of the ocf_mirrors::monitoring defined type to handle url building
instead of doing it the exact same way in python.

In the spirit of the old style, each mirror has a health symlink pointing to
the actual script at /opt/mirrors/bin/health.

### DIFF
--- a/modules/ocf_mirrors/files/debian-healthcheck
+++ b/modules/ocf_mirrors/files/debian-healthcheck
@@ -10,22 +10,35 @@ mirrors.kernel.org).
 """
 import argparse
 import sys
+from collections import namedtuple
 from datetime import timedelta
 
 import dateutil.parser
 import requests
 
+Mirror = namedtuple('Mirror', ['url', 'updated_at'])
+
 
 def get_updated(mirror_url):
     """Find the time the host was last updated.
 
-    >>> get_updated(args.local_url, 'jessie')
+    >>> get_updated(args.local_url)
     datetime.datetime(2015, 12, 26, 9, 9, 42, tzinfo=tzutc())
     """
     req = requests.get(mirror_url)
     req.raise_for_status()
     updated_line, = [line for line in req.text.splitlines() if line.startswith('Date: ')]
     return dateutil.parser.parse(updated_line.split(': ', 1)[1])
+
+
+def get_mirrors(url_first, url_second):
+    """Given two mirror urls, returns two Mirror
+    namedtuples ordered (asc) by update date."""
+
+    mirror_first = Mirror(url_first, get_updated(url_first))
+    mirror_second = Mirror(url_second, get_updated(url_second))
+    return (min(mirror_first, mirror_second, key=lambda z: z.updated_at),
+            max(mirror_first, mirror_second, key=lambda z: z.updated_at))
 
 
 if __name__ == '__main__':
@@ -42,20 +55,17 @@ if __name__ == '__main__':
         print('Local and upstream urls cannot be equal... Exiting!')
         sys.exit(1)
 
+    local_mirror, upstream_mirror = get_mirrors(args.url_first, args.url_second)
+
     # instead of comparing against current time, compare against the master server;
     # this helps avoid flaky monitoring if the upstream archive isn't updated
-    mirror_first = get_updated(args.url_first), args.url_first
-    mirror_second = get_updated(args.url_second), args.url_second
-    local_mirror = min(mirror_first, mirror_second, key=lambda z: z[0])
-    upstream_mirror = max(mirror_first, mirror_second, key=lambda z: z[0])
-
-    delta = upstream_mirror[0] - local_mirror[0]
+    delta = upstream_mirror.updated_at - local_mirror.updated_at
     if delta > timedelta(hours=24):
         print('Warning: {} has not been updated in {}'.format(args.project, delta))
-        print('    Local mirror: {}'.format(local_mirror[0]))
-        print('        {}'.format(local_mirror[1]))
-        print('    Upstream mirror: {}'.format(upstream_mirror[0]))
-        print('        {}'.format(upstream_mirror[1]))
+        print('    Local mirror: {}'.format(local_mirror.updated_at))
+        print('        {}'.format(local_mirror.url))
+        print('    Upstream mirror: {}'.format(upstream_mirror.updated_at))
+        print('        {}'.format(upstream_mirror.url))
         sys.exit(1)
 
 # vim: ft=python
--- a/modules/ocf_mirrors/files/debian-healthcheck
+++ b/modules/ocf_mirrors/files/debian-healthcheck
@@ -4,7 +4,7 @@
 This detects most problems caused by both failed syncs to our direct upstream
 as well as cases where the upstream is out-of-date.
 
-Ideally, "args.upstream_url" should refer to the *actual* upstream (e.g. Debian's
+Ideally, the upstream url argument should refer to the *actual* upstream (e.g. Debian's
 authoritative archive) and not just our upstream mirror (e.g.
 mirrors.kernel.org).
 """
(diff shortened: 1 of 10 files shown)


modules/ocf_mirrors/files/health
```diff
@@ -8,62 +8,62 @@ Ideally, "UPSTREAM_MIRROR" should refer to the *actual* upstream (e.g. Debian's
...
+def get_updated(mirror_url):
     """Find the time the host was last updated.
 
     >>> get_updated(LOCAL_MIRROR, 'jessie')
     datetime.datetime(2015, 12, 26, 9, 9, 42, tzinfo=tzutc())
     """
-    req = requests.get(release_url(mirror))
+    req = requests.get(mirror_url)
+    if req.status_code != 200:
+        return -1
```
chriskuehl: do `req.raise_for_status()` instead probably

modules/ocf_mirrors/files/health
```diff
@@ -8,62 +8,62 @@ Ideally, "UPSTREAM_MIRROR" should refer to the *actual* upstream (e.g. Debian's
...
+
+    # order doesn't actually matter but we'll impose one for clarity's sake
+    parser.add_argument('local_url',
+                        type=str,
+                        help='URL of local mirror'
+                        )
+    parser.add_argument('upstream_url',
+                        type=str,
+                        help='URL of upstream mirror'
+                        )
```
chriskuehl: weird indentation style, inconsistent with the rest of the script

modules/ocf_mirrors/files/health
```diff
@@ -8,62 +8,62 @@ Ideally, "UPSTREAM_MIRROR" should refer to the *actual* upstream (e.g. Debian's
...
+
+    args = parser.parse_args()
+    LOCAL_MIRROR = args.local_url
+    UPSTREAM_MIRROR = args.upstream_url
```
chriskuehl: why make these "constants"?

modules/ocf_mirrors/manifests/init.pp
```diff
@@ -178,6 +178,13 @@
     special => 'daily',
   }
 
+  file { '/opt/mirrors/bin/health':
+    source  => 'puppet:///modules/ocf_mirrors/health',
```
chriskuehl: this is only for debian-style healthchecks, can you call this `debian-healthcheck` or something?

modules/ocf_mirrors/manifests/monitoring.pp
```diff
@@ -8,15 +8,15 @@
     $upstream_protocol = 'http',
     $ensure = 'present',
   ) {
-
+  $local_url = "http://mirrors.ocf.berkeley.edu${local_path}/dists/${dist_to_check}/Release"
+  $upstream_url = "${upstream_protocol}://${upstream_host}${upstream_path}/dists/${dist_to_check}/Release"
   if $ensure == 'present' {
     file { "${project_path}/health":
-      content => template("ocf_mirrors/monitoring/${type}-health.erb"),
```
chriskuehl: ? not using type param anymore?

seems like we need to preserve that. this was originally written such that we could add different types of healthchecks.

modules/ocf_mirrors/manifests/init.pp
kkuehler: In regards to this and the comment below about the type param, I think ```ftpsync-healthcheck``` is probably the best name.

modules/ocf_mirrors/manifests/monitoring.pp
kkuehler: Yeah. I'll add the type back to the file name. Maybe a healthchecks folder in scripts is a good addition for the future? Right now the only healthchecks running are ftpsync checks.

modules/ocf_mirrors/manifests/init.pp
chriskuehl: i think `ftpsync-healthcheck` is a bit of a misnomer; ftpsync is just how we download the files, you could use rsync directly or `wget -r` or ftp to download the mirror contents, too. the healthcheck would work on any debian-compatible repo.

kkuehler:
@chriskuehl changes made.

modules/ocf_mirrors/manifests/monitoring.pp
```diff
@@ -8,16 +8,18 @@
...
-      mode    => '0755';
-    }
-
+      ensure => link,
+      target => "/opt/mirrors/bin/${type}-healthcheck",
+      owner  => mirrors,
+      group  => mirrors,
+    }->
```
jvperrin: I'd add a space between the `}` and `~` to be more consistent with our other uses of this for chaining

modules/ocf_mirrors/manifests/monitoring.pp
```diff
@@ -8,16 +8,18 @@
...
     cron { "${title}-health":
-      command => "${project_path}/health",
-      user    => 'mirrors',
+      command => "${project_path}/health ${local_url} ${upstream_url}",
```
jvperrin: I'm not a fan of positional arguments for this since I think it's tough to remember the correct order (although this isn't being run by humans), and makes it less clear what is being passed in to the script. Maybe having required flags or something would be better? I'm a little split on this because I think adding the local and upstream mirrors as flags makes it look like they are optional when they should be required as inputs.

modules/ocf_mirrors/manifests/monitoring.pp
```diff
@@ -27,7 +29,8 @@
     }
 
     cron { "${title}-health":
-      ensure => absent;
+      ensure => absent,
+      user   => mirrors;
```
jvperrin: Nice, I'm not sure this ever worked before (lol), so thanks for changing that.

kkuehler: Yeah, it would try and remove the service from root's crontab.

modules/ocf_mirrors/manifests/monitoring.pp
kkuehler: I had the exact same feelings as you when changing the script. I was really close to making them required flags. If it makes you feel better, the order doesn't *actually* matter for this script. We can make the positional params more flexible and just check that the same url isn't given twice?

jvperrin: The order does currently matter, since it subtracts the times from each other, although if it just took the larger one and subtracted the smaller one that would work ok and then the order wouldn't matter here (and you'd need to change the output so that it doesn't assume the first is local and the second is upstream). I think that might be better, although I don't see the order ever changing in how we call the script or anything. In general, if the order doesn't matter, then I think positional arguments are fine, but otherwise I'd be more inclined to lean towards flags.

modules/ocf_mirrors/manifests/monitoring.pp
jvperrin: lol nice

modules/ocf_mirrors/manifests/monitoring.pp
kkuehler: How about using ```abs``` and then what I said above and we're both happy? For example, using data from our debian mirrors, ```local - upstream```.

```python
str(parse('2017-12-21 04:56:19+00:00') - parse('2017-12-21 07:34:54+00:00'))
'-1 day, 21:21:25'
str(abs(parse('2017-12-21 04:56:19+00:00') - parse('2017-12-21 07:34:54+00:00')))
'2:38:35'
```

kkuehler:
Okay, @jvperrin, the most recent commit should address your concerns about the positional arguments; we could also do [required named arguments](https://stackoverflow.com/a/24181138), but something about that feels ... hacky?

modules/ocf_mirrors/files/debian-healthcheck
```diff
@@ -0,0 +1,61 @@
...
+    # ensure we aren't comparing a mirror against itself
+    args = parser.parse_args()
+    if args.url_first == args.url_second:
+        print('Local and upstream urls cannot be equal... Exiting!')
+        sys.exit(1)
+
+    # instead of comparing against current time, compare against the master server;
+    # this helps avoid flaky monitoring if the upstream archive isn't updated
+    mirror_first = get_updated(args.url_first), args.url_first
+    mirror_second = get_updated(args.url_second), args.url_second
```
jvperrin: I think it might be better to make these into dicts instead of tuples, since I think doing something like `upstream_mirror['url']` is much clearer than doing `upstream_mirror[1]`, and same with `upstream_mirror['updated_at']` and `upstream_mirror[0]` for instance.

modules/ocf_mirrors/files/debian-healthcheck
```diff
@@ -0,0 +1,61 @@
...
+    local_mirror = min(mirror_first, mirror_second, key=lambda z: z[0])
+    upstream_mirror = max(mirror_first, mirror_second, key=lambda z: z[0])
```
jvperrin: This feels a bit hacky to me, but I think if you convert these to dicts instead of tuples, as I mention above, that would make this a lot clearer (instead of just using index 0 with less context)

kkuehler:
@jvperrin Okay, I made your suggestions.

modules/ocf_mirrors/files/debian-healthcheck
dkess: I'm personally a fan of namedtuples.

jvperrin: Yeah, they are pretty nice, probably better than dictionaries here, especially because of the dotted notation you can use with them.

kkuehler: Done! Thanks for the suggestion, @dkess. Makes the code a lot more clear.

PR #265, opened 2017-12-26
fish sudo bangbang

Baisang:
pretty jank but I only use `!!` for `sudo !!` in bash anyway so...

it's not hard to make an alias for `!!` that just does something with `$history[1]` anyway, but it won't work with `sudo` in that case

it's not in fish because they want you to press up and then home and then type `sudo`. and also because magic words can be dumb. but i like this magic word

if yall other fish users dont want it i can just put it in my personal configs

### DIFF
--- a/modules/ocf/files/shell/config.fish
+++ b/modules/ocf/files/shell/config.fish
@@ -32,6 +32,15 @@ alias rm "rm -I"
 alias leetfish "set -gx FISH swim"
 alias noobfish "set -gx FISH noob"
 
+# for sudo !!
+function sudo
+    if test "$argv" = !!
+        eval command sudo $history[1]
+    else
+        command sudo $argv
+    end
+end
+
 # Logout
 function on_exit --on-process %self
     if klist > /dev/null 2>&1


kkuehler:
Thanks for doing this! The thing is, I'm also not sure the globalrc file is the best place for this, especially given that [the creators of fish shell are ideologically opposed to history substitution](https://fishshell.com/docs/current/faq.html#faq-history).

Baisang:
@kkuehler yep, they are. but they aren't using our servers, and we should be free to customize how we like.

again, I don't have any issue putting this into my personal configs, but if fish users on staff all like this there's no issue putting it in imo.

jvperrin:
This is why I wanted more reviewers than myself lol. I think there's enough people saying no on this that it should just be put in personal config. Maybe we could have some kind of repo that is packaged for distributing staff users' configs across all systems though so that we can have more customizable personal configs/dotfiles OCF-wide? I know that's a pretty common distribution pattern for these kinds of things anyway.

abizer:
We already kinda ship modified configs, e.g. vim keybindings in zshrc[0], so the cat is out of the bag as far as precedent goes. 

For what it's worth I think this sort of thing is fine to ship as it doesn't change any defaults, only adds things that you don't have to make any changes to not use.

[0]: https://github.com/ocf/puppet/blob/master/modules/ocf/files/shell/zshrc

Baisang:
I'll just close this since there's opposition against it. Thanks for weighing in

chriskuehl:
Traditionally we have shipped pretty minimal configs (as is polite to do on shared systems). There is lots of precedent -- at the OCF and elsewhere -- for not putting opinionated stuff in global configs (heck, we had to revert the new vimrc that Debian ships because it's so awful; Debian has thankfully reverted it themselves in buster).

If you look through RT and IRC logs you'll also find a few recent cases (past few years) of people complaining about stuff in our bashrc.

zsh is a bit different because it ships with unusable defaults (unlike bash) -- it pops up a config menu every time you log in to a new host, which for the OCF with 100+ hosts is nearly unbearable. I don't think we should use that as justification for adding even more global stuff (I agree we should delete the vi stuff though!).

I think jvperrin has the right idea above about the best approach for these problems.

PR #267, opened 2018-01-14
Add Timestamp Healthchecks

kkuehler:
Made the healthcheck script and ocf_mirrors::monitoring more generalized.

Healthchecks for apache, arch, centos, gnu, parrot, qt, tails.

### DIFF
--- a/modules/ocf_mirrors/files/healthcheck
+++ b/modules/ocf_mirrors/files/healthcheck
@@ -25,20 +25,24 @@ def update_func(healthcheck):
         return get_updated_debian
     elif healthcheck == 'ts':
         return get_updated_ts
+    elif healthcheck == 'datetime':
+        return get_updated_datetime
     else:
-        return get_updated_debian
+        raise ValueError('Unsupported type: {}'.format(healthcheck))
+
+
+def get_updated_datetime(mirror_url):
+    """Find the time the host was last synced"""
+    req = requests.get(mirror_url)
+    req.raise_for_status()
+    return dateutil.parser.parse(req.text.splitlines()[0])
 
 
 def get_updated_ts(mirror_url):
     """Find the time the host was last synced"""
-
     req = requests.get(mirror_url)
     req.raise_for_status()
-    try:
-        return datetime.utcfromtimestamp(int(req.text.split()[0]))
-    except (TypeError, ValueError):
-        # UTC Timestamp ... parrot
-        return dateutil.parser.parse(req.text.splitlines()[0])
+    return datetime.utcfromtimestamp(int(req.text.split()[0]))
 
 
 def get_updated_debian(mirror_url):
@@ -55,7 +59,6 @@ def get_updated_debian(mirror_url):
 def get_mirrors(url_first, url_second, get_updated):
     """Given two mirror urls, return two Mirror
     namedtuples ordered (asc) by sync date."""
-
     mirror_first = Mirror(url_first, get_updated(url_first))
     mirror_second = Mirror(url_second, get_updated(url_second))
     return (min(mirror_first, mirror_second, key=lambda z: z.updated_at),
--- a/modules/ocf_mirrors/files/healthcheck
+++ b/modules/ocf_mirrors/files/healthcheck
@@ -23,7 +23,7 @@ Mirror = namedtuple('Mirror', ['url', 'updated_at'])
 def update_func(healthcheck):
     if healthcheck == 'debian':
         return get_updated_debian
-    elif healthcheck == 'ts':
+    elif healthcheck == 'unix_timestamp':
         return get_updated_ts
     elif healthcheck == 'datetime':
         return get_updated_datetime
--- a/modules/ocf_mirrors/files/healthcheck
+++ b/modules/ocf_mirrors/files/healthcheck
@@ -24,7 +24,7 @@ def update_func(healthcheck):
     if healthcheck == 'debian':
         return get_updated_debian
     elif healthcheck == 'unix_timestamp':
-        return get_updated_ts
+        return get_updated_unix_timestamp
     elif healthcheck == 'datetime':
         return get_updated_datetime
     else:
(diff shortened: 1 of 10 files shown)


modules/ocf_mirrors/manifests/tails.pp
```diff
@@ -9,6 +9,12 @@
       recurse => true;
   }
 
+  ocf_mirrors::monitoring { 'tails':
+    type          => 'ts',
+    upstream_host => 'mirrors.kernel.org',
+    ts_path       => 'project/trace',
```
jvperrin: Instead of relying on another mirror here (albeit a very reliable one), I'd use the upstream at https://archive.torproject.org/amnesia.boum.org/tails/project/trace

(It was pretty hard to find the upstream here, but given the tails mirroring instructions say to mirror from rsync.torproject.org which goes to the same DNS entry as archive.torproject.org, this should be the true upstream)

modules/ocf_mirrors/manifests/monitoring.pp
```diff
@@ -6,19 +6,27 @@
...
-  $local_url = "http://mirrors.ocf.berkeley.edu${local_path}/dists/${dist_to_check}/Release"
-  $upstream_url = "${upstream_protocol}://${upstream_host}${upstream_path}/dists/${dist_to_check}/Release"
+
+  if $type == 'debian' {
+    $local_url = "http://mirrors.ocf.berkeley.edu${local_path}/dists/${dist_to_check}/Release"
+    $upstream_url = "${upstream_protocol}://${upstream_host}${upstream_path}/dists/${dist_to_check}/Release"
+  }
+  elsif $type == 'ts' {
+    $local_url ="http://mirrors.ocf.berkeley.edu${local_path}/${ts_path}"
+    $upstream_url = "${upstream_protocol}://${upstream_host}${upstream_path}/${ts_path}"
```
jvperrin: Since the first part of our mirror and the upstream mirror is the same, maybe put those in variables above the condition? I think that'll help make it a but clearer, especially if we need to change something for both types of health checks at once (or, more likely, add a third type of health check)

modules/ocf_mirrors/manifests/archlinux.pp
```diff
@@ -9,6 +9,12 @@
       recurse => true;
   }
 
+  ocf_mirrors::monitoring { 'archlinux':
+    type          => 'ts',
+    upstream_host => 'mirrors.kernel.org',
```
jvperrin: I wish there was a more official upstream here, but I can't find one (I think they whitelist access to the direct mirror, and might only allow rsync), so I think this is ok.

modules/ocf_mirrors/files/healthcheck
```diff
@@ -19,9 +20,29 @@ import requests
...
 
-def get_updated(mirror_url):
-    """Find the time the host was last updated.
+def update_func(healthcheck):
+    if healthcheck == 'debian':
+        return get_updated_debian
+    elif healthcheck == 'ts':
+        return get_updated_ts
+    else:
+        return get_updated_debian
```
jvperrin: This seems a bit redundant, maybe raise an error if the type is not one of the accepted ones? That seems like it would be a problem if it does actually happen.

modules/ocf_mirrors/files/healthcheck
```diff
@@ -19,9 +20,29 @@ import requests
...
+
+def get_updated_ts(mirror_url):
+    """Find the time the host was last synced"""
+
+    req = requests.get(mirror_url)
+    req.raise_for_status()
+    try:
+        return datetime.utcfromtimestamp(int(req.text.split()[0]))
+    except (TypeError, ValueError):
+        # UTC Timestamp ... parrot
```
jvperrin: Parrot?

jvperrin: Oh, I see, it just tries to parse it however it can, is this what was needed for parrot's health check? If so I'd make a new check for it instead of doing this, because that seems safer, then we'll know if something is going wrong more easily. Otherwise I could see a mirror returning something that maybe parses into a date way in the past or something, and then we don't know that it isn't syncing correctly any more.

modules/ocf_mirrors/manifests/tails.pp
kkuehler: omg thank you for finding this. i was looking forever yesterday and just gave up

kkuehler:
@jvperrin Changes made.

modules/ocf_mirrors/files/healthcheck
```diff
@@ -19,9 +20,33 @@ import requests
 Mirror = namedtuple('Mirror', ['url', 'updated_at'])
 
 
-def get_updated(mirror_url):
-    """Find the time the host was last updated.
+def update_func(healthcheck):
+    if healthcheck == 'debian':
+        return get_updated_debian
+    elif healthcheck == 'ts':
+        return get_updated_ts
```
chriskuehl: can you use a better name for ts? even spelling out "unix_timestamp" would be preferable if that's what it means

kkuehler:
@chriskuehl Done.

modules/ocf_mirrors/files/healthcheck
```diff
@@ -19,9 +20,33 @@ import requests
...
+def get_updated_ts(mirror_url):
+    """Find the time the host was last synced"""
+    req = requests.get(mirror_url)
+    req.raise_for_status()
+    return datetime.utcfromtimestamp(int(req.text.split()[0]))
+
+
+def get_updated_debian(mirror_url):
+    """Find the time the host was last updated.
     >>> get_updated(args.local_url)
```
jvperrin: This isn't correct now, the function has changed names.

modules/ocf_mirrors/files/healthcheck
```diff
@@ -19,9 +20,33 @@ import requests
...
+
+def get_updated_datetime(mirror_url):
+    """Find the time the host was last synced"""
+    req = requests.get(mirror_url)
+    req.raise_for_status()
+    return dateutil.parser.parse(req.text.splitlines()[0])
+
 
+def get_updated_ts(mirror_url):
+    """Find the time the host was last synced"""
```
jvperrin: Same as the comment above, I think this description should be different from that one and have more useful information.

modules/ocf_mirrors/files/healthcheck
```diff
@@ -19,9 +20,33 @@ import requests
...
+    elif healthcheck == 'unix_timestamp':
+        return get_updated_ts
+    elif healthcheck == 'datetime':
+        return get_updated_datetime
+    else:
+        raise ValueError('Unsupported type: {}'.format(healthcheck))
+
+
+def get_updated_datetime(mirror_url):
+    """Find the time the host was last synced"""
```
jvperrin: I'd make this function description more useful, like show what the format looks like generally on the mirror, maybe what this function then returns (although that's pretty clear anyway from how it works).

modules/ocf_mirrors/manifests/monitoring.pp
```diff
@@ -6,19 +6,28 @@
...
+  $local_base = "http://mirrors.ocf.berkeley.edu${local_path}"
+  $upstream_base = "${upstream_protocol}://${upstream_host}${upstream_path}"
+
+  if $type == 'debian' {
+    $local_url = "${local_base}/dists/${dist_to_check}/Release"
+    $upstream_url = "${upstream_base}/dists/${dist_to_check}/Release"
+  } elsif $type == 'unix_timestamp' or $type == 'datetime' {
+    $local_url ="${local_base}/${ts_path}"
+    $upstream_url = "${upstream_base}/${ts_path}"
+  }
```
jvperrin: Nit: add an extra newline here to separate it from the section below a bit.

kkuehler:
@jvperrin Comments + newline added, @jvperrin

PR #269, opened 2018-01-20
Add qrencode to packages

dkess:
`qrencode` could be useful for the desktop mapping project, and it's also a cool package to have.

### DIFF
--- a/modules/ocf/manifests/packages.pp
+++ b/modules/ocf/manifests/packages.pp
@@ -110,6 +110,7 @@ class ocf::packages {
     'python3-pip',
     'python3-requests',
     'python3-tabulate',
+    'qrencode',
     'quota',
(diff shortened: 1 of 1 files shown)


modules/ocf/manifests/packages.pp
```diff
@@ -110,6 +110,7 @@
     'python3-pip',
     'python3-requests',
     'python3-tabulate',
+    'qrencode',
```
chriskuehl: Do you really need this on *all* hosts at the OCF? Even hosts like LDAP, mysql, etc?

This probably belongs either in ocf_desktop packages or extrapackages (depending on if it's only useful with a GUI or can be useful on a headless server).

jvperrin:
Closed by #270.

PR #272, opened 2018-01-21
munin-mail-alert: compensate for adding cc to ocflib send_mail

abizer:
since https://github.com/ocf/ocflib/pull/114 was merged, munin alerts have been sent with from: help@ocf.berkeley.edu because that's the default in ocflib/misc/mail.py. This should fix that.

### DIFF
--- a/modules/ocf_stats/files/munin/mail-munin-alert
+++ b/modules/ocf_stats/files/munin/mail-munin-alert
@@ -78,7 +78,7 @@ def send_alert(host, plugin, mail_to=MAIL_TO):
         ldap_entry=ldap_entry,
         dns_names=dns_names,
     )
-    send_mail(mail_to, subject, body, MAIL_FROM)
+    send_mail(to=mail_to, subject=subject, body=body, cc=None, sender=MAIL_FROM)
 
 
 def main(argv):


chriskuehl:
We should consider using keyword-only arguments (a new py3 feature) to prevent this in the future.

E.g. if the function had been written like this:

```python
def send_mail(to, subject, body, *, sender=MAIL_FROM):
```

...then this bug wouldn't have been possible, since `sender` couldn't have been filled in via positional arg.

modules/ocf_stats/files/munin/mail-munin-alert
```diff
@@ -78,7 +78,7 @@ def send_alert(host, plugin, mail_to=MAIL_TO):
         ldap_entry=ldap_entry,
         dns_names=dns_names,
     )
-    send_mail(mail_to, subject, body, MAIL_FROM)
+    send_mail(to=mail_to, subject=subject, body=body, cc=None, sender=MAIL_FROM)
```
chriskuehl: no need to specify `cc=None`

kpengboy: I would suggest writing this with positional arguments for required params and keyword arguments for optional params, like:

    send_mail(mail_to, subject, body, sender=MAIL_FROM)

When ocf/ocflib#114 was merged, I slightly feared that this would happen. I should have raised my concerns then.

abizer: I know but I prefer to list all the arguments in cases like these, makes it slightly easier when reading the code again in the future after a long time.

PR #273, opened 2018-01-25
Add fallback for PuppetDB failure.

redplanks:
During the recent outage, we were unable to puppet new hosts since PuppetDB had failed. This caused Puppet to completely fail, since it depends on PuppetDB to collect all SSH keys for known_hosts. This fix adds a custom fact script to check if PuppetDB is running, and only collects/exports SSH keys if it is up.

In the future, if more features utilizing exported resources are added, they'll have to be wrapped in an if block with this variable.

### DIFF
--- a/modules/ocf/facts.d/puppetdb-running
+++ b/modules/ocf/facts.d/puppetdb-running
@@ -1,8 +1,13 @@
 #!/bin/bash
-set -euo pipefail
+set -uo pipefail
+fqdn=$(hostname -f)
 
-curl http://puppetdb:8081/pdb/query/v4/nodes
-if [[ "$?" -eq 0 ]]; then
+# Due to differences between the puppet and puppet-agent packages, the SSL directory may be in /etc/puppetlabs/puppet or /var/lib/puppet
+puppet_ssl_path=$(puppet config print ssldir)
+
+curl 'https://puppetdb:8081/pdb/query/v4/nodes' -s --tlsv1 --cacert $puppet_ssl_path/certs/ca.pem --cert $puppet_ssl_path/certs/$fqdn.pem --key $puppet_ssl_path/private_keys/$fqdn.pem > /dev/null
+
+if [[ $? -eq 0 ]]; then
     echo 'puppetdb_running=true'
 else
     echo 'puppetdb_running=false'
--- a/modules/ocf/facts.d/puppetdb-running
+++ b/modules/ocf/facts.d/puppetdb-running
@@ -1,13 +1,16 @@
 #!/bin/bash
-set -uo pipefail
-fqdn=$(hostname -f)
+set -euo pipefail
 
-# Due to differences between the puppet and puppet-agent packages, the SSL directory may be in /etc/puppetlabs/puppet or /var/lib/puppet
+fqdn=$(hostname -f)
+# Due to differences between the puppet and puppet-agent packages,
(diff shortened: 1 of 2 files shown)


modules/ocf/facts.d/puppetdb-running
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -uo pipefail
+fqdn=$(hostname -f)
+
+# Due to differences between the puppet and puppet-agent packages, the SSL directory may be in /etc/puppetlabs/puppet or /var/lib/puppet
+puppet_ssl_path=$(puppet config print ssldir)
+
+curl 'https://puppetdb:8081/pdb/query/v4/nodes' -s --tlsv1 --cacert $puppet_ssl_path/certs/ca.pem --cert $puppet_ssl_path/certs/$fqdn.pem --key $puppet_ssl_path/private_keys/$fqdn.pem > /dev/null
```
jvperrin: I think replacing this with something like `https://puppetdb:8081/pdb/meta/v1/version` or `https://puppetdb:8081/pdb/meta/v1/server-time` would be better, since then it doesn't have to return the details for all the nodes and instead only checks that the server is up. Also if you can wrap this to multiple lines it would probably be good, since it's quite long as written like this.

modules/ocf/facts.d/puppetdb-running
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -uo pipefail
+fqdn=$(hostname -f)
+
+# Due to differences between the puppet and puppet-agent packages, the SSL directory may be in /etc/puppetlabs/puppet or /var/lib/puppet
```
jvperrin: nit: wrap lines

modules/ocf/manifests/auth.pp
```diff
@@ -143,14 +143,16 @@
     $::ipaddress6,
   ), '')
 
-  # Export SSH keys from every host, and use them to populate the global list
-  # in /etc/ssh/ssh_known_hosts
-  @@sshkey { $::hostname:
-    host_aliases => $ssh_aliases,
-    key          => $::sshecdsakey,
-    type         => ecdsa-sha2-nistp256,
+  # Export SSH keys from every host if PuppetDB is running, and use them to populate the global list
```
jvperrin: nit: wrap lines

modules/ocf/facts.d/puppetdb-running
kpengboy: Also, the variables in this line ought to be quoted. Even though they aren't expected to contain a space, it's a good habit to get into.

modules/ocf/facts.d/puppetdb-running
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -uo pipefail
```
kpengboy: We can also use `-e` if you're willing to squeeze the curl line onto the if, like

    if curl blah blah blah \
            blah blah blah \
            blah blah blah; then
        echo 'puppetdb running=true'

Just saying...

redplanks:
@jvperrin It turns out that depending on if the puppet or puppet-agent package is installed, the Puppet SSL certs can be in different places. So the script was failing since it was looking in the wrong directory for the private key.

jvperrin:
Ah, that makes sense, cool.

PR #275, opened 2018-01-27
Reduce tearing with compositor

kkuehler:
Tearing is not completely gone, but this is much less awful.


modules/ocf_desktop/files/xsession/Xsession
```diff
@@ -118,3 +118,7 @@ sleep 8 && /usr/local/bin/fix-audio &
 
 # Source custom rc shared across desktops
 [ -f ~/remote/.desktoprc ] && . ~/remote/.desktoprc
+
+# Enable compositor
+xfconf-query -c xfwm4 -p /general/use_compositing -s false
```
chriskuehl: don't we already set it to false?
https://github.com/ocf/puppet/blob/d0b998a059bbdcd07b0c1433f53cff87438f97b7/modules/ocf_desktop/files/xsession/xfce4/xfconf/xfce-perchannel-xml/xfwm4.xml#L67

is that config file not working?

FWIW i think we should set it in a config file, not control it in a script at login time

chriskuehl: side note: if we really *aren't* using a compositor now, have we tried just turning on whatever the xfce default one and seeing if that's sufficient? less maintenance for us.

modules/ocf_desktop/files/xsession/Xsession
```diff
@@ -118,3 +118,7 @@ sleep 8 && /usr/local/bin/fix-audio &
...
+/usr/bin/compton --config $HOME/.config/compton/compton.conf --daemon
```
chriskuehl: not a huge fan of this pattern, it's pretty likely that this will break at some point in the future (typo in the config file, changes to compton when we upgrade desktops to a new version of debian, etc.) and nobody will notice since the output of this script just goes to `/dev/null` (and lack of compositor will take a while for anybody to notice)

modules/ocf_desktop/files/xsession/Xsession
kkuehler: No, we aren't using a compositor now; the config file works just fine. I just put that line in to make it clear that two compositors aren't running at the same time, so it's not really necessary.

What really pissed me off was when I'd drag a terminal somewhat fast and there would be a delay with 10 of them following it. Essentially, it looks like the few moments before your Windows 95 system crashes. Both the xfce compositor and compton fix this issue. But, for example, when running a [tearing test](https://www.youtube.com/watch?v=9hIRq5HTh5s), the xfce compositor still has noticeable tearing whereas compton does not. @tliu22 can vouch for this too.

I'd like if someone more well-versed in the graphics realm, such as @harrisonkuo, shared their opinion on this too.

modules/ocf_desktop/files/xsession/Xsession
kkuehler: I suppose that's fair enough. We can pass all the necessary parameters like this instead (added some new stuff that seems to increase performance). 
```bash
compton  --daemon --config /dev/null --backend glx  \
--vsync opengl-swc --glx-no-stencil --paint-on-overlay \
--glx-swap-method 3 --glx-no-rebind-pixmap  \
--xrender-sync  --xrender-sync-fence  --unredir-if-possible \
--detect-client-opacity
```

chriskuehl: That doesn't solve the problem. Imagine e.g. that a new version stops accepting a command line argument. Is there any way we can design this to fail loudly?

kkuehler: Passing an invalid arg makes `$? == 1`, and it's 0 otherwise. We could do a `notify-send` or write to ~/.xsession-errors, though no one checks that so it's not particularly loud.

jvperrin:
@harrisonkuo mentioned this to me yesterday, but if you run something like `glxgears` and then drag around a window at the same time on the same monitor, the window freezes up several times while dragging it (unless you drag the window _very_ slowly). I'm guessing this is still better than current tearing behavior, but it does seem like the compositor can cause some freezing issues when doing anything graphically intensive.

kkuehler:
@jvperrin it's also causing some monitors to glitch out at random so something is wrong.

kpengboy:
Closing this PR. If someone can get a compositor to work, feel free to make a new PR.

PR #276, opened 2018-01-27
backup-mysql: make a file per database

abizer:
This implements one of the items on the post-1/17/18-outage to-do list, which is to make more granular backups of mysql. This makes one file per mysql database. At the time of writing, that means 1845 files, about 2.8GB total.

[Here](https://i.fluffy.cc/N8GZgF3kNF9rzC1tqtc7ztM9jqqwLq8H.html) is an rough alternative that only separates the dump into ocf/non-ocf databases. 

Some statistics:
1. The script in this PR:

        root@hal:/opt/backups/scratch/mysqltest# time ../backup-mysql

        real    2m59.271s
        user    22m51.684s
        sys     4m20.868s

2. The alternative script (ocf/non-ocf)

        root@hal:/opt/backups/scratch/mysqltest# time ../backup-mysql-alt

        real    11m33.717s
        user    22m14.784s
        sys     3m37.496s

3. The existing script:

        root@hal:/opt/backups/scratch/mysqltest# time /opt/share/backups/backup-mysql

        real    11m19.520s
        user    22m8.972s
        sys     3m41.952s

I also changed the user in my.cnf to a read-only root user named `root_ro`, with the password in the usual location. 

Feedback is appreciated.

### DIFF
--- a/modules/ocf_backups/files/backup-mysql
+++ b/modules/ocf_backups/files/backup-mysql
@@ -1,10 +1,20 @@
 #!/bin/bash -eu
-# Creates a MySQL backup in the current directory. Intended to be called by
-# rsnapshot.
-set -o pipefail
-mysqldump \
-    --defaults-file=/opt/share/backups/my.cnf \
-    --events \
-    --triggers \
-    --all-databases --single-transaction | \
-    pigz > "mysql-$(date +%F).sql.gz"
+# Creates a MySQL backup in the current directory, with one file per database.
+# Intended to be called by rsnapshot.
+
+# credit:
+# https://mensfeld.pl/2013/04/backup-mysql-dump-all-your-mysql-databases-in-separate-files/
+
+set -euo pipefail
+
+databases=$(mysql --defaults-file=/opt/share/backups/my.cnf -e "SHOW databases;" | \
+		   grep -Ev "(Database|information_schema|performance_schema)")
+
+parallel -i \
+    sh -c 'mysqldump \
+             --defaults-file=/opt/share/backups/my.cnf \
+             --events \
+             --triggers \
+             --routines \
+             --single-transaction \
+             --databases {} | pigz > "mysql-{}-$(date +%F).sql.gz"' -- $databases


chriskuehl:
> I also changed the user in my.cnf to a read-only root user named root_ro, with the password in the usual location.

This is definitely an improvement over using `root`, but we can improve it further by using making an account exclusively for this role, e.g. `ocfbackups`, rather than an account based on the privileges it currently has.

modules/ocf_backups/files/backup-mysql
```diff
@@ -1,10 +1,20 @@
...
+# Creates a MySQL backup in the current directory, with one file per database.
+# Intended to be called by rsnapshot.
+
+# credit:
+# https://mensfeld.pl/2013/04/backup-mysql-dump-all-your-mysql-databases-in-separate-files/
+
+set -euo pipefail
+
+databases=$(mysql --defaults-file=/opt/share/backups/my.cnf -e "SHOW databases;" | \
+		   grep -Ev "(Database|information_schema|performance_schema)")
```
jvperrin: I mentioned yesterday that I thought anchoring this regex would be safer in case a username or something happens to contain `Database` or something, but given that the output here has a table format and everything, I think just leaving it like this would be the easiest and should be safe enough.

abizer:
`ocfbackups` already exists as the user we use for [Box backups](https://github.com/ocf/puppet/blob/master/modules/ocf_backups/files/upload-to-box#L140) - I'm not sure how good an idea it is to repurpose that user and grant it `SELECT ON *.*` for the purpose of this script. I'd trust your judgement on this more than mine, though. Maybe `ocf_backup` or something like that?

chriskuehl:
@abizer ah, I didn't realize that user already existed. Yeah, creating another may be better than sharing the one, though I would argue that two backup-related things sharing `ocfbackups` is still better than everything that needs read-only access sharing a `root-ro` account. Ideally we'd have an account for each thing that needs one.

jvperrin:
Feel free to rename `ocfbackups` if you want (and change it in the Box upload script), it could have a user named something else, whereas I think the user used for this more deserves the name. In general, the only thing the `ocfbackups` user does in the Box script is update the API key in a single table, so it doesn't do much. I'd be pretty happy combining the functionality of the two, but it's up to you, making a new user would be fine too.

abizer:
Combining the two shouldn't be that big a deal, I guess. I'll make the requisite changes.

abizer:
This is done, with `ocfbackups` handling both dumping from mysql and Box.com backups.

abizer:
I let this run on hal last night, it works, and the `zless mysql-abizer-*.sql.gz` also looks good, so I'll merge this.

PR #277, opened 2018-01-28
Add bzh key.

redplanks:


redplanks:
Ah, let me create a new key associated with my berkeley email then.

jvperrin:
Wait, you don't need to do that, you just put your Berkeley email in there so the backups get shared with your account, they are still encrypted with the key you put here. The first link I put you put the email associated with this key, and the second one you just put your account email, it's not related to the key at all.

redplanks:
Ah, okay. Sorry for the confusion on my part.

jvperrin:
That's fine, it's just confusing in that there are multiple lists of emails, but that they are for different purposes.

PR #278, opened 2018-01-30
Added more healthchecks

kkuehler:
centos-altarch, kde & kde-appdata, manjaro, and trisquel.

the kde healthcheck is a bit inefficient (massive ls-lR file) and can be done in one pass, instead of two, if necessary.


modules/ocf_mirrors/files/healthcheck
```diff
@@ -53,6 +57,19 @@ def get_updated_unix_timestamp(mirror_url):
...
 
+def get_updated_recursive_ls(mirror_url):
+    """Find the time the host was last synced.
+    Should contain the output of ls -lR
+    >>> get_updated_unix_timestamp(mirror_url)
+    datetime.datetime(2018, 1, 24, 10, 3)
+    """
+    req = requests.get(mirror_url)
+    req.raise_for_status()
+    dates = [' '.join(line.split()[5:7]) for line in req.text.splitlines()
```
jvperrin: I think a comment here with an example line would help a lot to show why you are getting these specific indices.

modules/ocf_mirrors/files/healthcheck
```diff
@@ -53,6 +57,19 @@ def get_updated_unix_timestamp(mirror_url):
     return datetime.utcfromtimestamp(int(req.text.split()[0]))
 
 
+def get_updated_recursive_ls(mirror_url):
+    """Find the time the host was last synced.
+    Should contain the output of ls -lR
```
jvperrin: Maybe mention that this selects the most recent date from the output of `ls -lR`?

modules/ocf_mirrors/manifests/centos.pp
```diff
@@ -8,10 +8,19 @@
...
-  ocf_mirrors::monitoring { 'centos':
+  ocf_mirrors::monitoring {
+    default:
       type          => 'unix_timestamp',
       upstream_host => 'mirror.centos.org',
-      ts_path       => 'TIME',
+      ts_path       => 'TIME';
+
+    'centos':
+    ;
```
jvperrin: I feel like this is confusing as it is, I'd just prefer a few things to be duplicated here instead of how it is now. Then at least it would be easier to read/maintain.

modules/ocf_mirrors/manifests/kde.pp
```diff
@@ -9,6 +9,22 @@
...
+      upstream_host => 'download.kde.org',
+      upstream_path => '',
+      ts_path       => 'ls-lR';
+
+    'kde-applicationdata':
+      type          => 'unix_timestamp',
+      upstream_host => 'files.kde.org',
+      upstream_path => '',
+      ts_path       => 'last-updated',
+      filename     => 'health-appdata',
```
jvperrin: nit: arrow alignment (and on the next line)

modules/ocf_mirrors/manifests/monitoring.pp
```diff
@@ -15,26 +16,26 @@
   if $type == 'debian' {
     $local_url = "${local_base}/dists/${dist_to_check}/Release"
     $upstream_url = "${upstream_base}/dists/${dist_to_check}/Release"
-  } elsif $type == 'unix_timestamp' or $type == 'datetime' {
+  } elsif $type == 'unix_timestamp' or $type == 'datetime' or $type == 'manjaro' or $type == 'recursive_ls' {
```
jvperrin: Maybe just put `else` here? Seems reasonable anyway and a bit cleaner, and if a strange type is given, the healthcheck script will catch it anyway.

modules/ocf_mirrors/manifests/trisquel.pp
```diff
@@ -11,4 +11,11 @@
       rsync_extra              => '--block-size=8192',
       cron_minute              => '55';
   }
+
+  ocf_mirrors::monitoring { 'trisquel':
+    type          => 'debian',
+    dist_to_check => 'flidas-security',
```
jvperrin: Why the security dist? Just curious, I'd imagine the main dist (`flidas`) or the updates dist (`flidas-updates`) would get updated more often, but I haven't done much research here.

modules/ocf_mirrors/manifests/centos.pp
```diff
@@ -8,10 +8,19 @@
...
+
+    'centos-altarch':
+      filename      => 'health-altarch',
+      upstream_path => '/altarch',
+      project_path  => '/opt/mirrors/project/centos';
```
jvperrin: Why not just put this in its own project? It has its own sync script, monitoring, directory on mirrors.o.b.e. Seems to me like it should just be separated out.

kkuehler: Cool. That'll actuallt make the puppet code a lot simpler too.

jvperrin: Sounds good to me, sure

modules/ocf_mirrors/manifests/trisquel.pp
kpengboy: This is troublesome because `flidas` won't be updated after 2021, after which this healthcheck will silently always succeed. On the other hand, files like `bandwidth-day.png` seem to get updated every day...

modules/ocf_mirrors/files/healthcheck
```diff
@@ -65,6 +82,18 @@ def get_updated_debian(mirror_url):
     return dateutil.parser.parse(updated_line.split(': ', 1)[1])
 
 
+def get_updated_manjaro(mirror_url):
+    """Find the time the host was last synced.
+    The last line on the page should be of the form "date=2018-01-25T21:52:18Z"
```
kpengboy: Is this explicitly specified? If not I would go for regex matching lest they decide to append more lines later.

modules/ocf_mirrors/manifests/centos.pp
kpengboy: If you're gonna do that, you might as well consider throwing in ubuntu-releases and trisquel-images into the mix.

modules/ocf_mirrors/manifests/trisquel.pp
kkuehler: I mean, yeah, that date should be in the headers, but I'd really prefer to wait until 2021 to give trisquel a unique healthcheck...

modules/ocf_mirrors/manifests/centos.pp
kkuehler: @kpengboy Sure. Looks like (this)[https://mirrors.ocf.berkeley.edu/trisquel-images/project/trace/mirrors.ocf.berkeley.edu] should work for trisquel. Can't seem to find any file to healthcheck in ubuntu releases. If you find one I'd be happy to add it.

modules/ocf_mirrors/manifests/trisquel.pp
kpengboy: However, by 2021 everyone will have forgotten that this needs to be updated.

kkuehler:
@jvperrin Finally got around to making these changes.

modules/ocf_mirrors/files/project/kde-appdata/sync-archive
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -eu
+/usr/local/bin/rsync-no-vanished -az --delete --safe-links \
+	rsync.kde.org::applicationdata /opt/mirrors/ftp/kde-applicationdata
```
jvperrin: Make sure to also remove this from `modules/ocf_mirrors/files/project/kde/sync-archive`

modules/ocf_mirrors/manifests/centos_altarch.pp
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  cron {
+    'centos-altarch':
+      command => '/opt/mirrors/project/centos-altarch/sync-altarch > /dev/null',
+      user    => 'mirrors',
+      hour    => '*/3',
+      minute  => '44';
+  }
+}
```
jvperrin: I think ideally we'd make a new class to simplify adding new mirrors, since there's a lot of duplicated content here (e.g. file owners, paths, etc.) that we could extract and put into a common class that handles adding a whole mirror with monitoring. Anyway, that's probably content for a future PR, not this one.

modules/ocf_mirrors/manifests/monitoring.pp
```diff
@@ -7,6 +7,7 @@
     $upstream_path = "/${title}",
     $upstream_protocol = 'http',
     $ts_path = undef,
+    $filename = 'health',
```
jvperrin: I think this could be removed again, since none of the health checks use this any more.

modules/ocf_mirrors/manifests/trisquel.pp
jvperrin: Honestly I think this is fine as it is, we'll probably have changed our mirroring by then anyway, and if we haven't, then it doesn't hurt to have this over no monitoring at all. I do agree we should try and make this more dynamic somehow, but not sure how to do it.

kkuehler: I mean, isn't trisquel 8 (`fildas`) still in alpha? We can update this when newer versions are released. I don't think it's a big deal, nor will it silently fail as new releases are publicly announced, and we will have to change one line in puppet.

modules/ocf_mirrors/manifests/centos_altarch.pp
```diff
@@ -0,0 +1,26 @@
...
+    'centos-altarch':
+      type          => 'unix_timestamp',
+      upstream_host => 'mirror.centos.org',
+      upstream_path => '/altarch',
+      ts_path       => 'TIME';
+  }
+
+  cron {
+    'centos-altarch':
+      command => '/opt/mirrors/project/centos-altarch/sync-altarch > /dev/null',
```
jvperrin: I think this has to be `sync-archive`, not `sync-altarch`, given the file was renamed when it was moved.

modules/ocf_mirrors/manifests/kde_appdata.pp
```diff
@@ -0,0 +1,27 @@
...
+      ensure  => directory,
+      source  => 'puppet:///modules/ocf_mirrors/project/kde-appdata/',
+      owner   => mirrors,
+      group   => mirrors,
+      mode    => '0755',
+      recurse => true;
+  }
+
+  ocf_mirrors::monitoring {
+    'kde-appdata':
```
jvperrin: I think this needs to be `kde-applicationdata` (or the `local_path` attribute needs to be changed) since the path is `http://mirrors.ocf.berkeley.edu/kde-applicationdata/last-updated`, not `http://mirrors.ocf.berkeley.edu/kde-appdata/last-updated`.

modules/ocf_mirrors/manifests/trisquel.pp
kpengboy: My concern is that people will forget to update it, and then it will silently fail.

jvperrin:
I do think it would be nice if this didn't include the merge commit, since it's not really relevant here, but otherwise this looks good.

PR #280, opened 2018-02-02
Centralize logging for all VMs

jvperrin:
Also organize syslog to group by host and by service, since before all of the logs were just put into one directory, so with logrotate it got a little crazy.

### DIFF
--- a/modules/ocf/manifests/logging.pp
+++ b/modules/ocf/manifests/logging.pp
@@ -1,0 +1,21 @@
+class ocf::logging {
+  package { 'rsyslog': }
+
+  service { 'rsyslog':
+    require => Package['rsyslog'],
+  }
+
+  augeas { 'remote-syslog':
+    context => '/files/etc/rsyslog.conf',
+    changes => [
+      'set entry[last()+1]/selector/facility *',
+      'set entry[last()]/selector/level *',
+      'set entry[last()]/action/protocol @@',
+      'set entry[last()]/action/hostname syslog',
+      'set entry[last()]/action/port 514',
+    ],
+    onlyif  => "match entry[action/hostname = 'syslog'] size == 0",
+    notify  => Service['rsyslog'],
+    require => Package['rsyslog'],
+  }
+}
--- a/modules/ocf_syslog/files/ocf.conf
+++ b/modules/ocf_syslog/files/ocf.conf
@@ -5,6 +5,16 @@ input(type="imtcp" port="514" ruleset="ocf-remote")
 module(load="imudp")
 input(type="imudp" port="514" ruleset="ocf-remote")
 
+# global options when writing to or creating files
+module(
+  load="builtin:omfile"
+  template="ocf-remote"
+  fileGroup="ocfstaff"
+  fileCreateMode="0640"
+  dirGroup="ocfstaff"
+  dirCreateMode="0750"
+)
+
 # template for remote log lines
 template(name="ocf-remote" type="list") {
     property(name="timegenerated" dateFormat="rfc3339")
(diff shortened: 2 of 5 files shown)


modules/ocf_syslog/files/ocf.conf
```diff
@@ -15,41 +25,46 @@ template(name="ocf-remote" type="list") {
...
-    property(name="msg")
-    constant(value="]\n")
+# template for remote docker log filenames
+template(name="ocf-docker-file" type="list") {
+    constant(value="/var/log/remote/services/")
+    # in jessie, syslogtag has a weird ":" appended, so we have to regex that out;
+    # the regex bits can be removed when we're totally on stretch.
+    property(name="syslogtag" regex.Expression="^[^:]+" regex.Type="ERE" regex.Match="0" securepath="replace")
+    constant(value="/")
+    property(name="syslogtag" regex.Expression="^[^:]+" regex.Type="ERE" regex.Match="0" securepath="replace")
```
jvperrin: I'm not sure how to remove this duplication, but maybe it's fine for now? I think we're actually on stretch everywhere now, so we can probably remove the more complicated regex stuff here at least, but the duplication of the property will still remain.

modules/ocf_syslog/files/ocf.conf
```diff
@@ -15,41 +25,46 @@ template(name="ocf-remote" type="list") {
...
 template(name="ocf-remote-file" type="list") {
     constant(value="/var/log/remote/")
-    # in jessie, syslogtag has a weird ":" appended, so we have to regex that out;
-    # the regex bits can be removed when we're totally on stretch.
-    property(name="syslogtag" regex.Expression="^[^:]+" regex.Type="ERE" regex.Match="0" SecurePath="replace")
+    property(name="hostname" securepath="replace")
+    constant(value="/")
+    property(name="programname" caseconversion="lower" securepath="replace")
+    constant(value="/")
+    property(name="programname" caseconversion="lower" securepath="replace")
```
jvperrin: Same here with the duplication of the program name.

modules/ocf_syslog/files/ocf.conf
```diff
@@ -15,41 +25,46 @@ template(name="ocf-remote" type="list") {
...
     # Make sure this at least matches all the items let through in the regex in
     # the stdin2syslog script so that real logs aren't put in errors.log
-    if re_match($syslogtag, '^[a-zA-Z0-9:_-]+$') then {
-        action(type="omfile" dynaFile="ocf-remote-file" template="ocf-remote" fileGroup="ocfstaff" fileCreateMode="0640")
+    #
+    # The hostname matches docker-based hostnames only (12 hex characters) to
+    # separate them out into service logs instead of per host since we go
+    # through docker hosts quite quickly and we want to group logs by service
+    # for docker, not by host.
+    if re_match($hostname, '^[a-f0-9]{12}$') and re_match($syslogtag, '^[a-zA-Z0-9:_-]+$') then {
```
jvperrin: I'm not particularly happy about this, but I also don't think we're going to have hosts that have hex-based names and are 12 characters long outside of docker, so this works decently well.

modules/ocf/manifests/logging.pp
```diff
@@ -0,0 +1,21 @@
...
+  augeas { 'remote-syslog':
+    context => '/files/etc/rsyslog.conf',
+    changes => [
+      'set entry[last()+1]/selector/facility *',
+      'set entry[last()]/selector/level *',
+      'set entry[last()]/action/protocol @@',
+      'set entry[last()]/action/hostname syslog',
+      'set entry[last()]/action/port 514',
+    ],
+    onlyif  => "match entry[action/hostname = 'syslog'] size == 0",
```
jvperrin: I kind of hate augeas now lol

modules/ocf_syslog/files/ocf.conf
kpengboy: Yes, we're definitely completely on stretch now for the dockers. Duplication of the property? Whatever...

modules/ocf_syslog/files/ocf.conf
kpengboy: Generating a combined syslog would be an exercise left to the reader...

modules/ocf_syslog/files/ocf.conf
```diff
@@ -5,6 +5,16 @@ input(type="imtcp" port="514" ruleset="ocf-remote")
 module(load="imudp")
 input(type="imudp" port="514" ruleset="ocf-remote")
 
+# global options when writing to or creating files
+module(
+  load="builtin:omfile"
+  template="ocf-remote"
+  fileGroup="ocfstaff"
```
kpengboy: Are you sure it is the best idea to make these logs readable by all ocfstaff? They do sometimes contain e.g. private keys that Puppet unwisely outputs and stuff like that...

modules/ocf_syslog/files/ocf.conf
jvperrin: You can do something like `tail -f */*.log` to show a stream of the incoming logs for a host, but then it also shows each of the names of the logs it is printing from. I think from an organization perspective though this is better than having one really long log that's harder to find things in or one directory with a ton of logs ending in `.13.gz` or something. Even a `grep -r` is pretty easy to do, but if others prefer a flatter organization of logs then I'd be ok with that.

modules/ocf_syslog/files/ocf.conf
jvperrin: I'm just following the current log format, but I'd agree that sometimes puppet prints out things that we don't want shared by everyone. Maybe I can set the logs for hosts (not services) to be readable by `ocfroot` as the group? I'll give it a shot anyway

jvperrin: Ok, I made logs from hosts only readable by `ocfroot` now in fc0fa59df72d91d2c349238291188b2513de5d28, and I think that's a good move since you're right that they sometimes contain more sensitive information.

modules/ocf_syslog/files/ocf.conf
```diff
@@ -15,41 +25,47 @@ template(name="ocf-remote" type="list") {
...
-    # This regex was fiendish, since the hyphen must be next to the end bracket
-    # for it to work.
+    # The regex for the syslogtag was fiendish, since the hyphen must be next
+    # to the end bracket for it to work.
     #
     # It uses POSIX ERE (http://www.regular-expressions.info/posixbrackets.html)
     # and has a tester at http://www.rsyslog.com/regex/
     #
     # Make sure this at least matches all the items let through in the regex in
     # the stdin2syslog script so that real logs aren't put in errors.log
```
kpengboy: Are we still doing errors.log?

jvperrin: No, so this comment is wrong, I'll fix it. I was tempted to keep `errors.log`, but the main problem is that I don't think anything would get put into it unless it doesn't have a hostname set in the request, which I don't see happening.

PR #281, opened 2018-02-03
disallow wall access for non-root users

tmochida:
Recently I noticed several users were promoting blockchain@berkeley using the wall utility to broadcast messages to all users on tsunami. Getting frivolous broadcast messages can be annoying for users, as it opens room for spam.

I considered running `mesg n` on shell login, but that means we have to add it to the start files for bash,zsh, etc. Restricting access to the utility would be a better way to prevent spam.

### DIFF
--- a/modules/ocf/files/shell/bash.bashrc
+++ b/modules/ocf/files/shell/bash.bashrc
@@ -11,6 +11,9 @@ export LC_ALL=en_US.UTF-8
 # Non-restrictive umask
 umask 022
 
+# ignore messages from others
+mesg n
+
 # Set up bash-completion
 [ -r /etc/bash_completion ] && source /etc/bash_completion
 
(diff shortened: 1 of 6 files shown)


chriskuehl:
@abizer not really sure how setting group to `ocfroot` is going to work, it needs to be run as gid `tty` (which is why it's `root:tty` and setgid by default). The best we can probably do is just do what tmochida did and tell staff to use sudo if they need to make announcements, I'm not aware of any way to restrict it to `ocfstaff` but still be setgid `tty` (okay, there are ways, like moving it into another directory which only `ocfstaff` can traverse into, but it doesn't really seem worth it).

Unbelievable that a club would abuse this to spam terminals -- imo it reflects very poorly on their members' maturity.

redplanks:
Make Berkeley the US's first blockchain-free city by 2018

kpengboy:
It wouldn't be too hard to add `mesg n` to each shell's startup file would it? Determined parties could work around `wall` by e.g. creating a script that `write`s to every tty. But `mesg n` can't be worked around.

tmochida:
@kpengboy I added mesg n for all shell startup I found in modules/ocf/files/shell, can you review them? I'm not familiar with csh or fish. Thanks

PR #282, opened 2018-02-04
Only do SSH GSSAPI key exchange with hosts in OCF network

kpengboy:
This should hopefully solve the issue of it taking forever to SSH into
hosts outside the OCF network because those hosts attempt to
contact their nonexistent Kerberos servers or whatever it is


kpengboy:
I'll be testing this on supernova, please report any problems

dkess:
Just tried SSH'ing to my personal server from supernova and definitely noticed the speed increase, thanks for this.

PR #284, opened 2018-02-05
ocf_desktop: fix parsing of who command

kpengboy:
The number of spaces in the time field varies with the
locale setting. To deal with this variation, find the display
using the last field of the output rather than by hardcoding
a field number.


shex1627:
Please merge asap to keep the data in good quality.

PR #285, opened 2018-02-05
Decal VM puppet module

abizer:
We'll be hosting advanced student VMs on our/XCF infrastructure this semester, and have provisioned a VM to serve as a staff VM / bastion host for students to use to get into their VMs from the lab (e.g. `ssh <user> -J decal`), to host misc. administrative scripts like our checkoff scripts, and to host the decal website. This module should help manage that VM.


hieradata/nodes/lowgpa.yaml
```diff
@@ -0,0 +1,5 @@
+classes:
+    - ocf_decal
+
+ocf::auth::glogin: [[ocfdecal, ALL]]
```
abizer: this LDAP group remains to be made

modules/ocf_decal/manifests/init.pp
```diff
@@ -0,0 +1,21 @@
+class ocf_decal {
+
+  include ocf_decal::website
+
+  class { 'apache':
+    default_vhost => false;
+  }
+
+  user { 'ocfdecal':
```
kpengboy: Why not use the existing `decal` user?

modules/ocf_decal/manifests/website.pp
```diff
@@ -0,0 +1,43 @@
...
+      mode   => '0755',
+      require => User['ocfdecal'];
+  }
+
+  apache::vhost { 'decal-redirect':
+    servername => 'decal.ocf.berkeley.edu',
+    serveraliases => [
+      'decal.ocf.io',
+      'decal.xcf.sh',
+      'decal.xcf.berkeley.edu'
```
kpengboy: Could add `decal` here too if you wish

modules/ocf_decal/manifests/init.pp
abizer: I figured it was marginally better/safer to create a new user for this purpose.

kpengboy: Are we still going to be using the existing decal user then

abizer: I'm not sure, but I think not.

kpengboy: Also, if you create users in puppet like this, you'll want to designate them as system users lest they collide with LDAP users

kpengboy: (Or alternatively fix the UID to something guaranteed not to be used in the future)

jvperrin: Agreed about making it a system user instead, we've had problems with the user ids colliding in the past.

modules/ocf_decal/manifests/website.pp
```diff
@@ -0,0 +1,43 @@
...
+    redirect_status => 'permanent',
+    redirect_dest   => 'https://decal.ocf.berkeley.edu',
+  }
+
+  apache::vhost { 'decal-ssl':
+    servername => 'decal.ocf.berkeley.edu',
+    serveraliases => [
+      'decal.ocf.io',
+      'decal.xcf.sh',
+      'decal.xcf.berkeley.edu',
```
jvperrin: I think these should all be redirecting too to the canonical domain unless you want all of these to work too.

modules/ocf_decal/manifests/website.pp
```diff
@@ -0,0 +1,43 @@
...
+
+    port => 443,
+    docroot => '/srv/www/decal',
+    docroot_owner => 'ocfdecal',
+    docroot_group => 'www-data',
+
+    ssl => true,
+    ssl_key => '/etc/letsencrypt/keys/decal.key',
+    ssl_cert => '/etc/ssl/certs/decal.crt',
+    ssl_chain => '/etc/ssl/certs/lets-encrypt.crt',
```
jvperrin: A lot of these are magical in nature, I think providing them via puppet (from the private share) or creating directories (for example for `/srv/www/decal`) using puppet would be better.

modules/ocf_decal/manifests/website.pp
```diff
@@ -0,0 +1,43 @@
+class ocf_decal::website {
+
+  file {
+    ['/srv/www', '/srv/www/decal']:
+      ensure => directory,
+      owner  => ocfdecal,
+      group  => www-data,
+      mode   => '0755',
```
jvperrin: Why do you need to specify the mode here? Isn't this the default for directories?

modules/ocf_decal/manifests/init.pp
abizer: this is done

modules/ocf_decal/manifests/website.pp
abizer: Yeah it is, I'll remove this

modules/ocf_decal/manifests/website.pp
abizer: I think it should be ok, the short URLs are kinda nice

modules/ocf_decal/manifests/website.pp
abizer: Agreed. This stuff was basically copied from various manifests in ocf_www. I previously had this stuff living in `/srv/ssl/decal`, I can put it all there again, I guess.

modules/ocf_decal/manifests/init.pp
abizer: I manually changed the uid and gid on the VM

modules/ocf_decal/manifests/website.pp
```diff
@@ -0,0 +1,59 @@
...
+      owner   => ocfdecal,
+      group   => www-data,
+      mode    => '0440',
+      require => User['ocfdecal'];
+    '/srv/ssl/decal/decal.key':
+      source  => 'puppet:///private/decal.key',
+      mode    => '0440';
+    '/srv/ssl/decal/decal.crt':
+      source  => 'puppet:///private/decal.crt',
+      mode    => '0440';
```
jvperrin: I think you want some dependencies here, like `require => File['/srv/ssl/decal']`, but it's not a big deal in general since these should be executed in order in more recent versions of puppet anyway. It's good to have dependencies here though so that it's clear which other rules/statements are being depended on to execute each of these.

modules/ocf_decal/manifests/website.pp
```diff
@@ -0,0 +1,59 @@
...
+    redirect_dest   => 'https://decal.ocf.berkeley.edu',
+  }
+
+  apache::vhost { 'decal-ssl':
+    servername => 'decal.ocf.berkeley.edu',
+    serveraliases => [
+      'decal',
+      'decal.ocf.io',
+      'decal.xcf.sh',
+      'decal.xcf.berkeley.edu',
```
jvperrin: I still think you'd want to redirect these to have a more canonical domain instead of 5 different domains that people can use to get to the website. It's the standard pattern we follow elsewhere anyway.

abizer: I believe you made these changes yourself, so that should be done/included in the latest commit.

jvperrin: I don't think so, I didn't change anything related to this, the only changes I made were to enable `mod_rewrite`, fix the redirect from http (adding an ending slash), and enable overrides from `.htaccess` files. I still think this should be done though.

jvperrin: (Can be done after merging this though, this one is not a big deal either way)

abizer:
oops

hieradata/nodes/lowgpa.yaml
jvperrin: Are we still going to make this if we use DigitalOcean VMs for the advanced track?

abizer:
No I don't think so. I also need to add your redirect to canonical vhost
thing

On Feb 15, 2018 6:24 PM, "Jason Perrin" <notifications@github.com> wrote:

*@jvperrin* approved this pull request.

Nice, looks good to me with your new changes.

In hieradata/nodes/lowgpa.yaml
<https://github.com/ocf/puppet/pull/285#discussion_r168663684>:

> @@ -0,0 +1,5 @@
+classes:
+    - ocf_decal
+
+ocf::auth::glogin: [[ocfdecal, ALL]]

Are we still going to make this if we use DigitalOcean VMs for the advanced
track?

—
<https://github.com/ocf/puppet/pull/285#pullrequestreview-97070415>, or mute
the thread

jvperrin:
Ah yeah, I was skimming and thought I saw that, but it was the HTTP -> HTTPS redirect instead.

jvperrin:
Merging per @abizer's request

PR #286, opened 2018-02-06
Adding datathon survey shorturl

dongkyunc:


kpengboy:
Was this added to the reserved usernames list as well?

PR #291, opened 2018-02-25
Improve log-mail script

jvperrin:
This makes a few changes to the log-mail script:

- Decode encoded subjects and real names in emails by RFC 2047. (I might be able to use [email.header.decode_header](https://docs.python.org/3/library/email.header.html#email.header.decode_header) for this instead, so I'll try that).
- More reliably obtain the recipient email address, since sometimes the `To` header can not contain an email so it has to come from the `Received` headers instead.
- Add CC'd addresses to the logged output to more easily tell if the email is potentially spam or not.

It also bumps the number of emails that a user has to send per day before being flagged by the mail examining script, because it seems to be a bit too spammy even with fairly low levels of email usage.

### DIFF
--- a/modules/ocf_mail/files/spam/logging/log-mail
+++ b/modules/ocf_mail/files/spam/logging/log-mail
@@ -1,15 +1,19 @@
 #!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
 # Appends JSON dictionary of NSA-style metadata about a message read from stdin
 # to /var/log/ocfmail.log for the purposes of monitoring spam. Should be run as
 # the postfix user.
 #
 # Note that `date` returned is the current date (and not one parsed from
 # message headers), since we don't really trust headers more than necessary.
+import base64
 import email
 import email.utils
 import io
 import json
 import pwd
+import quopri
 import re
 import sys
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


jvperrin:
Yup, looks like `email.header.decode_header` works well for this, so I'll just use that instead, but I'll leave it on that version for a day or two to make sure it works correctly to log a variety of emails.

modules/ocf_mail/files/spam/logging/log-mail
```diff
@@ -1,15 +1,19 @@
 #!/usr/bin/env python3
+# -*- coding: utf-8 -*-
```
chriskuehl: don't need this in python3

did this get added automatically or something?

modules/ocf_mail/files/spam/logging/log-mail
```diff
@@ -1,15 +1,19 @@
...
 #
 # Note that `date` returned is the current date (and not one parsed from
 # message headers), since we don't really trust headers more than necessary.
+import base64
 import email
 import email.utils
 import io
 import json
 import pwd
+import quopri
```
chriskuehl: wow. it is not every day i discover a new module in the stdlib :P

modules/ocf_mail/files/spam/logging/log-mail
```diff
@@ -61,16 +92,35 @@ def parse_received_for_auth(header):
...
+        return ('', match.group(1))
+
+
+def clean_addr(realname, addr):
     """Cleans up email address, attempting to normalize as much as possible.
 
     We keep the realname in case it says something useful (e.g. 'Cron Daemon').
 
-    >>> clean_addr('Chris Kuehl <cKuEhL@OCF.Berkeley.EDU>')
+    >>> clean_addr(('Chris Kuehl', '<cKuEhL@OCF.Berkeley.EDU>'))
```
chriskuehl: is this accurate? seems like it takes 2 arguments, not one?

modules/ocf_mail/files/spam/logging/log-mail
jvperrin: I added this because python gave a warning about the encoding when I ran it due to the `™` in the docstring of `mime_decode`, but it was doing the right thing anyway, so I agree this could just be removed.

modules/ocf_mail/files/spam/logging/log-mail
jvperrin: Yeah, I forgot a `*` here (or to separate the arguments, that would be clearer), good catch.

jvperrin:
Hmmm, I think `decode_header` is working, but it seems to be printing out nothing in the log where the encoded text should be, it's like it's being removed. It happens with both the built-in method and my custom method though, so I'm not sure what's going on there. I'll try and investigate more, maybe there's something wrong with how `json` is dumping to the file again, that was an issue I ran into before.

modules/ocf_mail/files/spam/logging/log-mail
chriskuehl: weird, how were you running it? i haven't seen any warnings like this in python3

jvperrin:
Ok, changing the input encoding to `utf-8` and opening the log file as `utf-8` fixed the encoding issues, so we should see non-ascii email subjects now which is nice.

PR #292, opened 2018-02-25
Pin marathon-lb to a specific version (rt#6870)

jvperrin:
This will make updating marathon-lb possible instead of the current problem where upgrading marathon-lb is a very manual process, because restarting ocf-lb doesn't work because it'll use the current version downloaded already instead of fetching a new version. Pinning to a specific version makes this both more stable and makes it possible to upgrade when wanted, and I think it's better than having a cronjob that upgrades for us since that'll be more unpredictable about when we might get breaking changes.

### DIFF
--- a/modules/ocf_mesos/files/master/load_balancer/ocf-lb.service
+++ b/modules/ocf_mesos/files/master/load_balancer/ocf-lb.service
@@ -4,7 +4,7 @@ After=docker.service
 Wants=docker.service
 
 [Service]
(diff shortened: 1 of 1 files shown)


kpengboy:
Good reasoning, although I wonder whether people _will_ remember to update this.

jvperrin:
I do worry about that too, but I think as is, this is better than it was previously at least, since we can at least determine what version we are running without having to dive into the container to try to figure that out.

PR #293, opened 2018-02-25
Use firewall_multi

kpengboy:
This allows us to eliminate some foreach loops and make our configs prettier.

### DIFF
--- a/manifests/site.pp
+++ b/manifests/site.pp
@@ -47,7 +47,7 @@ Cron {
   environment => 'PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin'
 }
 
-Firewall {
+Firewall_multi {
   before => Class['ocf::firewall::post'],
 }
 
--- a/modules/ocf/manifests/firewall/output_printers.pp
+++ b/modules/ocf/manifests/firewall/output_printers.pp
@@ -1,11 +1,9 @@
 class ocf::firewall::output_printers {
   # Allow output to printers
-  ['papercut', 'pagefault'].each |$printer| {
-    firewall { "899 allow output to ${printer}":
-      chain       => 'PUPPET-OUTPUT',
-      destination => $printer,
-      proto       => 'all',
-      action      => 'accept',
-    }
+  firewall_multi { '899 allow output (printers)':
+    chain       => 'PUPPET-OUTPUT',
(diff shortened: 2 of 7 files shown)


manifests/site.pp
```diff
@@ -47,7 +47,7 @@
   environment => 'PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin'
 }
 
-Firewall {
+Firewall_multi {
```
kpengboy: Leaving this as `Firewall` doesn't work because the `firewall_multi` resources in `post.pp` then get tangled in dependency cycles. The way I've chosen to fix this is to apply the resource default to `firewall_multi` instead; a side effect is that we mustn't use `firewall` resources directly or else they won't be subject to these ordering requirements. Should we enforce that with a pre-commit hook? Or should we try to lobby the maintainer of `firewall_multi` to forward dependencies from `firewall_multi` resources to `firewall` resources? Or should we go to the inconvenience of not using `firewall-multi` in `post.pp` (and `pre.pp` when input rules are merged in)?

modules/ocf/manifests/firewall/output_printers.pp
```diff
@@ -1,11 +1,9 @@
...
-  ['papercut', 'pagefault'].each |$printer| {
-    firewall { "899 allow output to ${printer}":
-      chain       => 'PUPPET-OUTPUT',
-      destination => $printer,
-      proto       => 'all',
-      action      => 'accept',
-    }
+  firewall_multi { '899 allow output (printers)':
+    chain       => 'PUPPET-OUTPUT',
+    destination => ['papercut', 'pagefault'],
```
abizer: hardcoding hostnames seems like it could be risky for future maintenance

kpengboy: What should we do instead?

kpengboy: @abizer

kpengboy:
Anyone else have comments?

manifests/site.pp
jvperrin: I think using `Firewall_multi` here for now is best, and then maybe try to talk to the maintainer of the package and see what they think is best? It's a concern that I think they should hopefully be receptive to.

PR #295, opened 2018-02-27
DeCal canonical redirect + mysql.conf

abizer:


modules/ocf_decal/manifests/init.pp
```diff
@@ -14,10 +14,24 @@
...
+    '/opt/ocfdecal':
+      ensure  => directory,
+      owner   => ocfdecal,
+      group   => ocfdecal,
+      require => User['ocfdecal'];
+    '/etc/decal_mysql.conf':
+      source  => 'puppet:///private/mysql.conf',
+      owner   => ocfdecal,
+      group   => ocfstaff,
+      mode    => '0660',
```
jvperrin: I'd probably make this `0440` and maybe put a notice at the top of the file or something, since you don't want people to be editing it on the server instead of on the puppet private share (in the vein of other files we do this with).

modules/ocf_decal/manifests/init.pp
```diff
@@ -14,10 +14,24 @@
...
+      require => User['ocfdecal'];
+  }
+
+  vcsrepo { '/opt/share/decal-utils':
```
jvperrin: Looks good, not sure how we're going to handle the terraform stuff though, but that can happen later I suppose.

modules/ocf_decal/manifests/website.pp
```diff
@@ -27,38 +27,52 @@
...
     docroot         => '/srv/www/decal',
     redirect_status => 'permanent',
     redirect_dest   => 'https://decal.ocf.berkeley.edu/',
   }
 
-  apache::vhost { 'decal-ssl':
-    servername => 'decal.ocf.berkeley.edu',
-    serveraliases => [
+  apache::vhost { 'decal-canonical-redirect':
+    servername      => 'decal.ocf.io',
```
jvperrin: I feel like there might be a better way to do this, but I can't think of what it is right now. This is fine though, this looks good to me. Thanks for changing this!

modules/ocf_decal/manifests/init.pp
abizer: terraform @ decal is done I think, if we can't fix the broken tfstate thing

modules/ocf_decal/manifests/website.pp
abizer: likewise, but this seemed like the simplest way to implement this

PR #296, opened 2018-02-27
Remove rustc

ethanhs:
The rustc package interferes with rustup.rs, which people should use.


jvperrin:
_sigh_ I always forget this with puppet, yeah, you'll have to ensure it's purged to actually remove the package.

ethanhs:
Well, that's annoying. I'll make that change now.

PR #301, opened 2018-03-09
various docker things

Baisang:
so i can do my dumb meme projects

### DIFF
--- a/hieradata/type/desktop.yaml
+++ b/hieradata/type/desktop.yaml
@@ -7,4 +7,7 @@ ocf::auth::gsudo: [ocfstaff]
 # Don't manage staff home directories on desktops.
 ocf::staff_users::noop: true
 
+# Allow ocfroot to use docker without sudo
+ocf::packages::docker::admin_group: ocfroot
+
 staff_only: false
--- a/modules/ocf_desktop/manifests/packages.pp
+++ b/modules/ocf_desktop/manifests/packages.pp
@@ -1,5 +1,6 @@
 class ocf_desktop::packages {
   include ocf::extrapackages
+  include ocf::packages::docker
 
   # Install packages specific to desktops
   #


hieradata/type/desktop.yaml
```diff
@@ -7,4 +7,7 @@ ocf::auth::gsudo: [ocfstaff]
 # Don't manage staff home directories on desktops.
 ocf::staff_users::noop: true
 
+# Allow ocfroot to use docker without sudo
+ocf::packages::docker::admin_group: ocfroot
```
Baisang: from my understanding, allowing all `ocfstaff` may be a security concern?

chriskuehl: docker == root

if ocfstaff has root on desktops, it's okay to give them docker
if only ocfroot has root on desktops, we should probably keep docker to ocfroot only

modules/ocf_desktop/manifests/packages.pp
```diff
@@ -21,6 +22,8 @@
       'gnome-icon-theme', 'redshift', 'xfce4-whiskermenu-plugin']:;
     # desktop helpers
     ['libimage-exiftool-perl']:;
+    # development
+    ['docker-compose']:;
```
chriskuehl: is this really new enough for you? personally i'd just install this in a venv instead

hieradata/type/desktop.yaml
kpengboy: ocfstaff has root on desktops, but most staff can't use it because they don't have /root principals.

modules/ocf_desktop/manifests/packages.pp
kpengboy: Also this package isn't desktop specific, so if we wanted to install this it would be better to define this resource in `ocf::packages::docker` and make it optionally installed depending on the value of a parameter.

Baisang:
@jvperrin yeah i have some scripts where i want to use docker in parallel across desktops.

modules/ocf_desktop/manifests/packages.pp
Baisang: @chriskuehl actually youre right, its way too old. ill remove this commit

PR #302, opened 2018-03-13
Add prometheus daemon and node_exporter to ocf_stats

dkess:
This PR should add the Prometheus monitoring daemon to dementors. Since this is just a trial run, it's a minimal configuration that only monitors dementors itself and leprosy (my staffvm).

In the future, we'd want to install `node_exporter` on all hosts, and install specialized exporters for specific software (like Apache). I'll document other goals in the project (other goals include Grafana for more dashboards, alerting, etc)

Right now, this setup will run the `node_exporter` HTTP server on port 9100, and the Prometheus dashboard on port 9090 on dementors. I'd like to at least reverse proxy the dashboard through Apache (suggestions welcome on how to do this). We should also discuss the security model for this-- my intuition tells me that metrics should by firewalled off from the outside world, and the dashboard should be restricted to `ocfstaff` through HTTP auth, but dissent/discussion about these points is welcome.

### DIFF
--- a/modules/ocf_stats/manifests/prometheus.pp
+++ b/modules/ocf_stats/manifests/prometheus.pp
@@ -2,6 +2,7 @@
 class ocf_stats::prometheus {
   class { '::prometheus':
     version        => '2.0.0',
+    extra_options  => '--web.listen-address="127.0.0.1:9090" --web.route-prefix=/prometheus'
     alerts         => {},
     scrape_configs => [
       { 'job_name' => 'prometheus',
--- a/modules/ocf_stats/manifests/prometheus.pp
+++ b/modules/ocf_stats/manifests/prometheus.pp
@@ -9,7 +9,7 @@ class ocf_stats::prometheus {
         'scrape_timeout'  => '10s',
         'static_configs'  => [
           { 'targets' => [ 'localhost:9090' ],
-          'labels'  => { 'alias' => 'Prometheus'}
+            'labels'  => { 'alias' => 'Prometheus' }
           }
         ]
       },
--- a/modules/ocf_stats/manifests/prometheus.pp
+++ b/modules/ocf_stats/manifests/prometheus.pp
@@ -2,7 +2,7 @@
 class ocf_stats::prometheus {
   class { '::prometheus':
     version        => '2.0.0',
-    extra_options  => '--web.listen-address="127.0.0.1:9090" --web.route-prefix=/prometheus'
+    extra_options  => '--web.listen-address="127.0.0.1:9090" --web.route-prefix=/prometheus',
     alerts         => {},
     scrape_configs => [
       { 'job_name' => 'prometheus',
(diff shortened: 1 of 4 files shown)


dkess:
Also, for a live demo of what the dashboard looks like, visit port 9090 on my staffvm.

modules/ocf_stats/manifests/prometheus.pp
```diff
@@ -0,0 +1,35 @@
+# prometheus daemon config
+class ocf_stats::prometheus {
+  class { '::prometheus':
+    version        => '2.0.0',
+    alerts         => {},
+    scrape_configs => [
+      { 'job_name' => 'prometheus',
```
abizer: nit: formatting consistency

modules/ocf_stats/manifests/prometheus.pp
```diff
@@ -0,0 +1,35 @@
...
+        ]
+      },
+      { 'job_name' => 'node',
+        'scrape_interval' => '5s',
+        'scrape_timeout'  => '5s',
+        'static_configs'  => [
+          { 'targets' => [
+            # A list of all hosts to scrape.
+            # We should probably figure out a better way of enumerating all the
+            # hosts.
```
abizer: would hiera or puppetdb help here? this is a similar problem to the firewall issues of enumerating hosts. Ideally though we wouldn't have to manually specify which hosts to collect stats from.

redplanks: [Querying from PuppetDB](https://puppet.com/docs/puppetdb/5.1/api/query/tutorial.html#querying-with-puppet-code) would probably be the best option.

dkess: This can be specified in hiera (see the end of [the README](https://forge.puppet.com/puppet/prometheus/readme))-- would that be useful?

modules/ocf_stats/manifests/prometheus.pp
dkess: What specifically do you take issue with? Other than `'labels'` being too far to the left I don't see what's wrong here

abizer: labels, there's no space after `"alias" => "Prometheus"` that sort of minor thing

chriskuehl:
Since this is just a trial run, I would vote to do it on another VM (not dementors) so it's easier to clean up if we decide not to pursue this path.

dkess:
I don't mind setting up another VM for this, but I don't really think this adds a lot of complexity to dementors. The Prometheus daemon itself is fairly light. Most of  the complexity in this project will come from installing metrics exporters on all the hosts anyways.

If we do decide not to pursue this path, what kind of "clean up" would you expect other than just rolling back the Puppet config?

chriskuehl:
deleting config files, purging packages, deleting users, etc.

rolling back puppet will leave all that stuff in-place

can you just use your staff VM as the host for now?

dkess:
Sure, I'll close this PR for now and work on getting a more stable setup for my staff VM.

chriskuehl:
FWIW I think it's perfectly fine to still have a PR open, personally I'd just add the class to a staff VM (or other host) rather than an existing prod host.

dkess:
Reopening and repurposing this PR to put Prometheus on my staff VM.

PR #304, opened 2018-03-20
remove vestigial jessie code for wallpapers

abizer:


modules/ocf_desktop/manifests/xsession.pp
```diff
@@ -52,18 +52,11 @@
...
-    $wallpaper = $staff_only ? {
-      true  => 'background-staff.svg',
-      false => 'background.svg',
-    }
+  $wallpaper = $staff_only ? {
+    true  => 'background-staff.svg',
+    false => 'background.svg',
   }
-
+
```
kpengboy: Remove this

kpengboy:
Also did you run pre-commit...

modules/ocf_desktop/manifests/xsession.pp
abizer: lol I did this from the github web interface

modules/ocf_desktop/manifests/xsession.pp
```diff
@@ -52,16 +52,9 @@
   }
 
   # wallpaper symlink
-  if $::lsbdistcodename == 'jessie' {
-    $wallpaper = $staff_only ? {
-      true  => 'background-staff.png',
-      false => 'background.png',
```
jvperrin: Another change to make is to remove both these background images from this repo, since we don't use them any more.

PR #305, opened 2018-03-26
add wakeonlan to desktops

Baisang:
i want to use `lab-wakeup` on desktops

i think this is generally useful for staff as well


abizer:
Just ssh into supernova and do it from there?

Baisang:
makes my script slightly more complicated but i just figure this would be nice to have on desktops, especially since we have `lab-wakeup` on desktops already but it fails when you try to run it due to `wakeonlan` not being installed

PR #307, opened 2018-03-27
[WIP] HPC Puppet module

redplanks:
I'd like to make a rolling release kind of PR where I leave this up, and switch the HPC nodes to my env to test the module out. Why?

1. I'm not sure I know what I'm doing, so criticism and suggestions would be very helpful.
2. SLURM strongly recommends that all nodes have a uniform config, and right now it's hard to tell if the configs on each node are the same, as well as who is making what changes.
3. Right now, we don't have any users, so breakages would be minimal-impact. I'd of course like to get the Puppet module into a stable state and add all the features we need before we start taking users.

Right now, I think the module is ready to be put on all the nodes for a rolling release, but please let me know if you think otherwise.


modules/ocf_hpc/files/slurmdbd.conf
```diff
@@ -0,0 +1,23 @@
+ArchiveEvents=yes
+ArchiveJobs=yes
+ArchiveResvs=yes
+ArchiveSteps=no
+ArchiveSuspend=no
+#ArchiveScript=/usr/sbin/slurm.dbd.archive
+#AuthInfo=/var/run/munge/munge.socket.2
+AuthType=auth/munge
+DbdHost=segfault
+DbdBackupHost=corruption
```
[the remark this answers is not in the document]
abizer: if possible, that is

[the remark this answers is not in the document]
redplanks: That sounds like a good idea.

redplanks:
Happy to report that it works! (I think...)

modules/ocf_hpc/files/slurmdbd.conf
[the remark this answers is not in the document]
redplanks: Unfortunately, SLURM does not accept CNAMEs in configs, since it validates that the hostname specified in the config for the controller and DBD host match the hostname given by hostname -s

modules/ocf_hpc/manifests/compute.pp
```diff
@@ -0,0 +1,34 @@
...
+    group   => 'slurm',
+  }
+  # Kernel option to enable memory as a consumable gres resource
+  augeas { 'grub':
+    incl    => '/etc/default/grub',
+    lens    => 'ShellVars.lns',
+    changes => [
+      "set GRUB_CMDLINE_LINUX '\"cgroup_enable=memory swapaccount=1\"'",
+    ],
+  } -> exec { 'update-grub':
```
kpengboy: This will cause `update-grub` to run every time Puppet runs, which probably isn't what you want.

Try instead setting the `refreshonly => true` parameter on the `update-grub` resource and changing the `->` to a `~>`. This way, `update-grub` will only run when augeas actually modifies grub's configuration file.

modules/ocf_hpc/manifests/controller.pp
```diff
@@ -0,0 +1,35 @@
...
+
+  $slurmdbd_mysql_password = hiera('hpc::controller::slurmdbd_mysql_password')
+
+  package { 'slurmdbd':
+  } -> file { '/etc/slurm-llnl/slurmdbd.conf':
+    source  => 'puppet:///modules/ocf_hpc/slurmdbd.conf',
+    mode    => '0600',
+    owner   => 'slurm',
+    group   => 'slurm',
+  } -> augeas { 'slurmdbd.conf':
```
kpengboy: Shouldn't this kind of thing be done with templates?

modules/ocf_hpc/manifests/singularity.pp
```diff
@@ -0,0 +1,22 @@
...
+    source => 'puppet:///modules/ocf_hpc/singularity.conf',
+    mode   => '0644',
+    owner  => 'root',
+    group  => 'root',
+  }
+
+  # Each running singularity container consumes a loop device,
+  # and the default max loop devices is 8.
+  # Requires reboot after change.
+  augeas { 'modules':
```
kpengboy: Consider naming this resource something like `Set singularity max loop devices`. Resource names must be globally unique across all types, and `modules` is an awfully generic name.

modules/ocf_hpc/manifests/singularity.pp
```diff
@@ -0,0 +1,22 @@
+class ocf_hpc::singularity {
+
+  package { 'singularity-container':
+    install_options => ['-t stretch-backports'],
+  }
```
kpengboy: The canonical way to do this in OCF puppet code is:

```puppet
ocf::repackage { 'singularity-container':
    backport_on => ['stretch'],
}
```

modules/ocf_hpc/manifests/compute.pp
```diff
@@ -0,0 +1,35 @@
...
+  # Kernel option to enable memory as a consumable gres resource
+  augeas { 'grub':
+    incl        => '/etc/default/grub',
+    lens        => 'ShellVars.lns',
+    changes     => [
+      "set GRUB_CMDLINE_LINUX '\"cgroup_enable=memory swapaccount=1\"'",
+    ],
+    refreshonly => true
+  } ~> exec { 'update-grub':
+    user => 'root',
```
kpengboy: You'll still want to set `refreshonly => true`

redplanks: Ah I set it in augeas for some reason, mb

modules/ocf_hpc/templates/slurmdbd.conf.erb
```diff
@@ -0,0 +1,24 @@
...
+PurgeResvAfter=1month
+PurgeStepAfter=1month
+PurgeSuspendAfter=1month
+LogFile=/var/log/slurm-llnl/slurmdbd.log
+#PidFile=/var/tmp/jette/slurmdbd.pid
+SlurmUser=slurm
+StorageType=accounting_storage/mysql
+StorageUser=ocfhpc
+StoragePass=<%= @slurmdbd_mysql_password %>
+StorageHost=mysql.ocf.berkeley.edu
```
jvperrin: Can probably just do `mysql` here instead of putting the whole FQDN if you'd like to be a little less verbose, but both will work.

modules/ocf_hpc/manifests/compute.pp
```diff
@@ -0,0 +1,34 @@
+class ocf_hpc::compute {
+  require ocf_hpc
+
+  # install proprietary nvidia drivers
+  package { ['nvidia-driver', 'nvidia-settings', 'nvidia-cuda-toolkit']:; }
```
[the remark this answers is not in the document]
redplanks: cuDNN is a userspace library, so users can install that on their Singularity containers (nvidia apparently provides Docker images with cuDNN which can be built to Singularity). Also, there's no cuDNN package for Debian, so it would be a pain to puppet.

PR #310, opened 2018-04-06
ocf_mail: raise max message size to 25MiB

kpengboy:
A member was trying to send attachments with a size of slightly larger than 10MB, and found that his emails were getting truncated and not going through.

Gmail and probably other mail providers use 25MB as the max attachment size, so let's use that instead.


kpengboy:
Any security considerations I may have missed?

chriskuehl:
nothing i can think of

PR #311, opened 2018-04-07
Add bash completion for how command

kpengboy:
I've found it frequently useful in the past. I've been using this completion for several months now and nothing has blown up.


daradib:
Neat! I made it a little faster so the minimum length limit isn't needed. Please test that it still works for you.

kpengboy:
Thanks @daradib, this is a lot faster.

kpengboy:
It still seems to take a suspiciously slow amount of time on jessie hosts, but whatever..?

PR #312, opened 2018-04-08
 Re-enable arrow alignment puppet-lint check

kpengboy:
It used to give pathological results for things like

```puppet
thing => {
  a => b,
  b => c,
},
```

but it seems to have been fixed in newer versions of puppet-lint.

Also auto-fix Puppet style errors.

### DIFF
--- a/modules/ocf/manifests/nginx_proxy.pp
+++ b/modules/ocf/manifests/nginx_proxy.pp
@@ -32,32 +32,32 @@ define ocf::nginx_proxy(
         proxy_redirect   => $proxy_redirect,
         proxy_set_header => concat($base_headers, $proxy_set_header),
 
-        listen_port => 443,
-        ssl         => true,
-        ssl_cert    => $ssl_cert,
-        ssl_key     => $ssl_key,
-        ssl_dhparam => $ssl_dhparam,
+        listen_port      => 443,
+        ssl              => true,
+        ssl_cert         => $ssl_cert,
+        ssl_key          => $ssl_key,
+        ssl_dhparam      => $ssl_dhparam,
 
-        add_header => merge({
+        add_header       => merge({
           # HSTS header
           'Strict-Transport-Security' => 'max-age=31536000',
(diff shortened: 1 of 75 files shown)


modules/ocf/manifests/nginx_proxy.pp
```diff
@@ -32,32 +32,32 @@
...
+        ssl_dhparam      => $ssl_dhparam,
 
-        add_header => merge({
+        add_header       => merge({
           # HSTS header
           'Strict-Transport-Security' => 'max-age=31536000',
         }, $headers),
 
-        * => $nginx_options;
+        *                => $nginx_options;
```
jvperrin: lol this one is a bit questionable, but alright

modules/ocf_puppet/manifests/puppetboard.pp
```diff
@@ -19,7 +19,7 @@
       $::hostname,
       $::fqdn,
     ],
-    proxy => 'http://lb.ocf.berkeley.edu:10009',
-    ssl   => true,
+    proxy          => 'http://lb.ocf.berkeley.edu:10009',
+    ssl            => true,
```
jvperrin: Meh, ok sure

jvperrin: Maybe putting this before the server aliases would make it more obvious about the lining up, but I don't think it really matters.

modules/ocf_stats/manifests/prometheus.pp
```diff
@@ -5,27 +5,31 @@
...
         'static_configs'  => [
-          { 'targets' => [ 'localhost:9090' ],
-            'labels'  => { 'alias' => 'Prometheus' }
-          }
-        ]
+          {
+            'targets' => [ 'localhost:9090' ],
+            'labels'  => { 'alias' => 'Prometheus' },
+          },
+        ],
```
jvperrin: This is actually pretty nice

PR #314, opened 2018-04-10
fileserver.conf is no longer a template

abizer:
please correct me if I'm wrong but this file no longer appears to be a template


redplanks:
The file is in fact no longer a template, but is now broken (lol)

PR #315, opened 2018-04-10
Make Brave less needy

kkuehler:
There was no documentation on this, so I was just deleting files in `~/.config/brave` until I found one that reset the browser settings. `~/.config/brave/session-store-1` is created after a user starts the browser, but I found that creating one with minimal settings will cause brave to keep these and add the other defaults on start.

Starts brave with [these settings](https://i.fluffy.cc/H3dJwg9SDLmlpgsSlVncqP6CvrdsmdQN.png), as recommended by someone in their Discord. It also closely mirrors the ephemeral mode we run our other browsers in. Lastly, gets rid of the annoying message where Brave asks to be the default browser and make OCF the default homepage.

Unfortunately I could not find a place to put this so that the settings are global. Again, very little documentation.

### DIFF
--- a/modules/ocf_desktop/files/skel/config/brave/session-store-1
+++ b/modules/ocf_desktop/files/skel/config/brave/session-store-1
@@ -1,0 +1,14 @@
+{
+    "settings":{
+      "general.check-default-on-startup":false,
+      "general.is-default-browser":false,
+      "general.homepage":"https://www.ocf.berkeley.edu/about/lab/open-source",
+      "general.startup-mode":"homePage",
+      "shutdown.clear-cache":true,
+      "shutdown.clear-all-site-cookies":true,
+      "shutdown.clear-autofill-data":true,
+      "shutdown.clear-downloads":true,
+      "shutdown.clear-autocomplete-data":true,
+      "shutdown.clear-history":true
+   }
+}
--- a/modules/ocf_desktop/files/skel/config/brave/session-store-1
+++ b/modules/ocf_desktop/files/skel/config/brave/session-store-1
@@ -1,14 +1,15 @@
 {
-    "settings":{
-      "general.check-default-on-startup":false,
-      "general.is-default-browser":false,
-      "general.homepage":"https://www.ocf.berkeley.edu/about/lab/open-source",
-      "general.startup-mode":"homePage",
-      "shutdown.clear-cache":true,
-      "shutdown.clear-all-site-cookies":true,
-      "shutdown.clear-autofill-data":true,
-      "shutdown.clear-downloads":true,
-      "shutdown.clear-autocomplete-data":true,
-      "shutdown.clear-history":true
+    "settings": {
+      "general.check-default-on-startup": false,
+      "general.is-default-browser": false,
+      "general.homepage": "https://www.ocf.berkeley.edu/about/lab/open-source",
+      "general.startup-mode": "homePage",
+      "shutdown.clear-cache": true,
+      "shutdown.clear-all-site-cookies": true,
+      "shutdown.clear-autofill-data": true,
+      "shutdown.clear-downloads": true,
+      "shutdown.clear-autocomplete-data": true,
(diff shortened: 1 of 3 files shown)


modules/ocf_desktop/files/skel/config/brave/session-store-1
```diff
@@ -0,0 +1,15 @@
+{
+    "settings": {
+      "general.check-default-on-startup": false,
+      "general.is-default-browser": false,
+      "general.homepage": "https://www.ocf.berkeley.edu/about/lab/open-source",
```
jvperrin: Get this information [from hiera](https://github.com/ocf/puppet/blob/master/hieradata/common.yaml#L23) instead?

modules/ocf_desktop/files/skel/config/brave/session-store-1
```diff
@@ -0,0 +1,15 @@
...
+      "general.startup-mode": "homePage",
+      "shutdown.clear-cache": true,
+      "shutdown.clear-all-site-cookies": true,
+      "shutdown.clear-autofill-data": true,
+      "shutdown.clear-downloads": true,
```
jvperrin: Does this delete files in the Downloads directory when it is closed, or does it just clear them from inside the browser itself? The reason I ask is that I don't think we want to remove the files themselves, just the information stored in the browser (so I'm hoping it does that)

modules/ocf_desktop/files/skel/config/brave/session-store-1
jvperrin: This means that this file would need to be a template, but you can do something like this:

https://github.com/ocf/puppet/blob/master/modules/ocf/templates/firefox/prefs.js.erb#L1

modules/ocf_desktop/files/skel/config/brave/session-store-1
kkuehler: Just download history. Kinda a misleading variable name. :<

jvperrin: Oh ok, that's good then, I was hoping that's what it did.

modules/ocf_desktop/files/skel/config/brave/session-store-1
kkuehler: Yeah, good idea. In that case would it be better to put the `file` declaration in [brave.pp](https://github.com/ocf/puppet/blob/master/modules/ocf/manifests/packages/brave.pp)?

```  
file {
    '/etc/skel/.config/brave/session-store-1':
      content => template('ocf/brave/session-store-1.erb'),
      require => Package['brave'];
  }
```

jvperrin: Yeah, I think it probably would be better to do that, is there no system-wide config we could use at all? :/

kkuehler: [I don't think so](https://github.com/brave/browser-laptop/wiki/Filesystem-Hierarchy). After a bit of poking around it looks like they load the app state [here](https://github.com/brave/browser-laptop/blob/0b9e0a8b122196554d28c2b63176fc81faef7460/app/sessionStore.js#L961). The [getStoragePath()](https://github.com/brave/browser-laptop/blob/0b9e0a8b122196554d28c2b63176fc81faef7460/app/sessionStore.js#L58-L60) uses `app.getPath()` which, according to [Electron](https://github.com/electron/electron/blob/master/docs/api/app.md#appgetpathname), points to `~/.config`. Couldn't find any method that looks for global settings first =/. No one on their Discord has gotten back to me either.

jvperrin: From poking around the package installed, that doesn't add anything either, so I think just going with what you have is ok, maybe add a `TODO` about it though so that we can switch if/when a global config is added.

kkuehler:
@jvperrin changes made.

PR #317, opened 2018-04-16
Add Prometheus node_exporter to all hosts

dkess:
This PR adds the `node_exporter` to all hosts, and configures Prometheus to monitor all hosts.


modules/ocf/manifests/init.pp
```diff
@@ -12,6 +12,7 @@
   include ocf::munin::node
   include ocf::networking
   include ocf::packages
+  include ocf::prometheus_node
```
abizer: I'd organize this like the munin class above, e.g. `ocf::prometheus::node`, to maintain the existing pattern

modules/ocf_stats/manifests/prometheus.pp
```diff
@@ -1,5 +1,9 @@
 # prometheus daemon config
 class ocf_stats::prometheus {
+  # The list of nodes to monitor-- for now, monitor all hosts.
+  $nodes_query = '["from", "nodes", ["=", "expired", null]]'
+  $nodes = sort(puppetdb_query($nodes_query).map |$value| { $value["certname"] })
+
   class { '::prometheus':
     version        => '2.0.0',
     extra_options  => '--web.listen-address="127.0.0.1:9090" --web.external-url=http://127.0.0.1/prom',
```
abizer: I'd also call this `/prometheus` instead of just `/prom` since it's marginally more descriptive for someone reading this (and the future nginx config, I imagine) in the future.

modules/ocf/manifests/init.pp
dkess: Does it make sense to have a folder with just one file in it? That's the only reason why I did it this way.

dkess:
This PR is done and is open for comments, please review the changes.

modules/ocf_stats/files/prometheus/gen-prometheus-nodes
```diff
@@ -0,0 +1,37 @@
...
+from sys import argv
+
+from ocflib.infra.hosts import hosts_by_filter
+
+hosts = []
+for vals in hosts_by_filter('(|(type=server)(type=desktop)(type=staffvm))'):
+    cn, ip, host_type = vals['cn'][0], vals['ipHostNumber'][0], vals['type'][0]
+
+    # Don't include hozer machines
+    if cn.startswith('hozer-'):
```
redplanks: Why not include hozers?

modules/ocf_stats/manifests/prometheus.pp
```diff
@@ -1,8 +1,21 @@
...
 class ocf_stats::prometheus {
+  file {
+    '/usr/local/bin/gen-prometheus-nodes':
+      source => 'puppet:///modules/ocf_stats/prometheus/gen-prometheus-nodes',
+      mode   => '0755';
+  }
+
+  cron { 'gen-prometheus-nodes':
+    command => '/usr/local/bin/gen-prometheus-nodes /var/local/prometheus-nodes.json',
+    user    => 'root',
```
redplanks: I think the default value is root, so the line isn't strictly necessary.

modules/ocf_stats/manifests/prometheus.pp
```diff
@@ -1,8 +1,21 @@
...
+    minute  => '03',
```
kkuehler: nit: just do '3'.

EDIT: Saw the `refresh_interval` below.
~Also, is this meant to run every 3 minutes? If so you are looking for `*/3`. Just asking because the 3rd minute of every hour seems a little arbitrary~

kkuehler: Also, every hour seems a little excessive for how often the nodes are updated. Realistically, how often do these change?

modules/ocf_stats/manifests/prometheus.pp
```diff
@@ -1,8 +1,21 @@
...
+    require => File['/usr/local/bin/gen-prometheus-nodes'];
+  }
+
   class { '::prometheus':
     version        => '2.0.0',
-    extra_options  => '--web.listen-address="127.0.0.1:9090" --web.external-url=http://127.0.0.1/prom',
+    extra_options  => '--web.listen-address="127.0.0.1:9090" --web.external-url=http://127.0.0.1/prometheus --storage.tsdb.retention=2y',
     alerts         => {},
```
kkuehler: Is this necessary?

modules/ocf_stats/manifests/prometheus.pp
```diff
@@ -20,20 +33,14 @@
...
+
+        'file_sd_configs' => [
           {
-            'targets' => [
-              # A list of all hosts to scrape.
-              # We should probably figure out a better way of enumerating all the
-              # hosts.
-              'leprosy:9100',
-            ],
+            'files'            => [ '/var/local/prometheus-nodes.json' ],
```
kkuehler: So what happens if we puppet a new machine at 12:04 (assuming you keep the cron job to update the nodes at the 3rd minute of every hour)? The file `/var/local/prometheus-nodes.json` doesn't exist. How will prometheus handle this?

modules/ocf_stats/files/prometheus/gen-prometheus-nodes
dkess: I took this from `gen-munin-nodes`, but I guess it doesn't hurt to monitor hozers.

modules/ocf_stats/manifests/prometheus.pp
dkess: Again this was taken from the munin configuration, but I'll definitely change this one because it seems silly.

modules/ocf_stats/manifests/prometheus.pp
dkess: Retention is 14 days by default. After some discussion in the `#prometheus` IRC channel, I was told that it makes sense to retain more until we set up a proper long-term storage backend (which will happen in a seprate PR).

modules/ocf_stats/manifests/prometheus.pp
dkess: Not sure what you mean? This file will always exist. If we puppet a machine at the 4 minute mark, we'll have to either wait an hour for `gen-prometheus-nodes` to run again, or just manually run it ourselves.

modules/ocf_stats/manifests/prometheus.pp
dkess: I took this from the munin config which does it at the 3rd minute as well. It doesn't really hurt to query LDAP every hour, and that's about how long we can acceptably wait before monitoring starts for a new host, so I think it's fine.

modules/ocf_stats/files/prometheus/gen-prometheus-nodes
jvperrin: I think including hozers would be fine, might be nice to see stats on them if some service is being tested, etc.

modules/ocf_stats/manifests/prometheus.pp
```diff
@@ -1,8 +1,25 @@
...
 class ocf_stats::prometheus {
+  file {
+    '/usr/local/bin/gen-prometheus-nodes':
+      source => 'puppet:///modules/ocf_stats/prometheus/gen-prometheus-nodes',
+      mode   => '0755';
+  }
+
+  cron { 'gen-prometheus-nodes':
+    command => '/usr/local/bin/gen-prometheus-nodes /var/local/prometheus-nodes.json',
+    minute  => '03',
```
jvperrin: I'm guessing this was taken from `gen-munin-nodes`:
https://github.com/ocf/puppet/blob/26109d27c9c19d1b683c9a32af428a0a01ebe51c/modules/ocf_stats/manifests/munin.pp#L27-L35

The `03` minute argument there existed because munin runs every 5 minutes, and generating the nodes while munin was going to run gave problems (rt#4712), as mentioned in the comment, so I'd probably make this on the hour or something instead to make it more regular and predictable.

modules/ocf_stats/manifests/prometheus.pp
```diff
@@ -1,8 +1,25 @@
...
+    require => File['/usr/local/bin/gen-prometheus-nodes'];
+  }
+
+  exec { 'gen-promethues-nodes-initial':
+    command => 'usr/local/bin/gen-prometheus-nodes /var/local/prometheus-nodes.json',
+    onlyif  => 'test ! -f /var/local/prometheus-nodes.json',
```
jvperrin: A better way to do this is to use the [`creates` parameter](https://puppet.com/docs/puppet/5.3/types/exec.html#exec-attribute-creates) here. I suppose my feelings are that having this exec here seems a bit strange since there's a cronjob to do the same thing but regularly, but I see the problem that this file won't exist for up to an hour on bootstrapping. If the file doesn't exist for that hour, does Prometheus error or does it just show no data?

modules/ocf_stats/manifests/prometheus.pp
jvperrin: This isn't true any more with the exec resource on bootstrapping, right?

modules/ocf/manifests/prometheus/node.pp
```diff
@@ -0,0 +1,3 @@
+class ocf::prometheus::node {
+  include prometheus::node_exporter
+}
```
jvperrin: This is probably fine, but I feel like just including the node exporter directly in the init manifest would also be fine, you don't have to make this wrapper class if there's nothing to put in it yet.

modules/ocf_stats/manifests/prometheus.pp
kkuehler: Yee, he fixed it in a later commit after we chatted on irc

modules/ocf_stats/manifests/prometheus.pp
dkess: Haven't tested it, but I would expect Prometheus to just do nothing if the file doesn't exist, and then detect when the file gets created. I decided to put the exec command here since I think if we ever re-provision the stats machine, it would be nice to collect stats right away.

modules/ocf/manifests/prometheus/node.pp
dkess: Sure, I'll get rid of this file.

modules/ocf_stats/manifests/prometheus.pp
dkess: Done. As an aside, maybe `gen-munin-nodes` should do the same atomic file replace that's used in my script?

jvperrin: Yeah, that's probably a better idea than what I did.

modules/ocf_stats/files/prometheus/gen-prometheus-nodes
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+"""Generate a list of hosts to give to prometheus for stats generation"""
+import json
+import os
+import tempfile
+from sys import argv
+
+from ocflib.infra.hosts import hosts_by_filter
+
+hosts = []
```
kpengboy: Consider using a `main()` function here

PR #318, opened 2018-04-18
kill labstats 2: electric boogaloo

abizer:
session tracking now tries to work as such:

Every minute, a cronjob runs `update.sh` - if a user is logged in, it makes a request to ocfweb to refresh the session. If no user is logged in, the state is inactive, and no request is made.

On user login, lightdm/session-setup triggers `update-delay.sh`, which calls `update.sh`, which, since a user is logged in, has STATE="active", and the request to ocfweb to results in creation of a new session.

On user logout, lightdm/session-cleanup explictly sets STATE="cleanup" through update-delay.sh -> update.sh, which results in a request to ocfweb explicitly closing the session. This is necessary because the next time the cronjob runs, no user will be logged in, the state will be inactive, and no request will be made to ocfweb, thus relying on periodic session cleanup to close the session.

ca.crt is also being removed because I don't see it being used anywhere anymore.


dkess:
Is there a mechanism for cleaning up a session in the case where the cleanup request never gets sent? (ie computer does a hard shutdown)

modules/ocf_desktop/files/stats/update.sh
```diff
@@ -1,15 +1,21 @@
...
-	DATA="state=active&user=$CUR_USER"
-fi
+[[ -n "$CUR_USER" ]] && STATE="active" || STATE=${1:-"inactive"}
 
-curl --data "$DATA" \
-    https://labstats.ocf.berkeley.edu:444/update.cgi \
-	2>/dev/null
+# use 3 states for this minor optimization: only make requests to ocfweb
+# when a user is logged or once during cleanup, otherwise, stay silent
+if [ "$STATE" != "inactive" ]; then
```
jvperrin: What's the point of having an inactive state if it's never used? This was why I was confused yesterday, because it seems to me you only need two states, one for when a user is logged in (`active`), and another one that fires when a user logs out (`inactive` or `cleanup`, whichever you prefer). Then, as long as the logout script fires (or even if it doesn't, due to the `_cleanup_sessions()` function), then the user should be logged out as long as you send whatever this end state is. Essentially you only need to update if a user is logged in, or on cleanup, which I think only needs 2 states overall.

modules/ocf_desktop/files/stats/update.sh
```diff
@@ -1,15 +1,21 @@
 #!/bin/bash
-# imvirt spews dmesg errors on stderr
-# https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=842226
-[ "$(imvirt 2>/dev/null)" == "Physical" ] || exit 0 # only count hours on desktops
```
jvperrin: Why is this removed? I think we still only want to count hours on _physical_ desktops, not on VMs set up like desktops (for example).

abizer: good point, I did not consider that case.

jvperrin: It's not especially relevant right now since all our desktop setups are physical, but there used to be a VM set up like a desktop (informally, but it had the same scripts and stuff that a desktop had)

jvperrin:
@dkess https://github.com/ocf/puppet/blob/master/modules/ocf_stats/files/stats/bin/close-old-sessions

modules/ocf_desktop/files/stats/update.sh
abizer: Ah, I see what you mean. I've been treating `inactive` as a flag but it can be implied by
```
[[ -n "$CUR_USER" ]] && STATE="active" || STATE="$1"
if [[ -n "$STATE" ]]; then
  curl ...
fi
```
while maintaining the same optimization, and only using `active` and `cleanup` as explicit states.

modules/ocf_desktop/files/stats/update.sh
kpengboy: Maybe edit the comment to explain why there might be non-physical "desktops" then

modules/ocf_desktop/files/stats/update.sh
```diff
@@ -1,15 +1,19 @@
...
+# "cleanup" triggers closing of the session
+# other files to look at: session-cleanup and session-setup in /etc/lightdm
 
 CUR_USER=$(who | awk '$NF == "(:0)" { print $1 }')
-DATA="state=inactive"
 
-if [ -n "$CUR_USER" ]; then
-	DATA="state=active&user=$CUR_USER"
-fi
+[[ -n "$CUR_USER" ]] && STATE="active" || STATE="$1"
```
jvperrin: What does the last part of this do? Could you add a comment about what this does to make it a bit clearer?

modules/ocf_desktop/manifests/stats.pp
```diff
@@ -27,9 +27,10 @@
   }
 
   cron { 'labstats':
-    ensure  => present,
-    command => '/opt/stats/update.sh > /dev/null',
-    user    => 'ocfstats',
-    minute  => '*';
+    ensure      => present,
+    environment => 'PATH=/usr/local/sbin:/usr/local/bin:/bin:/usr/sbin:/usr/bin:/opt/puppetlabs/bin',
```
jvperrin: Could we just add `/opt/puppetlabs/bin` in the global cron path to make this cleaner?

https://github.com/ocf/puppet/blob/26109d27c9c19d1b683c9a32af428a0a01ebe51c/manifests/site.pp#L38-L48

abizer:
@jvperrin @dkess @kpengboy @kkuehler 

any remaining thoughts?

modules/ocf_desktop/manifests/stats.pp
```diff
@@ -27,9 +27,10 @@
   }
 
   cron { 'labstats':
-    ensure  => present,
-    command => '/opt/stats/update.sh > /dev/null',
-    user    => 'ocfstats',
-    minute  => '*';
+    ensure      => present,
+    environment => 'PATH=/usr/local/sbin:/usr/local/bin:/bin:/usr/sbin:/usr/bin',
```
kpengboy: Don't think you need or want this line?

kpengboy:
@abizer whenever you're ready...

PR #319, opened 2018-04-18
Puppet overheat

jvperrin:
It turns out that not that much actually needed to be done to get puppet working well on the Raspberry Pi (the LED sign). Currently it's off for some repairs, but I tested puppet with it yesterday and we could turn it back on, but it wouldn't work with the sign for now.


modules/ocf/manifests/networking.pp
```diff
@@ -39,8 +39,12 @@
   } else {
     # Find the first network interface that starts with 'en', since those are
     # ethernet interfaces. (Won't work for the raspberry pi, since it uses wifi)
-    $ifaces_array = split($::interfaces, ',')
-    $br_iface = grep($ifaces_array, 'en.+')[0]
+    if $::lsbdistid == 'Raspbian' {
+      $br_iface = 'wlan0'
```
jvperrin: I would like to find a good way to get this without hardcoding it, but I don't think there's an easy way to do it and we don't want it to use ethernet instead of wifi when configured.

modules/ocf/manifests/packages.pp
```diff
@@ -125,6 +127,16 @@
     ]:;
   }
 
+  # Packages to only install on Debian (not on Raspbian for example)
+  if $::lsbdistid == 'Debian' {
+    package {
+      [
+        'aactivator',
+        'fluffy',
```
jvperrin: I think we can make an apt repo and just upload these (along with ocflib if we can build it correctly) if we'd like to remove this restriction.

modules/ocf/templates/networking/interfaces.erb
```diff
@@ -16,3 +16,8 @@ iface <%= @iface %> inet6 static
   address <%= @ipaddress6 %>
   netmask <%= @netmask6 %>
   gateway <%= @gateway6 %>
+
+<%- if @lsbdistid == 'Raspbian' -%>
+pre-up wpa_supplicant -B -Dnl80211 -i<%= @iface %> -c /etc/wpa_supplicant/wpa_supplicant.conf -f /var/log/wpa_supplicant.log
```
jvperrin: I should probably include `/etc/wpa_supplicant/wpa_supplicant.conf` in here pulling the password from the private share.

modules/ocf/manifests/networking.pp
```diff
@@ -54,7 +58,7 @@
         # necessarily correct, falling back will help make sure VMs upgraded
         # from jessie have networking configured when they reboot, since they
         # retain the old 'eth0' interface name until after reboot.
-        # XXX find out how to predict new interface names
+        # XXX: find out how to predict new interface names
```
chriskuehl: this is probably better as a TODO anyway, whereas XXX is just sort of a general "beware, here there be dragons"-type comment

modules/ocf/manifests/packages.pp
chriskuehl: sounds reasonable to me. may not be worth the effort though.

i'm more interested in ocflib than these tbh

modules/ocf/manifests/networking.pp
```diff
@@ -39,8 +39,12 @@
   } else {
     # Find the first network interface that starts with 'en', since those are
     # ethernet interfaces. (Won't work for the raspberry pi, since it uses wifi)
```
kpengboy: Might want to update this comment

modules/ocf/manifests/networking.pp
kpengboy: What about something like

```puppet
$ifaces_array = split($::interfaces, ',')
$wl_ifaces = grep($ifaces_array, 'wl.+')
$br_iface = length($wl_ifaces) > 0 ? {
  true  => $wl_ifaces[0]
  false => grep($ifaces_array, 'en.+')[0]
}

kpengboy: For that matter `br_iface` is a terrible variable name for non-hypervisors

jvperrin: Hmmm, yeah that would be better, I'll do that and maybe rename the variable too if I can think of a better name, good idea.

modules/ocf/manifests/packages.pp
jvperrin: Yeah, same on ocflib, I think that's definitely more important.

modules/ocf/manifests/networking.pp
jvperrin: I think it's still accurate, this won't work for the raspberry pi, since it uses wifi and this section is for ethernet, but it could be clarified.

kpengboy: Or the comment can be moved down into some if block. And then we can add a similar one about the first network interface that starts with 'wl'

kpengboy:
bump

jvperrin:
I'm leaving this for now because the raspberry pi is currently off and I'd like to check things still work on it. I'd also like to make sure we're mirroring the extra apt repo that the raspberry pi uses and address existing comments while being able to test that my changes actually work before merging this.

jvperrin:
Ok, I booted `overheat` again and tested this worked (and made a few small changes), I think this should be ready to merge unless anyone has any other comments.

PR #320, opened 2018-04-22
Use pam_mkhomedir everywhere

### LINKED ISSUE #98: replace ocf::staff_users with pam_mkhomedir
we can probably use a lot of the logic from pam_mkhomedir on desktops

jvperrin:
This'll also be useful for the HPC project too to not have to create home directories for everyone in the `ocfhpc` group when they can just be automatically created on login!

(Fixes #98)


jvperrin:
Hmmm, that's a good point, I did check that almost all the LDAP users have home directories that exist (the `mysql` user doesn't, for example, but all the other LDAP users have home directories that exist, although they are stuff like `/opt/share` or `/` but at least they won't be newly created).

PR #321, opened 2018-04-23
Print notify

kkuehler:
This PR is sorta large. Redis is moved to a new server (hozer-65, for now) and desktops get notifications when a print job fails. This is mostly focused on sending notifications when a user exceeds his or her quota, but some other errors can be added later.

For anyone interested in testing this, trigger a desktop to my environment and use `lpr -H dev-printhost` [blank.pdf](https://unix.stackexchange.com/questions/277892/how-do-i-create-a-blank-pdf-from-the-command-line) to print.

TODOS:
-----------
- ~Rename hozer-65~
- ~Stop using a self-signed cert~ (not relevant to this PR; can be done in puppet private)
- (Maybe?) move some of the pubsub logic to ocflib
- ~Could also add `/etc/sudoers` rule that allows `ocfbroker` to use `notify-send` as `root` to send notifications, though I prefer the current solution because the listener script will automatically kill the `ocfbroker` script on logout.~

Input welcome!


modules/ocf_desktop/files/xsession/print-notify-handler
```diff
@@ -0,0 +1,51 @@
...
+def subscribe(host, password, *channels):
+    rc = redis_connection(host=host, password=password)
+    sub = rc.pubsub(ignore_subscribe_messages=True)
+    sub.subscribe(channel for channel in channels)
+    return sub
+
+
+BROKER_AUTH = '/opt/share/broker/broker.conf'
+
+
```
dkess: Why are there 2 empty lines here?

modules/ocf_desktop/files/xsession/print-notify-listener
```diff
@@ -0,0 +1,27 @@
...
+    exit 0
+}
+
+trap clean_up INT TERM EXIT
+
+pipe=$HOME/tmp/print-notify
+icon=/opt/share/xsession/images/ocf-color-256.png
+
+if [[ ! -p $pipe ]]; then
+    mkfifo $pipe
```
dkess: Perhaps delete an old pipe if it already exists (for example, if a session didn't get cleaned up for some reason).

dkess: You should quote variables (`"$pipe"`)

hieradata/nodes/hozer-65.yaml
```diff
@@ -0,0 +1,2 @@
+classes:
```
abizer: you should probably rename this and promote hozer-65 to a prod vm

modules/ocf/templates/broker/broker.conf.erb
```diff
@@ -0,0 +1,3 @@
+[broker]
+host = hozer-65.ocf.berkeley.edu
```
abizer: likewise don't hardcode these things, add a cname to the new vm for `broker.ocf.berkeley.edu` or something like that

modules/ocf_broker/manifests/init.pp
```diff
@@ -0,0 +1,51 @@
...
+    $redis_password = assert_type(Pattern[/^[a-zA-Z0-9]*$/], hiera('broker::redis::password'))
+
+    augeas { '/etc/redis/redis.conf':
+        lens      => 'Spacevars.simple_lns',
+        incl      => '/etc/redis/redis.conf',
+        changes   => [
+          'set port 6379',
+          "set requirepass ${redis_password}",
+          'set appendonly yes',
+          'rm save',
```
abizer: inconsistent quotation

modules/ocf_desktop/files/xsession/print-notify-handler
```diff
@@ -0,0 +1,51 @@
...
+import redis
+
+
+redis_connection = functools.partial(redis.StrictRedis,
+                                     host='broker',
+                                     port=6378,
+                                     ssl=True,
+                                     db=0,
+                                     password=None)
```
abizer: maybe format this so the arguments start at the next line and one indent back? I think that's a style we use elsewhere

modules/ocf_desktop/files/xsession/print-notify-listener
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash -eu
```
abizer: minor nit, but we tend to do this as
```
#!/bin/bash
set -euo pipefail
```

modules/ocf_printhost/manifests/enforcer.pp
```diff
@@ -1,6 +1,8 @@
 class ocf_printhost::enforcer {
   package { ['cups-tea4cups', 'mariadb-client']: }
 
+  $redis_password = assert_type(Pattern[/^[a-zA-Z0-9]*$/], hiera('broker::redis::password'))
```
abizer: I'd take this opportunity to move enforcer.conf out of the private share and into the regular puppet repo as a template, like we do with ocfweb.conf or the broker.conf.erb in this PR

modules/ocf_broker/manifests/init.pp
jvperrin: I believe to interpolate variables you need double quotes, so I think this is fine.

jvperrin: Yeah, just checked, this is necessary to get this to work. For example, [this commit](https://github.com/ocf/puppet/commit/7491c0c7cfa7a03c639facd03a42c5a9db4a8406).

modules/ocf/templates/broker/broker.conf.erb
jvperrin: I'd just put `broker` here instead of the FQDN too, since that's simpler and a bit cleaner.

modules/ocf_desktop/files/xsession/print-notify-handler
```diff
@@ -0,0 +1,51 @@
...
+
+
+def subscribe(host, password, *channels):
+    rc = redis_connection(host=host, password=password)
+    sub = rc.pubsub(ignore_subscribe_messages=True)
+    sub.subscribe(channel for channel in channels)
+    return sub
+
+
+BROKER_AUTH = '/opt/share/broker/broker.conf'
```
jvperrin: This should go further up in the file, definitely above the `subscribe` method anyway.

modules/ocf_desktop/files/xsession/print-notify-listener
```diff
@@ -0,0 +1,27 @@
...
+function clean_up {
+    jobs -pr | xargs pgrep -P | xargs sudo -u ocfbroker kill
+    exit 0
+}
+
+trap clean_up INT TERM EXIT
+
+pipe=$HOME/tmp/print-notify
```
jvperrin: I think you should put the value here in quotes, not that `$HOME` should have spaces in it, but I guess it might.

modules/ocf_desktop/files/xsession/print-notify-listener
jvperrin: Agreed, variables should be quoted pretty much everywhere in this file.

modules/ocf/templates/broker/broker.conf.erb
kkuehler: I already added a CNAME for it, I just blanked here :P

modules/ocf_broker/manifests/init.pp
abizer: I know, so we may as well use double-quotes everywhere I think, because the one line with doubles and everything else with singles is slightly triggering me lol

modules/ocf_desktop/manifests/printnotify.pp
```diff
@@ -0,0 +1,27 @@
+class ocf_desktop::printnotify {
+  user { 'ocfbroker':
+    ensure => present,
+    shell  => '/bin/false',
+  }
+
+  # enable regular users to run notification script as ocfbroker
+  file { '/etc/sudoers.d/broker':
+    content => "ALL ALL=(ocfbroker) NOPASSWD: /opt/share/puppet/print-notify-handler\n\
+ALL ALL=(ocfbroker) NOPASSWD: /bin/kill\n",
```
jvperrin: At this point I think just making a file for this would be better than putting it inline here.

modules/ocf_desktop/files/xsession/print-notify-listener
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash -eu
+function clean_up {
+    jobs -pr | xargs pgrep -P | xargs sudo -u ocfbroker kill
```
jvperrin: Why is this necessary?

modules/ocf_printhost/files/enforcer
```diff
@@ -147,18 +149,26 @@ def send_printer_mail(message, job, quo):
     send_mail_user(job.user, message.subject, body)
 
 
-def prehook(c, job):
+def prehook(c, r, job):
     quo = quota.get_quota(c, job.user)
     if job.pages > quo.daily:
         send_printer_mail(INSUFFICIENT_QUOTA_MESSAGE, job, quo)
+        msg = job.user + '\nYour print job failed due to insufficient pages.  Your job was ' \
+            + job.pages + 'pages and you have ' + quo.daily + ' pages remaining today.'
```
jvperrin: I'd use format strings for this instead, I think that would make this quite a bit cleaner.

modules/ocf_printhost/files/enforcer
```diff
@@ -147,18 +149,26 @@ def send_printer_mail(message, job, quo):
...
+        r.publish(job.user, msg)
         sys.exit(255)
 
 
-def posthook(c, job, success):
+def posthook(c, r, job, success):
     if success:
         quota.add_job(c, job)
+        msg = job.user + "\nYour print job '" + job.doc_name + "' was accepted and sent to " + job.printer + '.'
+        r.publish(job.user, msg)
```
jvperrin: Same here on using format strings

modules/ocf_broker/manifests/init.pp
jvperrin: Eh, I'd just leave it as is, I think having single quotes except when needing double quotes makes more sense.

modules/ocf_desktop/files/xsession/print-notify-listener
kkuehler: The child process is running as ocfbroker, so it will not automatically be killed on logout.

kkuehler:
@jvperrin @dkess @abizer Okay guys. I think I made all the suggested changes. Let me know.

modules/ocf_printhost/files/enforcer
```diff
@@ -25,14 +25,14 @@ from textwrap import dedent
 from traceback import format_exc
 
 import ocflib.printing.quota as quota
+import redis
```
abizer: I have a feeling pre-commit won't like this unordered import

kpengboy: Well, seems like the commits have been passing the CI tests?

hieradata/nodes/hozer-65.yaml
abizer: @kkuehler don't forget to make this a proper VM before merging

kpengboy: Also when you make it a proper VM please make sure to change its IP to be something less than 65; I'd like the start of the hozer range to stay as 65.

modules/ocf_desktop/files/xsession/print-notify-listener
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -euo pipefail
+
+function clean_up {
+    jobs -pr | xargs pgrep -P | xargs sudo -u ocfbroker kill
```
jvperrin: This seems kinda complicated, why does it need to be composed like this?

modules/ocf_desktop/files/xsession/print-notify-listener
```diff
@@ -0,0 +1,28 @@
...
+    exit 0
+}
+
+trap clean_up INT TERM EXIT
+
+pipe="$HOME/tmp/print-notify"
```
jvperrin: I think still putting this as a dot file or hiding it a bit more would make more sense, otherwise I could see (a few) users getting confused by finding it and then potentially deleting it.

modules/ocf_desktop/files/xsession/print-notify-listener
```diff
@@ -0,0 +1,28 @@
...
+icon=/opt/share/xsession/images/ocf-color-256.png
+
+[[ -p "$pipe" ]] && rm "$pipe"
+mkfifo $pipe
+
+sudo -u ocfbroker /opt/share/puppet/print-notify-handler > "$pipe" &
+
+# listener
+while true
```
jvperrin: I'm assuming this doesn't busy-wait and use extra CPU by being in a loop with no pauses?

modules/ocf_desktop/files/xsession/print-notify-handler
```diff
@@ -0,0 +1,53 @@
...
+    host = config.get('broker', 'host')
+    password = config.get('broker', 'password')
+    return host, password
+
+
+def main():
+    username = os.environ.get('SUDO_USER')
+
+    if not username:
+        raise RuntimeError('Unable to read SUDO_USER.')
```
jvperrin: I'd report this [via ocflib](https://github.com/ocf/ocflib/blob/292f0d835d20f55512380945ecacd73f0b25a5fd/ocflib/misc/mail.py#L88) since otherwise I don't think we'll ever see it.

modules/ocf_desktop/templates/sudoers.d/broker
```diff
@@ -0,0 +1,2 @@
+ALL ALL=(ocfbroker) NOPASSWD: /opt/share/puppet/print-notify-handler
+ALL ALL=(ocfbroker) NOPASSWD: /bin/kill
```
jvperrin: I'd definitely feel more comfortable if the `ocfbroker` user didn't have the ability to kill any process. Could the handler script check some file or something to terminate instead?

modules/ocf_printhost/files/enforcer
```diff
@@ -103,13 +103,26 @@ ENFORCER_ERROR_MESSAGE = Message(
         """) + MAIL_SIGNATURE
 )
 
+NOTIFY_QUOTA_MESSAGE = dedent("""\
+        Your print job failed due to insufficient pages. Your job was
+        {pages} pages, and you have {quota} pages remaining today.""")
```
jvperrin: I'd put the ending parentheses on the next line, since otherwise the indentation looks like it doesn't have an end. Like this:

```
NOTIFY_QUOTA_MESSAGE = dedent("""\
        Your print job failed due to insufficient pages. Your job was
        {pages} pages, and you have {quota} pages remaining today."""
)
```

Same with the other two messages

modules/ocf_printhost/files/enforcer
```diff
@@ -147,19 +160,29 @@ def send_printer_mail(message, job, quo):
...
 
 
-def posthook(c, job, success):
+def posthook(c, r, job, success):
+    msg = ''
     if success:
         quota.add_job(c, job)
+        msg = NOTIFY_JOB_ACCEPTED.format(
+            document=job.doc_name,
+            printer=job.printer)
```
jvperrin: For this (and the one above this), I'd format it like this:

```
msg = NOTIFY_JOB_ACCEPTED.format(
    document=job.doc_name,
    printer=job.printer,
)
```

Then it's clear where the statement ends and you get a trailing comma!

modules/ocf_printhost/files/enforcer
```diff
@@ -147,19 +160,29 @@ def send_printer_mail(message, job, quo):
...
     else:
         quo = quota.get_quota(c, job.user)
+        msg = NOTIFY_JOB_ERROR
```
jvperrin: Could the document name be put here too like in the success message?

modules/ocf_printhost/manifests/enforcer.pp
```diff
@@ -19,7 +23,10 @@
       mode   => '0500';
 
     '/opt/share/enforcer/enforcer.conf':
-      source    => 'puppet:///private/enforcer/enforcer.conf',
+      content   => template(
+        'ocf_printhost/enforcer/enforcer.conf.erb',
+        'ocf/broker/broker.conf.erb',
+      ),
```
jvperrin: Nice, this is much better, and I like how you've done the combination of the two here so you can use the broker config twice.

modules/ocf_printhost/manifests/enforcer.pp
```diff
@@ -1,6 +1,10 @@
 class ocf_printhost::enforcer {
   package { ['cups-tea4cups', 'mariadb-client']: }
 
+  $mysql_password = assert_type(Pattern[/^[a-zA-Z0-9]*$/], hiera('ocfprinting::mysql::password'))
+
```
jvperrin: I'd just remove this newline, both lines I'd say are related (and similar) enough to be together

modules/ocf_desktop/files/xsession/print-notify-listener
kkuehler: Putting a `sleep 1` at the end of each iteration is probably a good idea.

jvperrin: I mean if this doesn't use extra CPU, then you don't need to do that, but I just wanted to check if it did or not when you were testing it

modules/ocf_desktop/files/xsession/print-notify-listener
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -euo pipefail
+
+function clean_up {
+    jobs -pr | xargs pgrep -P | xargs sudo -u ocfbroker kill
+    exit 0
+}
+
+trap clean_up INT TERM EXIT
```
kpengboy: INT and TERM are redundant since you already specified EXIT and this script is being run by bash

modules/ocf_desktop/files/xsession/print-notify-listener
```diff
@@ -0,0 +1,28 @@
...
+icon=/opt/share/xsession/images/ocf-color-256.png
+
+[[ -p "$pipe" ]] && rm "$pipe"
+mkfifo $pipe
+
+sudo -u ocfbroker /opt/share/puppet/print-notify-handler > "$pipe" &
+
+# listener
+while true
+do
```
kpengboy: Interesting you put the `then` on the same line but don't do the same for `do`...

modules/ocf_desktop/files/xsession/print-notify-listener
kkuehler: nah it was using 0%

modules/ocf_desktop/files/xsession/print-notify-listener
```diff
@@ -0,0 +1,28 @@
...
+    exit 0
+}
+
+trap clean_up INT TERM EXIT
+
+pipe="$HOME/tmp/print-notify"
+icon=/opt/share/xsession/images/ocf-color-256.png
+
+[[ -p "$pipe" ]] && rm "$pipe"
+mkfifo $pipe
```
kpengboy: Does this pass shellcheck? If not, then shouldn't we hook shellcheck into this repo...

modules/ocf_desktop/templates/sudoers.d/broker
kpengboy: ...This gives any user the ability to kill processes running as ocfbroker.
Still, it'd be better to narrow this ability...

modules/ocf_desktop/files/xsession/print-notify-listener
kpengboy: I think `/run/user/$UID/` is the appropriate place for these kinds of fifos and sockets.

modules/ocf_desktop/files/xsession/print-notify-listener
```diff
@@ -0,0 +1,28 @@
...
+    jobs -pr | xargs pgrep -P | xargs sudo -u ocfbroker kill
+    exit 0
+}
+
+trap clean_up INT TERM EXIT
+
+pipe="$HOME/tmp/print-notify"
+icon=/opt/share/xsession/images/ocf-color-256.png
+
+[[ -p "$pipe" ]] && rm "$pipe"
```
kpengboy: Typically one uses `rm -f`

modules/ocf_desktop/templates/sudoers.d/broker
kkuehler: Other way around. Users can kill as ocfbroker. That said, we could have the handler process clean up and die on BrokenPipeError which would remove the need for all this. We would just need the handler to send 'alive' messages to actually receive that error without getting a print notification. This also wouldn't work if I add a sleep to the while true.

modules/ocf_desktop/files/xsession/print-notify-listener
kkuehler: Yeah, whoops. Missed another :(

modules/ocf_desktop/files/xsession/print-notify-listener
kpengboy: There is a way to keep the pipe open at all times. Before the while loop write (on a solitary line):

```bash
3< "$pipe"
```

This opens up the pipe as FD 3 in the shell script. Then in the loop below, you can replace the read statement with:

```bash
read -u 3 msg
```

kpengboy: The `read` statement is blocking so this shouldn't busy-loop.

modules/ocf_desktop/files/xsession/print-notify-listener
kpengboy: Actually, even better: have the handler [detect when the pipe is closed](https://stackoverflow.com/questions/28315446/detect-when-reader-closes-named-pipe-fifo)

modules/ocf_desktop/files/xsession/print-notify-listener
kkuehler: Wow, great info. nit: before the loop it needs to be
```bash
exec 3< "$pipe" 
```
I think that's probably the best way to do this. Then we don't need to give users the perm to kill as `ocfbroker` (a one liner means no need for an extra puppet template, so +1 that), and no `xargs` nonsense on cleanup. The `cleanup` method could actually be reduced to a much more intuitive one-liner:
```bash
exec 3<&- && rm -f "$pipe"
```
This will cause the python script to die when it tries to print. That means we'll have to change the loop to a `while True` to send periodic messages (the `redis` library actually has that in their demo code, anyway, not `listen()`). This would also be the case if we were checking for the existence of a file, anyway.

modules/ocf_desktop/files/xsession/print-notify-handler
```diff
@@ -0,0 +1,53 @@
...
+    ssl=True,
+    db=0,
+    password=None
+)
+
+
+def subscribe(host, password, *channels):
+    rc = redis_connection(host=host, password=password)
+    sub = rc.pubsub(ignore_subscribe_messages=True)
+    sub.subscribe(channel for channel in channels)
```
kpengboy: Um...we can't just write `sub.subscribe(channels)`?

modules/ocf_desktop/files/xsession/print-notify-listener
kkuehler: See below in my response to your using fd 3. I came to the same conclusion. I also think a thread is a bit overkill for code as complicated as this already is.

modules/ocf_desktop/files/xsession/print-notify-listener
kpengboy: > nit: before the loop it needs to be:

Derp lol

> `rm -f "$pipe" && exec 3<&- && exit`

Well, you wouldn't need the `exit` since it's already going to exit. The `exec 3<&-` might not be required either since the FD would get closed on program exit anyway.

modules/ocf_desktop/templates/sudoers.d/broker
kpengboy: Also, as a general note, files that don't use Puppet/ERB template features should go in `files` rather than `templates`. I figure you probably know this already, but would just like to remind you...

modules/ocf_desktop/files/xsession/print-notify-listener
kpengboy: > to a `while True` to send periodic messages

I don't think you would necessarily need to send periodic messages, something like

```
while True:
    message = p.get_message()
    if message:
        write it to the pipe
    poll stdout for a POLLERR event with timeout 0.1 or whatever
    if poll returns a nonempty list:
        break
```

should work.

Or if we're adventurous we could find which FD is the redis socket and directly pass that and stdout to `poll`. That's probably fragile though.

modules/ocf_desktop/files/xsession/print-notify-listener
```diff
@@ -0,0 +1,23 @@
...
+
+sudo -u ocfbroker /opt/share/puppet/print-notify-handler > "$pipe" &
+
+# listener
+exec 3< "$pipe"
+while true; do
+    if [[ ! -p "$pipe" ]]; then
+        break
+    fi
+    if read -u 3 msg < "$pipe"; then
```
kpengboy: I don't think the `< "$pipe"` is actually necessary here?

keur: ¯\_(ツ)_/¯
(ﾉ◕ヮ◕)ﾉ*:･ﾟ✧
(ﾉ◕ヮ◕)ﾉ*:･ﾟ✧
(ﾉ◕ヮ◕)ﾉ*:･ﾟ✧
(ﾉ◕ヮ◕)ﾉ*:･ﾟ✧
(ﾉ◕ヮ◕)ﾉ*:･ﾟ✧
(ﾉ◕ヮ◕)ﾉ*:･ﾟ✧

PR #325, opened 2018-05-14
ocf_desktop: purge ayatana-indicator-application

kpengboy:


kpengboy:
Do you all think I should explicitly install remmina without any dependencies or is what I have fine

dkess:
Can you give more context on what this is? What do we use it for?

kpengboy:
Installing remmina pulls in ayatana-indicator-application as an indirect recommended dependency.

Without ayatana-indicator-application:

Without ayatana-indicator-application

kpengboy:
With ayatana-indicator-application:

With ayatana-indicator-application

modules/ocf_desktop/manifests/packages.pp
```diff
@@ -53,12 +53,15 @@
 
   # Remove some packages
   package {
+    # causes the fcitx menu to vanish
```
jvperrin: Maybe say what the fcitx menu does for those who don't know it is for switching keyboard layouts?

PR #326, opened 2018-05-26
Get death working with stretch

jvperrin:
This PR took quite a while to get (finally) working, but it moves us from:

- fastcgi (unmaintained) + suexec for vhosts
- suphp (also unmaintained) for userdirs

to fcgid + suexec for both vhosts and userdirs, both of which are maintained for the forseeable future. We do have to add a [patch on top of suexec](https://github.com/ocf/apache2-suexec-ocf/pull/1) to get this to work, but we were already doing that anyway, so this doesn't add much more complexity to our setup.

Alternatives considered:
- [mpm-itk](http://mpm-itk.sesse.net): Tried it and didn't get it working well, would need to use mod_php too
- [php-fpm](https://secure.php.net/manual/en/install.fpm.php): Got this working pretty well, but would need to rebuild the configuration for new users, it didn't work well with 50k+ users, and only really started working reasonably when narrowed down to the ~2k users that actually use php. We might want to consider this in the future though, and I've pushed my work on it so far to a [separate branch](https://github.com/ocf/puppet/tree/php-fpm-pools) to keep it around.
- Foward port [suphp](https://www.suphp.org/Home.html): Didn't really want to do this since we've done this already and eventually we just have to become the maintainers for the whole package.

Things to do:
- [x] Make sure that the lack of an auto-updating feature with mod_fcgid doesn't make it too unusable and update ocfweb docs to reflect the lack of an autoupdate.
- [ ] Add more test cases to the [ggroup account](https://www.ocf.berkeley.edu/~ggroup/)
- [x] Merge and build the [modified suexec package](https://github.com/ocf/apache2-suexec-ocf/pull/1) (probably do it automatically with Jenkins for consistency?)
- [x] Test speed/reliability vs current death to make sure that dev-death can handle the same traffic that death does without the errors that death is getting currently from mod_fastcgi.
- [x] Test this on jessie with dev-death running jessie to see if it works there. I don't see why it shouldn't work, but I'd like to test it for a bit there first to see.
- [x] Remake dev-death running stretch and check that everything puppets correctly, since I have been messing around with this host quite a bit so I might have some unpuppeted changes.

Thoughts?


modules/ocf_www/files/vhost-web.jinja
```diff
@@ -36,16 +36,14 @@
             Options +ExecCGI
             AllowOverride None
             Require all granted
-            SetHandler fastcgi-script
+            SetHandler fcgid-script
         </Directory>
 
-        <FilesMatch "\.ph(p3?|tml)$">
+        <FilesMatch "\.ph(p[3457]?|tml)$">
```
chriskuehl: why is six afraid of seven? because seven eight nine

this joke doesn't really fit but i couldn't think of a better one to comment on the lack of a php 6

:partycat:

modules/ocf_www/manifests/mod/fcgid.pp
```diff
@@ -0,0 +1,11 @@
+class ocf_www::mod::fcgid {
```
chriskuehl: do we have any way for users to restart their fastcgi processes?

it's probably not essential/blocking, but i suspect people will complain about the lack of it eventually

jvperrin: I don't think there is a way currently, I'd like to play around with that a bit more and potentially set CGI execution time to a lower value to make the effects of this less annoying for people working on their websites live. (One of my TODOs in the PR description)

Were there many complaints before #3 was merged?

modules/ocf_www/files/vhost-web.jinja
jvperrin: Haha, that actually fits pretty well, let's just see if there will be a PHP 9 or not, maybe it was eaten? ;)

:partycat: actually shows up on my phone but not on desktop, thanks GitHub :(

jvperrin:
Lol thanks, I would run for SM, but I heard the time commitment is pretty high :P

modules/ocf_www/manifests/mod/fcgid.pp
chriskuehl: Yeah. I can't find them in my inbox, but that's why we switched.

chriskuehl: Probably less of a concern now with apphosting -- though it will still affect individuals /shrug

jvperrin: Yeah, I would like to make this not a problem for users if I can. I really wish that fastcgi was still maintained or that fcgid had a similar autoupdate feature, I've found a few threads online where people were asking for one but I couldn't find anything useful there. I'm hoping that lowering the time that scripts run by fcgid are alive for will help with this so that the docs can just say something like "wait for 30 seconds" or something would work decently well and still yield acceptable performance. After all, most websites aren't that popular (except dev-dev-vhost from health checks), so that helps.

PR #329, opened 2018-06-10
i3 sucks

keur:
Making the i3 setup on the TV (slightly) more usable.

If anyone is interested, I am happy to make the status bar a little less jank
(for example, removing the TV's "battery status").

### DIFF
--- a/modules/ocf_tv/files/i3/config
+++ b/modules/ocf_tv/files/i3/config
@@ -172,6 +172,9 @@ mode "volume" {
 bindsym $mod+Shift+v mode "volume"
 bindsym $mod+v exec --no-startup-id pavucontrol
 
+# htop
+bindsym $mod+i exec --no-startup-id $term -e htop
+
 # Web Browser
 bindsym $mod+w exec --no-startup-id google-chrome-stable
 bindsym $mod+Shift+w exec --no-startup-id $browseralt
--- a/modules/ocf_tv/files/i3/config
+++ b/modules/ocf_tv/files/i3/config
@@ -39,8 +39,8 @@ bindsym $mod+Return exec $term
 # kill focused window
 bindsym $mod+Shift+q kill
 
-# start dmenu (a program launcher)
-bindsym $mod+d exec dmenu_run
+# start rofi (a program launcher)
+bindsym $mod+d exec rofi -show run
 # There also is the (new) i3-dmenu-desktop which only displays applications
 # shipping a .desktop file. It is a wrapper around dmenu, so you need that
 # installed.
--- a/modules/ocf_tv/files/i3/config
+++ b/modules/ocf_tv/files/i3/config
@@ -11,7 +11,7 @@
 
 exec feh --bg-scale '/opt/share/background.png'
 
-set $mod Mod1
+set $mod Mod4
 
 # Font for window titles. Will also be used by the bar unless a different font
 # is used in the bar {} block below.
--- a/modules/ocf_tv/files/i3/config
+++ b/modules/ocf_tv/files/i3/config
@@ -12,6 +12,8 @@
 
 set $mod        Mod4
 set $term       i3-sensible-terminal
+set $browser    google-chrome-stable
+set $browseralt firefox
 
 # Run on startup
 exec_always feh --no-startup-id --bg-scale '/opt/share/background.png'
@@ -150,6 +152,10 @@ mode "resize" {
 }
 
 bindsym $mod+r mode "resize"
+
+# Web Browser
+bindsym $mod+w exec --no-startup-id google-chrome-stable
+bindsym $mod+Shift+w exec --no-startup-id $browseralt
 
 # Start i3bar to display a workspace bar (plus the system information i3status
 # finds out, if available)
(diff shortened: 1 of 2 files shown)


modules/ocf_tv/files/i3/config
```diff
@@ -9,9 +9,14 @@
 #
 # Please see http://i3wm.org/docs/userguide.html for a complete reference!
 
-exec feh --bg-scale '/opt/share/background.png'
 
-set $mod Mod1
+set $mod        Mod4
```
dkess: This could be a bad idea, since we don't want the super key to conflict from inside vncviewer

modules/ocf_tv/files/i3/config
```diff
@@ -29,13 +34,13 @@ font pango:DejaVu Sans Mono 8
...
-bindsym $mod+Return exec i3-sensible-terminal
+bindsym $mod+Return exec $term
 
 # kill focused window
 bindsym $mod+Shift+q kill
 
-# start dmenu (a program launcher)
-bindsym $mod+d exec dmenu_run
+# start rofi (a program launcher)
+bindsym $mod+d exec rofi -show run
```
dkess: If `drun` is available we should probably use it, since that gives more sensible program names.

modules/ocf_tv/files/i3/config
```diff
@@ -69,14 +74,14 @@ bindsym $mod+Shift+Right move right
 bindsym $mod+semicolon split h
 
 # split in vertical orientation
-bindsym $mod+v split v
+bindsym $mod+slash split v
```
dkess: What is the reason for making this change?

modules/ocf_tv/files/i3/config
```diff
@@ -69,14 +74,14 @@ bindsym $mod+Shift+Right move right
...
 
 # enter fullscreen mode for the focused container
 bindsym $mod+f fullscreen
 
 # change container layout (stacked, tabbed, toggle split)
 bindsym $mod+s layout stacking
-bindsym $mod+w layout tabbed
+bindsym $mod+t layout tabbed
```
dkess: Again, why move away from the default here?

modules/ocf_tv/files/i3/config
```diff
@@ -148,6 +153,32 @@ mode "resize" {
...
+}
+
+bindsym $mod+Shift+v mode "volume"
+bindsym $mod+v exec --no-startup-id pavucontrol
+
+# htop
+bindsym $mod+i exec --no-startup-id $term -e htop
+
+# Web Browser
+bindsym $mod+w exec --no-startup-id google-chrome-stable
```
dkess: Should be `$browser`  (if we end up doing this at all)

modules/ocf_tv/files/i3/config
abizer: I'd hesitate from changing some of the core keybindings away from the defaults, it will interfere a lot with people's muscle memory.

modules/ocf_tv/files/i3/config
abizer: same as above

modules/ocf_tv/files/i3/config
```diff
@@ -148,6 +153,32 @@ mode "resize" {
...
+bindsym $mod+Shift+w exec --no-startup-id $browseralt
```
dkess: I'm not opposed to the "volume mode," (since it could make adjusting volume easy while on vncviewer) but are these shortcuts necessary when we have the launcher?

modules/ocf_tv/files/i3/config
kkuehler: ; and v together make literally no sense.

modules/ocf_tv/files/i3/config
kkuehler: I want w for web, and because mod+w for tabbed layout makes no mnemonic sense.

modules/ocf_tv/files/i3/config
kkuehler: Yeah, whoops I set the variable and forgot that.

modules/ocf_tv/files/i3/config
```diff
@@ -148,6 +153,32 @@ mode "resize" {
 
 bindsym $mod+r mode "resize"
 
+# Volume Control
```
abizer: it may be better to do this the same way `ocf-tv` does it? https://github.com/ocf/utils/blob/87cbb68404c72bd21ce0b908e4b71303f1a34909/staff/lab/ocf-tv#L117

modules/ocf_tv/files/i3/config
kkuehler: They are much faster. Kinda the entire point of i3.

modules/ocf_tv/files/i3/config
kkuehler: mod+semicolon is not a default and was already changed. mod+semicolon together with mod+v has to be the stupidest thing I've ever seen.

abizer: mod+semicolon is there because that's what was in nickimp's config when he shipped this. I'd advocate removing it and returning to the defaults.

modules/ocf_tv/files/i3/config
kkuehler: Yeah, it is. Good idea.

keur:
will maybe revisit later. i agree the defaults on the TV are good.

PR #331, opened 2018-06-17
Remove bash startup/logout files in /etc/skel

kpengboy:
Since we moved to creating home directories via `pam_mkhomedir`, the default `.bashrc`, `.profile`, and `.bash_logout` have been getting copied into new home directories. These files unfortunately tend to interfere with what we already set up globally in `/etc/bash.bashrc` etc. (For example, the settings in `.bashrc` overwrite our custom bash prompt.) Thus, we should remove these files.

### DIFF
--- a/modules/ocf/manifests/packages/shell.pp
+++ b/modules/ocf/manifests/packages/shell.pp
@@ -45,6 +45,12 @@ class ocf::packages::shell {
       require => Package['bash'];
   }
 
+  # Settings in these files are handled by our custom bash.bashrc and bash.bash_logout
+  file {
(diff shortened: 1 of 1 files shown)


kpengboy:
@jvperrin Back when we created home directories using Puppet, it didn't copy the contents of `/etc/skel` into those home directories, so there wasn't a problem.

jvperrin:
Yeah, I get that, I was confused where the files in `/etc/skel` were even coming from, but it looks like some are managed by puppet and some are provided by the `bash` package, so that makes more sense why they aren't all found in this repo.

PR #333, opened 2018-07-06
Generate desktop leases periodically (rt#7106)

jvperrin:
If pestilence was rebuilt without access to LDAP (for instance), then it would not have a list of any desktops, and any new desktops were also not added to the list of leases. Same with if a desktop MAC address changes and then needs to be rebuilt, so this fixes those problems by regenerating the desktop leases list periodically.

Potentially could use python sets here instead of sorting the lists, but I think having the output file sorted by hostname also makes it a bit easier to read through if that is ever necessary.

### DIFF
--- a/modules/ocf_dhcp/files/gen-desktop-leases
+++ b/modules/ocf_dhcp/files/gen-desktop-leases
@@ -1,11 +1,24 @@
 #!/usr/bin/env python3
+import subprocess
+
 from ocflib.infra.hosts import hosts_by_filter
+
+DESKTOP_LEASES_FILE = '/etc/dhcp/desktop-leases.conf'
 
 desktops = hosts_by_filter('(type=desktop)')
 
-for desktop in desktops:
-    print('host {} {{ hardware ethernet {}; fixed-address {}; }}'.format(
+with open(DESKTOP_LEASES_FILE, 'r') as f:
+    old_contents = sorted([line.strip() for line in f])
+
+new_contents = []
+for desktop in sorted(desktops, key=lambda d: d['cn'][0]):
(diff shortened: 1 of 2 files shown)


modules/ocf_dhcp/files/gen-desktop-leases
```diff
@@ -1,11 +1,24 @@
...
+    new_contents.append('host {} {{ hardware ethernet {}; fixed-address {}; }}'.format(
         desktop['cn'][0],
         desktop['macAddress'][0],
         desktop['ipHostNumber'][0],
     ))
+
+if new_contents and new_contents != old_contents:
+    with open(DESKTOP_LEASES_FILE, 'w') as f:
+        f.write('\n'.join(new_contents) + '\n')
+    subprocess.check_call(('systemctl', 'restart', 'isc-dhcp-server'))
```
kkuehler: What? [You're telling me you don't want to call systemd through DBus?](https://github.com/systemd/systemd/blob/8255430754d3836b5f4713733341d6e2e8bd8b98/src/core/dbus-unit.c#L644)

Kidding.

jvperrin: o boi

kpengboy: Hey, it's [only a few lines of code](https://stackoverflow.com/questions/33646374/starting-a-systemd-service-via-python) :P

```python
import dbus
sysbus = dbus.SystemBus()
systemd1 = sysbus.get_object('org.freedesktop.systemd1', '/org/freedesktop/systemd1')
manager = dbus.Interface(systemd1, 'org.freedesktop.systemd1.Manager')
job = manager.RestartUnit('isc-dhcp-server.service', 'fail')
```

PR #334, opened 2018-07-06
ocf_staffvm: staffvm must have owner in hiera

### LINKED ISSUE #298: /etc/sudoers is broken if staffvm and no owner
The `/etc/sudoers` file becomes broken if the hiera file to add an owner is not added quickly enough, since it contains the following line (notice the single space at the front):

```
 ALL=(ALL) NOPASSWD: ALL
```

This appears to be because the [hiera lookup for an owner](https://github.com/ocf/puppet/blob/c4a9b1c6b9041297fa64ac418bb53b2cec99fcc2/hieradata/type/staffvm.yaml#L3) returns an empty string in the `usudo` array, leading to a line being added in the [`/etc/sudoers` template](https://github.com/ocf/puppet/blob/master/modules/ocf/templates/sudoers.erb), but not one with an actual user. Probably should just filter out empty or invalid users in the [`auth.pp` manifest](https://github.com/ocf/puppet/blob/master/modules/ocf/manifests/auth.pp)?

kkuehler:
Fixes #298

### DIFF
--- a/modules/ocf_staffvm/manifests/init.pp
+++ b/modules/ocf_staffvm/manifests/init.pp
@@ -1,4 +1,9 @@
 class ocf_staffvm {
+  $owner = lookup('owner', { 'default_value' => undef })
+  if $owner == undef {
+    fail('A staffvm must have an owner defined in hiera!')
+  }
+
   include ocf_staffvm::docker_group
   include ocf_staffvm::firewall
 }


modules/ocf_staffvm/manifests/init.pp
```diff
@@ -1,4 +1,9 @@
 class ocf_staffvm {
   include ocf_staffvm::docker_group
   include ocf_staffvm::firewall
+
+  $owner = lookup('owner', { 'default_value' => undef })
```
jvperrin: I don't think you necessarily need to provide a default value here, if this lookup fails, then puppet will fail anyways:

> \<DEFAULT VALUE\> [...] Defaults to no value; if Hiera can’t find a normal value, the lookup (and compilation) will fail.

Having the customized error message does help to make this a bit nicer though, so that does make this approach better on that front.

kkuehler: Yeah, I'm aware a lookup fails. But just putting a lookup in there that does not make it readily apparent what is going on.

jvperrin: I mean you could put a comment explaining it if you want, and I think that would be pretty readable, but I think the real advantage of doing it the way you have here is the customizable error message, so let's just stick with this.

kpengboy: @kkuehler If you want this custom error message you'll need to put the two `include` lines below the owner check. Otherwise Puppet will error out in the `lookup('owner')` statement in the `ocf_staffvm::docker_group` class instead.

jvperrin: That's true, it'll only error if it's tagged to install docker, but it should still go before that include so that the custom error takes priority.

kkuehler:
@kpengboy ah, good catch!

PR #335, opened 2018-07-09
ocf_stats: set docroot owner and group for munin

kpengboy:
By default the Puppet apache module creates the docroot and sets
its ownership to root:root. This prevents munin from subsequently
copying files into the docroot upon its first run, because the
permissions do not allow for it.

Set the owner and group correctly to fix this.


kpengboy:
Was there some reason we weren't already doing this?

kpengboy:
Closed in favor of #336

PR #336, opened 2018-07-09
Modernize munin code

kpengboy:
Fix some bugs in bootstrapping Munin server, remove unneeded Apache configuration options, and fix some issues with use of server CNAME

### DIFF
--- a/modules/ocf_stats/manifests/init.pp
+++ b/modules/ocf_stats/manifests/init.pp
@@ -5,7 +5,6 @@ class ocf_stats {
     default_vhost => false,
   }
 
-  include apache
   include ocf_stats::labstats
   include ocf_stats::munin
 }
--- a/modules/ocf_stats/manifests/init.pp
+++ b/modules/ocf_stats/manifests/init.pp
@@ -3,9 +3,7 @@ class ocf_stats {
 
   class { '::apache':
     default_vhost => false,
-    mpm_module    => 'prefork'
   }
-  include '::apache::mod::php'
 
   include apache
   include ocf_stats::labstats
--- a/modules/ocf_stats/manifests/munin.pp
+++ b/modules/ocf_stats/manifests/munin.pp
@@ -1,7 +1,6 @@
 # munin master config
 class ocf_stats::munin {
   include ocf::firewall::allow_web
-  include ocf_ssl::default_bundle
 
   package {
     ['munin', 'nmap']:;
--- a/modules/ocf_stats/manifests/munin.pp
+++ b/modules/ocf_stats/manifests/munin.pp
@@ -48,6 +48,8 @@ class ocf_stats::munin {
       servername    => "${cname}.ocf.berkeley.edu",
       port          => 443,
       docroot       => '/var/cache/munin/www/static/',
+      docroot_owner => 'munin',
+      docroot_group => 'munin',
 
       ssl           => true,
       ssl_key       => "/etc/ssl/private/${::fqdn}.key",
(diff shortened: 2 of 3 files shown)


modules/ocf_stats/manifests/munin.pp
```diff
@@ -38,16 +43,13 @@
...
-    'prod' => 'munin',
-  }
-
   apache::vhost {
     'munin':
       servername    => "${cname}.ocf.berkeley.edu",
       port          => 443,
       docroot       => '/var/cache/munin/www/static/',
+      docroot_owner => 'munin',
+      docroot_group => 'munin',
```
kpengboy: Is there a reason these aren't currently set?

kpengboy:
Also what are people's thoughts on EPP vs ERB?

modules/ocf_stats/manifests/munin.pp
```diff
@@ -1,7 +1,6 @@
 # munin master config
 class ocf_stats::munin {
   include ocf::firewall::allow_web
-  include ocf_ssl::default_bundle
```
jvperrin: Isn't this still needed? I don't see how else the certs are getting provided (I'll be changing this soon to use Let's Encrypt though, but that's besides the point)

modules/ocf_stats/templates/munin/mail-munin-alert.epp
```diff
@@ -107,3 +112,9 @@ def main(argv):
 
 if __name__ == '__main__':
     main(argv)
+
+
+# Local Variables:
+# mode: python
+# End:
+# vim: ft=python
```
jvperrin: Is all of this needed? Isn't there a one-line addition that's usually just used at the bottom of files instead of all of this?

modules/ocf_stats/manifests/init.pp
```diff
@@ -3,9 +3,7 @@
 
   class { '::apache':
     default_vhost => false,
-    mpm_module    => 'prefork'
```
jvperrin: Nice, this is good to remove, the event mpm should work fine with this I imagine

modules/ocf_stats/manifests/munin.pp
```diff
@@ -97,7 +98,9 @@
           sethandler   => 'fcgid-script',
           auth_require => 'all granted',
         }
-      ];
+      ],
+
```
jvperrin: Personally I'd remove this extra line so that it's clearer what section the require belongs to.

modules/ocf_stats/manifests/munin.pp
jvperrin: I don't believe so? Not sure how it's currently set up though.

modules/ocf_stats/manifests/munin.pp
kpengboy: It's already included in the main `ocf_stats` class.

modules/ocf_stats/templates/munin/mail-munin-alert.epp
kpengboy: The first three are for emacs, the last one for vim

modules/ocf_stats/manifests/munin.pp
kpengboy: I put the extra line because we're arranging attributes by category here and I thought `require` and `directories` don't fit into one category. Thoughts?

modules/ocf_stats/manifests/munin.pp
kpengboy: @chriskuehl ^

PR #339, opened 2018-07-22
Move prometheus daemon to dementors and configure vhost

dkess:
Does what is says on the tin.  This has been tested on dev-dementors.

### DIFF
--- a/modules/ocf_stats/manifests/prometheus.pp
+++ b/modules/ocf_stats/manifests/prometheus.pp
@@ -1,5 +1,8 @@
 # prometheus daemon config
 class ocf_stats::prometheus {
+  include apache::mod::proxy
+  include apache::mod::proxy_http
+
   file {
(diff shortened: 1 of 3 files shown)


modules/ocf_stats/manifests/prometheus.pp
```diff
@@ -47,4 +50,39 @@
...
+
+      headers             => ['always set Strict-Transport-Security max-age=31536000'],
+      proxy_preserve_host => true,
+      request_headers     => ['set X-Forwarded-Proto https'],
+
+      rewrites            => [
+        {rewrite_rule => '^/(.*)$ http://127.0.0.1:9090/$1 [P]'},
+      ];
+
+    'prometheus-redirect':
```
kkuehler: Would recommend calling this `prometheus-http-redirect` instead.

PR #340, opened 2018-07-25
Add grafana to dementors

dkess:
This PR adds Grafana. Some items of note:

* I haven't configured signin, since I think just letting everyone be anonymous should be good enough. If we decide we want some kind of LDAP integration, we can work on that later.
* Specifying the Grafana package and version number is silly, but the way they install the package is broken, so we have to do it this way.
* The `/grafana` HTTP subdir is defined in `prometheus.pp`, which is a bit wonky since I'd like it to be in `grafana.pp`. Suggestions for how to handle this better are welcome.

### DIFF
--- a/modules/ocf/manifests/packages/grafana/apt.pp
+++ b/modules/ocf/manifests/packages/grafana/apt.pp
@@ -1,0 +1,13 @@
+class ocf::packages::grafana::apt {
+  apt::key { 'grafana':
+    id     => '418A7F2FB0E1E6E7EABF6FE8C2E73424D59097AB',
+    source => 'https://packagecloud.io/gpg.key',
+  }
+
+  apt::source { 'grafana':
+    location => 'https://packagecloud.io/grafana/stable/debian/',
+    release  => 'stretch',
+    repos    => 'main',
+    require  => Apt::Key['grafana'],
+  }
+}
--- a/modules/ocf_stats/manifests/grafana.pp
+++ b/modules/ocf_stats/manifests/grafana.pp
@@ -1,0 +1,62 @@
+# grafana config
+class ocf_stats::grafana {
+  class { 'ocf::packages::grafana::apt':
+    stage => first,
+  }
+
+  $canonical_url = $::host_env ? {
+    'dev'  => 'dev-prometheus',
+    'prod' => 'prometheus',
+  }
+
+  class { 'grafana':
+    install_method           => 'repo',
+
+    # Unfortunately, puppet-grafana requires that we specify a version number
+    # here.
+    version                  => '5.2.1',
+    manage_package_repo      => false,
+
+    cfg                      => {
+      app_mode         => 'production',
+      server           => {
+        http_addr => '127.0.0.1',
+        http_port => 8990,
+        domain    => "${canonical_url}.ocf.berkeley.edu",
+        root_url  => "https://${canonical_url}.ocf.berkeley.edu/grafana/",
+      },
+      database         => {
+        type => 'sqlite3',
+        path => 'grafana.db',
+      },
+      users            => {
+        auto_assign_org      => true,
+        allow_sign_up        => false,
+        auto_assign_org_role => 'Editor',
+      },
+      'auth.anonymous' => {
+        enabled  => true,
+      },
+      security         => {
+        disable_gravatar => true,
+      },
+    },
+
+    provisioning_datasources => {
+      apiVersion  => 1,
+      datasources => [
+        {
+          name      => 'Prometheus',
+          type      => 'prometheus',
+          access    => 'direct',
+          url       => 'https://prometheus.ocf.berkeley.edu/',
+          isDefault => true,
+        },
+      ],
+    },
+
+    # TODO: it's possible to save dashboards in Puppet with
+    # provisioning_dashboards. Once we settle on dashboards, we should put them
+    # here so they can be saved in VCS.
+  }
+}
--- a/modules/ocf_stats/manifests/grafana.pp
+++ b/modules/ocf_stats/manifests/grafana.pp
@@ -5,8 +5,8 @@ class ocf_stats::grafana {
   }
 
   $canonical_url = $::host_env ? {
-    'dev'  => 'dev-prometheus',
-    'prod' => 'prometheus',
+    'dev'  => 'dev-prometheus.ocf.berkeley.edu',
+    'prod' => 'prometheus.ocf.berkeley.edu',
   }
 
   class { 'grafana':
@@ -22,8 +22,8 @@ class ocf_stats::grafana {
       server           => {
         http_addr => '127.0.0.1',
         http_port => 8990,
-        domain    => "${canonical_url}.ocf.berkeley.edu",
-        root_url  => "https://${canonical_url}.ocf.berkeley.edu/grafana/",
+        domain    => $canonical_url,
+        root_url  => "https://${canonical_url}/grafana/",
       },
       database         => {
         type => 'sqlite3',
@@ -49,7 +49,7 @@ class ocf_stats::grafana {
           name      => 'Prometheus',
           type      => 'prometheus',
           access    => 'direct',
-          url       => 'https://prometheus.ocf.berkeley.edu/',
+          url       => "https://${canonical_url}/",
           isDefault => true,
         },
       ],
--- a/modules/ocf_stats/manifests/prometheus.pp
+++ b/modules/ocf_stats/manifests/prometheus.pp
@@ -71,6 +71,7 @@ class ocf_stats::prometheus {
       request_headers     => ['set X-Forwarded-Proto https'],
 
       rewrites            => [
+        {rewrite_rule => '^/grafana/(.*)$ http://127.0.0.1:8990/$1 [P]'},
         {rewrite_rule => '^/(.*)$ http://127.0.0.1:9090/$1 [P]'},
       ];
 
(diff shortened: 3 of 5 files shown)


modules/ocf_stats/manifests/prometheus.pp
```diff
@@ -71,6 +71,7 @@
       request_headers     => ['set X-Forwarded-Proto https'],
 
       rewrites            => [
+        {rewrite_rule => '^/grafana/(.*)$ http://127.0.0.1:8990/$1 [P]'},
```
jvperrin: Why not use `grafana.ocf.berkeley.edu` as a separate vhost instead of having this as a path under prometheus?

modules/ocf_stats/manifests/grafana.pp
```diff
@@ -0,0 +1,62 @@
...
+    'prod' => 'prometheus.ocf.berkeley.edu',
+  }
+
+  class { 'grafana':
+    install_method           => 'repo',
+
+    # Unfortunately, puppet-grafana requires that we specify a version number
+    # here.
+    version                  => '5.2.1',
+    manage_package_repo      => false,
```
jvperrin: Is there a reason why the whole apt key and repo need to be specified manually instead of letting the module manage both of them? That might fit the version issue too, although I'm not sure on that.

modules/ocf_stats/manifests/grafana.pp
```diff
@@ -0,0 +1,62 @@
...
+      app_mode         => 'production',
+      server           => {
+        http_addr => '127.0.0.1',
+        http_port => 8990,
+        domain    => $canonical_url,
+        root_url  => "https://${canonical_url}/grafana/",
+      },
+      database         => {
+        type => 'sqlite3',
+        path => 'grafana.db',
```
jvperrin: What kind of data does this store? While this could be using mysql, I think if you want do it this way instead of using docker, then using sqlite is probably fine. Does this data need to be backed up at all?

modules/ocf_stats/manifests/grafana.pp
```diff
@@ -0,0 +1,62 @@
...
+      },
+      users            => {
+        auto_assign_org      => true,
+        allow_sign_up        => false,
+        auto_assign_org_role => 'Editor',
```
jvperrin: With anonymous auth enabled, does this mean that anyone can edit any of the dashboards without authenticating? I think allowing anonymous access is probably fine, but I don't think anonymous editing should be enabled.

modules/ocf_stats/manifests/grafana.pp
```diff
@@ -0,0 +1,62 @@
...
+      },
+      'auth.anonymous' => {
+        enabled  => true,
+      },
+      security         => {
+        disable_gravatar => true,
```
jvperrin: Why? We use these for staff hours, so they should be reasonably useful if enabled

jvperrin: Also, without an admin user here, who would have access to administer the grafana instance? Would that all be done through puppet? (That's fine if it's the case, I'm just not sure what your plan on this is)

modules/ocf_stats/manifests/grafana.pp
jvperrin: I initially looked into installing grafana from the official Debian repo, but found that it's quite out-of-date and that [updating it with new packaging is a pain](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=835210) due to lack of upstream support.

modules/ocf/manifests/packages/grafana/apt.pp
```diff
@@ -0,0 +1,13 @@
+class ocf::packages::grafana::apt {
+  apt::key { 'grafana':
+    id     => '418A7F2FB0E1E6E7EABF6FE8C2E73424D59097AB',
+    source => 'https://packagecloud.io/gpg.key',
+  }
+
+  apt::source { 'grafana':
+    location => 'https://packagecloud.io/grafana/stable/debian/',
```
jvperrin: Using just http should be fine here too, since the packages are signed, but we should also already have `apt-transport-https` installed so this still works. Might want to add that package as a dependency here if you use https?

modules/ocf_stats/manifests/grafana.pp
```diff
@@ -0,0 +1,62 @@
...
+
+  class { 'grafana':
+    install_method           => 'repo',
+
+    # Unfortunately, puppet-grafana requires that we specify a version number
+    # here.
+    version                  => '5.2.1',
+    manage_package_repo      => false,
+
+    cfg                      => {
```
kkuehler: weird indentation

modules/ocf_stats/manifests/grafana.pp
```diff
@@ -0,0 +1,62 @@
...
+        http_addr => '127.0.0.1',
+        http_port => 8990,
+        domain    => $canonical_url,
+        root_url  => "https://${canonical_url}/grafana/",
+      },
+      database         => {
+        type => 'sqlite3',
+        path => 'grafana.db',
+      },
+      users            => {
```
kkuehler: is this supposed to be more readable?

modules/ocf_stats/manifests/grafana.pp
```diff
@@ -0,0 +1,62 @@
...
+  $canonical_url = $::host_env ? {
+    'dev'  => 'dev-prometheus.ocf.berkeley.edu',
+    'prod' => 'prometheus.ocf.berkeley.edu',
+  }
+
+  class { 'grafana':
+    install_method           => 'repo',
+
+    # Unfortunately, puppet-grafana requires that we specify a version number
+    # here.
```
kkuehler: remove the 'here' and just have one comment line

modules/ocf_stats/manifests/grafana.pp
```diff
@@ -0,0 +1,62 @@
...
+        domain    => $canonical_url,
+        root_url  => "https://${canonical_url}/grafana/",
+      },
+      database         => {
+        type => 'sqlite3',
+        path => 'grafana.db',
+      },
+      users            => {
+        auto_assign_org      => true,
+        allow_sign_up        => false,
```
kkuehler: grafana also supports user auth through ldap. i know ldap is a pain but we should probably aim for this in the future?

modules/ocf_stats/manifests/grafana.pp
kkuehler: Change to `Viewer`?

modules/ocf_stats/manifests/grafana.pp
```diff
@@ -0,0 +1,62 @@
...
+    version                  => '5.2.1',
+    manage_package_repo      => false,
+
+    cfg                      => {
+      app_mode         => 'production',
+      server           => {
+        http_addr => '127.0.0.1',
+        http_port => 8990,
+        domain    => $canonical_url,
+        root_url  => "https://${canonical_url}/grafana/",
```
kkuehler: nice. i like this

modules/ocf_stats/manifests/prometheus.pp
kkuehler: i agree here. it seems odd to have grafana under the prometheus vhost when prometheus is grafana's data source. the other way around would make more sense, but separate vhosts is better

modules/ocf_stats/manifests/grafana.pp
dkess: The way the module sets up the repo leads to errors in our setup. I'd be willing to look into this more and/or make a PR on their end to fix it.

modules/ocf_stats/manifests/grafana.pp
dkess: I'm pretty sure the linter enforced this, but I can check again.

jvperrin:
Ok, I'd like to make sure the anonymous dashboards are actually emphemeral, since otherwise we could end up with a large sqlite database just from anonymous users creating random dashboards that we don't actually care about.

If there's an admin account set up and not puppeted, it sounds like the database should be backed up (along with the other stuff we back up in ocf_backups), or the admin user should be puppeted. If this is the only thing that makes the database need to be backed up, then puppeting it would be much better in my opinion. I think it's also worth testing setting it up from scratch to make sure that it goes smoothly, since that's the main reason why we actually use puppet.

I think the lack of user functionality and the dashboard creation through puppet are both good, and using sqlite is fine too, I'm just concerned that it's not possible to recreate the current setup without some extra work.

chriskuehl:
I think overall it's probably preferable to use mysql over sqlite as long as mysql being down won't break alerting, though it's not a big deal as long as we back up the sqlite db.

dkess:
I agree we should prioritize being able to recreate the setup from scratch. It's possible to specify the admin password in the config file, but it's only a "one-time" thing-- changing the password in the admin interface touches the database, and the config option will be ignored.

I'll change this to use MySQL instead of sqlite. Grafana does have alerting features, but I don't expect us to use them. Prometheus has its own alerting system that doesn't depend on any database.

Re: anonymous user permissions, Gitlab has an openly accessible Grafana instance at https://monitor.gitlab.net/, but it doesn't seem to let anonymous users make new dashboards (unless I just can't find this functionality). I may be wrong about what I said earlier, in that case we'll have to figure out how to handle authentication.

dkess:
Gonna close this PR since I think Marathon is a better place for Grafana.

PR #341, opened 2018-07-25
Proposal: set apache::default_vhost to false in hieradata

kpengboy:
It seems that we always set `apache::default_vhost` to `false` whenever we include the `apache` module. Setting it in hieradata would allow us to eliminate some duplicate code (and also make it easier to split munin and prometheus into separate modules).

What do you all think? Is this a worthwhile change? Would this be an abuse of hiera?


chriskuehl:
if it works, i think it's fine to ship. i probably have a slight preference for not doing this, but no strong opposition.

the main downsides i see are:
* it's more implicit (meh)
* it might break silently during an upgrade (passing in a non-existent class parameter *will* cause the puppet run to fail loudly, but a hiera value that nothing reads won't notice; this might cause the default vhost to silently get enabled without us noticing)
* it could break third-party modules that configure apache (if we were to use any in the future, and if they were to depend on the default vhost), similar to how some of our defaults for the `file` resource used to break things; i think this is fairly unlikely though

PR #342, opened 2018-07-27
Change mesos sub filter to fix log links

jvperrin:
The JSONP data returned by Mesos has changed format a bit (which broke the stdout/stderr links for quite a while now). It now includes the IP of each of the slaves, along with a different order to the data that was originally replaced in 6db3df5d30567780ae084418f3d9f64b7e8bbfcf.

I'd like to not include the IPs of each of the slaves in the hiera data, but I don't see a great way to not include it. Doing a DNS lookup inside puppet seems like an even worse move to me...


jvperrin:
Hmmm, that's a good idea, that might be possible somehow

kpengboy:
> Doing a DNS lookup inside puppet seems like an even worse move to me...

Then again, we already do this for the firewall...

PR #344, opened 2018-08-04
Squelch ISP false positives about outdated Apache

kpengboy:
By hiding the minimal version number, it might be possible to prevent ISP's security scanners from getting the impression that we run an outdated version of Apache.

What do you all think about doing this? Yea? Nay?


jvperrin:
Sure, that it removes the header is enough, I'd definitely be in favor of shipping this! (of course this won't work with staff VMs, but we can fix those manually if needed)

jvperrin:
Agreed, I forgot about those since they're behind an nginx proxy, I guess those need to be updated per-repo to add the apache option

PR #346, opened 2018-08-05
Enable HTTPS for dev apphosts

jvperrin:
This enables HTTPS on sites like https://dev-app-ocf-berkeley-edu.apphost.ocf.berkeley.edu/ and https://hkn-eecs-berkeley-edu.apphost.ocf.berkeley.edu/, which is mostly just useful for when groups are setting up their apphost but don't have the DNS set up for it yet.

Maybe we should redirect these dev hosts to the actual domains if the DNS is working for their domain? It's a little strange to have multiple domains that work to access a site, but we've also had it like that for quite a while without any problems, so I don't think it needs to be done right now.

I believe this is actually the last non-HTTPS set of sites we offer! (I'm probably wrong on that, but I can't think of any others)

### DIFF
--- a/modules/ocf_www/files/build-vhosts
+++ b/modules/ocf_www/files/build-vhosts
@@ -55,7 +55,7 @@ class SSL(namedtuple('SSL', ('fqdn',))):
         return str(SSL_BUNDLE / (self.fqdn + '.bundle'))
 
 
-class System_SSL(namedtuple('System_SSL', ('fqdn',))):
+class SystemSSL(namedtuple('SystemSSL', ('fqdn',))):
     """Tells where to find the standard OCF SSL files for this host (wildcards,
     fqdn.o.b.e, etc.).
     """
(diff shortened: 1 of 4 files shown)


modules/ocf_www/files/build-vhosts
```diff
@@ -53,6 +55,20 @@ class SSL(namedtuple('SSL', ('fqdn',))):
         return str(SSL_BUNDLE / (self.fqdn + '.bundle'))
 
 
+class System_SSL(namedtuple('System_SSL', ('fqdn',))):
```
kpengboy: Why the funny naming convention?

jvperrin: Dunno, have any better names? I'd be open to suggestions

kpengboy: What about `SystemSSL`?

jvperrin: Oh, I see what you mean about the naming convention now, that makes more sense. Sure, this would be better

jvperrin:
Good point, but they could have multiple apphosts, so that's why it's like this I believe

PR #349, opened 2018-08-17
allow traffic to xcf webserver, remove migrated xcf machines

abizer:
drm and ponzi have been migrated to AS26073


kpengboy:
@abizer It would've been better to do it by adding the needed firewall rule. `ocf::firewall::reject_unrecognized_input` is really intended only as an emergency kill switch.

You may find the rules in `ocf_www` useful as a reference.

PR #351, opened 2018-08-20
[HPC] Autoadd SLURM users from LDAP.

redplanks:
Compares which users are in SLURM and which are in the "ocfhpc" LDAP group; adds new users to SLURM. Since user-account combinations are unique, the script is idempotent.

Instead of using a variable in the script to set which account new users go under, would it be better to use a command-line flag?

Being tested right now on segfault; "guser" and "jvp" were successfully added on the script's last run.

### DIFF
--- a/modules/ocf_hpc/files/add_slurm_users
+++ b/modules/ocf_hpc/files/add_slurm_users
@@ -1,8 +1,9 @@
 #!/usr/bin/env python3
-'''
+"""
 Script that adds users in the 'ocfhpc' LDAP group to SLURM as a SLURM user.
-'''
+"""
 import subprocess
+import sys
 
 from ocflib.account import search
 from ocflib.account import utils
@@ -11,23 +12,22 @@ DEFAULT_ACCOUNT = 'users'
 
 
 def main():
-    ocfhpc_ldap_user_set = set(utils.list_group('ocfhpc'))
-
     # This produces a newline-separated list of SLURM users.
     ocfhpc_slurm_user_list = subprocess.check_output(
         ('sacctmgr', '-noP', 'list', 'users', 'format=User'),
-        universal_newlines=True,
-    ).split('\n')[:-1]
-    ocfhpc_slurm_user_set = set(ocfhpc_slurm_user_list)
+    ).decode('utf-8').split('\n')[:-1]
 
-    users_to_be_added = ocfhpc_ldap_user_set.difference(ocfhpc_slurm_user_set)
+    ocfhpc_slurm_user_set = set(ocfhpc_slurm_user_list)
+    ocfhpc_ldap_user_set = set(utils.list_group('ocfhpc'))
+
+    users_to_be_added = ocfhpc_ldap_user_set - ocfhpc_slurm_user_set
 
     if not users_to_be_added:
         return
 
     for new_user in users_to_be_added:
         if not search.user_exists(new_user):
-            print('User {} in the HPC LDAP group does not exist.'.format(new_user))
+            print('User {} in the HPC LDAP group does not exist.'.format(new_user), file=sys.stderr)
             continue
 
         # Add a new user to SLURM and commit changes immediately.
--- a/modules/ocf_hpc/files/add_slurm_users
+++ b/modules/ocf_hpc/files/add_slurm_users
@@ -14,11 +14,10 @@ def main():
     ocfhpc_ldap_user_set = set(utils.list_group('ocfhpc'))
 
     # This produces a newline-separated list of SLURM users.
-    ocfhpc_slurm_user_list = subprocess.run(
-        ['sacctmgr', '-noP', 'list', 'users', 'format=User'],
-        stdout=subprocess.PIPE,
-        check=True
-    ).stdout.decode().split('\n')[:-1]
+    ocfhpc_slurm_user_list = subprocess.check_output(
+        ('sacctmgr', '-noP', 'list', 'users', 'format=User'),
+        universal_newlines=True,
+    ).split('\n')[:-1]
     ocfhpc_slurm_user_set = set(ocfhpc_slurm_user_list)
 
     users_to_be_added = ocfhpc_ldap_user_set.difference(ocfhpc_slurm_user_set)
@@ -27,13 +26,12 @@ def main():
         return
 
     for new_user in users_to_be_added:
-
         if not search.user_exists(new_user):
             print('User {} in the HPC LDAP group does not exist.'.format(new_user))
             continue
 
         # Add a new user to SLURM and commit changes immediately.
-        subprocess.run(['sacctmgr', '-i', 'add', 'user', new_user, 'account={}'.format(DEFAULT_ACCOUNT)])
+        subprocess.run(('sacctmgr', '-i', 'add', 'user', new_user, 'account={}'.format(DEFAULT_ACCOUNT)))
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


modules/ocf_hpc/manifests/controller.pp
```diff
@@ -24,4 +24,15 @@
       File['/etc/slurm-llnl/cgroup.conf']
     ],
   }
+
+  # Set up script to autoadd LDAP group members to SLURM.
+  file { '/usr/local/bin/add_slurm_users.py':
```
abizer: I'd remove the extension

modules/ocf_hpc/files/add_slurm_users.py
```diff
@@ -0,0 +1,35 @@
...
+    ocfhpc_slurm_user_set = set(ocfhpc_slurm_user_list)
+
+    users_to_be_added = ocfhpc_ldap_user_set.difference(ocfhpc_slurm_user_set)
+
+    if not users_to_be_added:
+        return
+
+    for new_user in users_to_be_added:
+        # Add a new user to SLURM and commit changes immediately.
+        subprocess.run(['sacctmgr', '-i', 'add', 'user', new_user, 'account={}'.format(DEFAULT_ACCOUNT)])
```
Porrster: I would double check (if you haven't already) that this command adds users correctly, because in the past I believe I've had to specify the cluster and/or partition.

redplanks: Yep, it does. Default cluster is ocf-hpc, and apparently it doesn't need a partition.

modules/ocf_hpc/manifests/controller.pp
redplanks: The full path?

abizer: no like leave off the .py

modules/ocf_hpc/manifests/controller.pp
```diff
@@ -24,4 +24,15 @@
       File['/etc/slurm-llnl/cgroup.conf']
     ],
   }
+
+  # Set up script to autoadd LDAP group members to SLURM.
+  file { '/usr/local/bin/add_slurm_users':
+    source => 'puppet:///modules/ocf_hpc/add_slurm_users.py',
```
jvperrin: Removing the extension on this file (within the puppet repo) is probably best, since it's not in the added file, so it keeps things more consistent.

modules/ocf_hpc/manifests/controller.pp
```diff
@@ -24,4 +24,15 @@
...
+    owner  => 'slurm',
+    mode   => '0755',
+  } -> cron { 'add_slurm_users':
+    command => '/usr/local/bin/add_slurm_users',
+    user    => 'slurm',
```
jvperrin: Nice, non-root is great, thanks for doing this.

modules/ocf_hpc/manifests/controller.pp
```diff
@@ -24,4 +24,15 @@
       File['/etc/slurm-llnl/cgroup.conf']
     ],
   }
+
+  # Set up script to autoadd LDAP group members to SLURM.
+  file { '/usr/local/bin/add_slurm_users':
+    source => 'puppet:///modules/ocf_hpc/add_slurm_users.py',
+    owner  => 'slurm',
```
jvperrin: Is this user added by anything in puppet? If so, then an extra dependency should probably be added here so the file and cronjob aren't added before the user actually exists. Anyway, not a big deal on this though.

jvperrin:
Replying to @Porrster, I think every 15 minutes is perfectly fine, it's a very lightweight script, probably could even sync more often if you really want, but given how often new people are added, 15 minutes seems fine.

modules/ocf_hpc/manifests/controller.pp
redplanks: The slurm user is added when the "slurm-wlm" package is installed in the ocf_hpc class, so I think the "require ocf_hpc" line in controller.pp should take care of it.

modules/ocf_hpc/files/add_slurm_users
```diff
@@ -0,0 +1,38 @@
...
+
+DEFAULT_ACCOUNT = 'users'
+
+
+def main():
+    ocfhpc_ldap_user_set = set(utils.list_group('ocfhpc'))
+
+    # This produces a newline-separated list of SLURM users.
+    ocfhpc_slurm_user_list = subprocess.check_output(
+        ('sacctmgr', '-noP', 'list', 'users', 'format=User'),
```
kkuehler: Adding `AdminLevel=None` to this command will exclude unnecessary entries, like you and wporr.

modules/ocf_hpc/files/add_slurm_users
```diff
@@ -0,0 +1,38 @@
...
+    ocfhpc_slurm_user_set = set(ocfhpc_slurm_user_list)
+
+    users_to_be_added = ocfhpc_ldap_user_set.difference(ocfhpc_slurm_user_set)
+
+    if not users_to_be_added:
+        return
+
+    for new_user in users_to_be_added:
+        if not search.user_exists(new_user):
+            print('User {} in the HPC LDAP group does not exist.'.format(new_user))
```
kkuehler: I'd go for a more meaningful error, such as:

"User {} in the HPC LDAP group not managed by slurm."

modules/ocf_hpc/files/add_slurm_users
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
+'''
```
kpengboy: PEP8 recommends `"""` only

modules/ocf_hpc/files/add_slurm_users
```diff
@@ -0,0 +1,38 @@
...
+DEFAULT_ACCOUNT = 'users'
+
+
+def main():
+    ocfhpc_ldap_user_set = set(utils.list_group('ocfhpc'))
+
+    # This produces a newline-separated list of SLURM users.
+    ocfhpc_slurm_user_list = subprocess.check_output(
+        ('sacctmgr', '-noP', 'list', 'users', 'format=User'),
+        universal_newlines=True,
```
kpengboy: Elsewhere in the codebase when this pattern occurs, we don't specify `universal_newlines` but rather `decode('utf-8')` the result. It may be better to maintain consistency. I'll leave it up to your discretion.

modules/ocf_hpc/files/add_slurm_users
kpengboy: Nit: I would print this to stderr

This line should never be printed, right?

modules/ocf_hpc/files/add_slurm_users
redplanks: wporr and I are in the ocfhpc ldap group though, so it'll try to create users for us again in that case, and if we're set to a non-default account, it'll create a new user-account association.

modules/ocf_hpc/files/add_slurm_users
redplanks: Yeah, I can't really think of any cases where this would be printed.

modules/ocf_hpc/manifests/controller.pp
```diff
@@ -24,4 +24,15 @@
       File['/etc/slurm-llnl/cgroup.conf']
     ],
   }
+
+  # Set up script to autoadd LDAP group members to SLURM.
+  file { '/usr/local/bin/add_slurm_users':
+    source => 'puppet:///modules/ocf_hpc/add_slurm_users',
+    owner  => 'slurm',
+    mode   => '0755',
+  } -> cron { 'add_slurm_users':
```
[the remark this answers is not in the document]
redplanks: Ah, I'll have to rewrite the rest of the manifests in ocf_hpc then. Will make a new commit after merging.

redplanks:
No plans for auto-removing users right now, but if we have to do it a lot, it shouldn't be too hard of a modification, hopefully.

PR #352, opened 2018-08-21
[HPC] Automatically configure NVIDIA GPU generic resources in SLURM.

redplanks:
Produces output in slurm.conf:
```
NodeName=corruption NodeAddr=169.229.226.18 CPUs=40 Sockets=2 CoresPerSocket=10 ThreadsPerCore=2 RealMemory=257855 Gres=gpu:nv1080:1,gpu:nv1080:1,gpu:nv1080:1,gpu:nv1080:1, State=UNKNOWN
```

and in gres.conf:
```
# These are the gres devices available on this node.

Name=gpu File=/dev/nvidia0 Type=nv1080
Name=gpu File=/dev/nvidia1 Type=nv1080
Name=gpu File=/dev/nvidia2 Type=nv1080
Name=gpu File=/dev/nvidia3 Type=nv1080
```

Requesting multiple GPUs in SLURM using `--gres=gpu:4` or `--gres=gpu:nv1080:4` still works; thankfully SLURM is intelligent enough to group GPUs of the same type.

I also assume that the `index` field supplied by `nvidia-smi` corresponds to the number of the GPU's devfile; the docs don't say so hopefully that's so.

### DIFF
--- a/modules/ocf_hpc/facts.d/nvidia_gpu_info
+++ b/modules/ocf_hpc/facts.d/nvidia_gpu_info
@@ -1,0 +1,36 @@
+#!/usr/bin/env python3
+import re
+import shutil
+import subprocess
+
+
+def main():
+    if not shutil.which('nvidia-smi'):
+        return
+
+    try:
+        nvidia_smi_output = subprocess.run(['nvidia-smi', '--query-gpu=name,index', '--format=csv,noheader'],
+                                           stdout=subprocess.PIPE,
+                                           universal_newlines=True,
+                                           check=True,
+                                           ).stdout
+    except subprocess.CalledProcessError:
+        return
+
+    nvidia_gpu_list = nvidia_smi_output.split('\n')[:-1]
+
+    nvidia_gpu_info_dict = {}
+
+    for gpu_string in nvidia_gpu_list:
+        # Used for determining what the numbers of the GPU's device files in /dev
+        gpu_index = int(gpu_string.split(',')[1])
+        # Strip all except numbers and add 'nv' in front to distinguish from number of GPUs
+        # when specifying type in SLURM.
+        gpu_model_number = 'nv' + re.sub('[^0-9]', '', gpu_string.split(',')[0])
+        nvidia_gpu_info_dict[gpu_index] = gpu_model_number
+
+    print('nvidia_gpu_info={}'.format(nvidia_gpu_info_dict))
+
+
+if __name__ == '__main__':
+    main()
--- a/modules/ocf_hpc/facts.d/nvidia_gpu_info
+++ b/modules/ocf_hpc/facts.d/nvidia_gpu_info
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import json
 import re
 import shutil
 import subprocess
@@ -29,7 +30,7 @@ def main():
         gpu_model_number = 'nv' + re.sub('[^0-9]', '', gpu_string.split(',')[0])
         nvidia_gpu_info_dict[gpu_index] = gpu_model_number
 
-    print('nvidia_gpu_info={}'.format(nvidia_gpu_info_dict))
+    print('nvidia_gpu_info={}'.format(json.dumps(nvidia_gpu_info_dict)))
 
 
 if __name__ == '__main__':
--- a/modules/ocf_hpc/facts.d/nvidia_gpu_info
+++ b/modules/ocf_hpc/facts.d/nvidia_gpu_info
@@ -5,30 +5,27 @@ import shutil
 import subprocess
 
 
+def strip_non_ints(string):
+    return re.sub('[^0-9]', '', string)
+
+
 def main():
     if not shutil.which('nvidia-smi'):
+        print('NVIDIA drivers not installed; machine probably does not have NVIDIA GPUs.')
         return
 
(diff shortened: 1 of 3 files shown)


modules/ocf_hpc/facts.d/nvidia_gpu_info
```diff
@@ -0,0 +1,40 @@
...
+import shutil
+import subprocess
+
+
+def main():
+    if not shutil.which('nvidia-smi'):
+        return
+
+    try:
+        nvidia_smi_output = subprocess.run(['nvidia-smi', '--query-gpu=name,index', '--format=csv,noheader'],
```
jvperrin: I'd also use `subprocess.check_output` here, like I mentioned in #351, I think it'll clean things up a fair amount.

modules/ocf_hpc/facts.d/nvidia_gpu_info
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
+import json
+import re
+import shutil
+import subprocess
+
+
+def main():
+    if not shutil.which('nvidia-smi'):
```
jvperrin: Wouldn't the check below for `subprocess.CalledProcessError` handle this? In any case, I think it's probably better to keep this check here and remove the exception handling, since failing loudly on that is probably good.

kkuehler: might be worth outputting "are you sure you are on an HPC machine?"

modules/ocf_hpc/facts.d/nvidia_gpu_info
```diff
@@ -0,0 +1,40 @@
...
+        nvidia_smi_output = subprocess.run(['nvidia-smi', '--query-gpu=name,index', '--format=csv,noheader'],
+                                           stdout=subprocess.PIPE,
+                                           universal_newlines=True,
+                                           check=True,
+                                           ).stdout
+    except subprocess.CalledProcessError:
+        return
+
+    nvidia_gpu_list = nvidia_smi_output.split('\n')[:-1]
+
```
kkuehler: nit: random newline

modules/ocf_hpc/facts.d/nvidia_gpu_info
```diff
@@ -0,0 +1,40 @@
...
+    nvidia_gpu_info_dict = {}
+
+    for gpu_string in nvidia_gpu_list:
+        # Used for determining what the numbers of the GPU's device files in /dev
+        gpu_index = int(gpu_string.split(',')[1])
+        # Strip all except numbers and add 'nv' in front to distinguish from number of GPUs
+        # when specifying type in SLURM.
+        gpu_model_number = 'nv' + re.sub('[^0-9]', '', gpu_string.split(',')[0])
```
kkuehler: Instead of calling split twice I'd do this...

```python
gpu_model, gpu_index = gpu_string.split(',', 1)
gpu_model_number = 'nv' + re.sub('[^0-9]', '', gpu_model)
nvidia_gpu_info_dict[int(gpu_index)] = gpu_model_number
...
```

Doing that split with the regex and string concatenation makes the code unnecessarily confusing imo. Honestly wouldn't hurt to put the regex deletion in a helper method called `strip_ints`.

modules/ocf_hpc/templates/nodes-partitions.erb
```diff
@@ -1,6 +1,6 @@
-# Currently there is a 'Gres=...' entry that should be changed to dynamically handle
-# gres resources on selected nodes
+# We do a nested iteration over the nvidia_gpu_info fact to configure GRES.
+# In addition, nvidia_gpu_info comes in as a JSON-formatted string, so we have to convert back into a hash first.
 <% @slurm_nodes_facts.each do |facts| -%>
-NodeName=<%= facts["hostname"] %> NodeAddr=<%= facts["ipaddress"] %> CPUs=<%= facts["processorcount"] %> Sockets=<%= facts["physicalprocessorcount"] %> CoresPerSocket=<%= facts["cores_per_socket"] %> ThreadsPerCore=<%= facts["threads_per_core"] %> RealMemory=<%= facts["memorysize_mb"].floor %> Gres=gpu:2 State=UNKNOWN
+NodeName=<%= facts["hostname"] %> NodeAddr=<%= facts["ipaddress"] %> CPUs=<%= facts["processorcount"] %> Sockets=<%= facts["physicalprocessorcount"] %> CoresPerSocket=<%= facts["cores_per_socket"] %> ThreadsPerCore=<%= facts["threads_per_core"] %> RealMemory=<%= facts["memorysize_mb"].floor %> Gres=<% JSON.parse(facts["nvidia_gpu_info"]).each do |_, model| -%>gpu:<%= model %>:1,<% end -%> State=UNKNOWN
```
kkuehler: dank

redplanks:
@jvperrin segfault is running the test env for this PR, and it looks like it doesn't have the fact, according to Puppetboard. Should I make the fact blank if the machine doesn't have any NVIDIA cards?

redplanks:
I could probably build the string used in slurm.conf from the Ruby hash using Puppet functions, if that would be more clear.

Porrster:
Looks good to me! Thanks for the PR.

PR #353, opened 2018-08-22
Update the opstaff shorturl for Fall 2018

EnnFour:


dkess:
Change `opstaff-schedule` too

PR #357, opened 2018-08-27
Add python3.7(-dev/-venv) to packages

ethanhs:
This will allow experimenting with moving ocflib and other OCF projects to newer versions of Python which are not only are faster, but also make static typing nicer and easier.


ethanhs:
I changed this to be in packages because we eventually want to shift utils and other things to 3.7.

PR #359, opened 2018-08-28
Add wildcard shorturl for Discourse threads

dkess:
Any objections to this?


abizer:
you should add a discourse.ocf.berkeley.edu DNS CNAME first to your staffvm until we can move discourse to Marathon

jvperrin:
Yeah I'd try to move this to marathon before setting up this, it'll just need updating otherwise

dkess:
This now points directly discourse.o.b.e, which redirects to the forum on leprosy.

Once Discourse is deployed to production, we can reverse the direction of this redirect (so leprosy points to discourse).

PR #360, opened 2018-08-28
add ocfstaff@lists.berkeley.edu to staff@ alias

abizer:
I made ocfstaff@lists.berkeley.edu so we can have a group for people
who are interested in being staff, or receiving staff updates, without
necessarily needing to give those people additional privileges, especially
since only a tiny fraction of the people on cn=ocfstaff use their privs
at any time. Using this group, we can migrate all 'staff' to the mailing list
where they can continue to get updates, and should they want to get privs,
they can sign up for staff hours and get added to the ocfstaff LDAP group.

The group has the additional benefit of archiving all messages sent to staff
which could be useful for documentation purposes down the road.

### DIFF
--- a/modules/ocf_mail/files/site_ocf/update-aliases
+++ b/modules/ocf_mail/files/site_ocf/update-aliases
@@ -5,7 +5,7 @@ set -o pipefail
 
 staff=$(getent group ocfstaff | cut -d':' -f4)
 
-echo "staff-current: ${staff}" > /etc/aliases-group
+echo "staff-current: ocfstaff@lists.berkeley.edu,${staff}" > /etc/aliases-group
 /usr/sbin/postalias hash:/etc/aliases-group
 
 # update aliases-local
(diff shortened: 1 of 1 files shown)


abizer:
I'm not sure if this is a good way to implement this, though. Very open to suggestions as to how to do this better.

abizer:
I agree that we should send more mail to staff@. One of the benefits of the lists.berkeley.edu approach, I think, is that anyone with an @berkeley.edu can join the group themselves and get updates, which will hopefully inspire some people to come hang out at the lab, where they can then get added to the ocfstaff LDAP group should they be so interested. If we are moving to a future where there is no supernova, then the 'log into supernova' aspect of staff becomes moot, and if they need to run approve, then they should already have staff hours, which would qualify them to be in the ocfstaff LDAP group. For the time being this seems like a reasonable compromise between helping people get more info about the OCF and giving them incentives to join, while also managing the over-proliferation of privileges. Thoughts?

kpengboy:
I'm slightly concerned about the risk of confusion between the ocfstaff
LDAP group and the ocfstaff@lists.berkeley.edu mailing list. Ideally the
latter would be renamed. Otherwise no objections.

PR #363, opened 2018-09-08
make ocfbroker a system group

kkuehler:
clashed with the `ocf` gid on some desktops... thus overriding it and every user became part of `ocfbroker`. since `ocfbroker` has no home dir, i think pam didnt bother mounting the remote from tsunami. pretty hilarious.


kpengboy:
Thanks for fixing this.

There's still a few more changes that need to be made; namely, the existing
ocfbroker users will need to be deleted.

PR #366, opened 2018-09-12
Postgres @ OCF

abizer:
This introduces the preliminary Puppet aspect of a Postgres service @ OCF. The provisional host is hozer-72, but I'm going to create a new machine at .66 to serve as the Postgres host. 

I'm planning on doing role and database creation by means of a Celery task like how account creation currently works, rather than by seeding the root password to tsunami like `makemysql` does (and hopefully migrate the mysql method to this create-like service as well). `makepgsql` will then be a wrapper around calling those Celery tasks. Thoughts on this approach? In the meantime, roles and databases can be created manually (for discourse, puppetdb, mastodon, etc. for example).

Also need to add this to the backups, among other things.

### DIFF
--- a/hieradata/nodes/lethe.yaml
+++ b/hieradata/nodes/lethe.yaml
@@ -1,0 +1,2 @@
+classes:
+  - ocf_postgres
--- a/hieradata/nodes/lethe.yaml
+++ b/hieradata/nodes/lethe.yaml
@@ -1,4 +1,2 @@
 classes:
   - ocf_postgres
-
-ocf::ssl::default::owner: root
--- a/hieradata/nodes/lethe.yaml
+++ b/hieradata/nodes/lethe.yaml
@@ -1,2 +1,4 @@
 classes:
   - ocf_postgres
+
+ocf::ssl::default::owner: root
--- a/modules/ocf/manifests/ssl/bundle.pp
+++ b/modules/ocf/manifests/ssl/bundle.pp
@@ -1,8 +1,8 @@
 define ocf::ssl::bundle(
   Boolean $use_lets_encrypt = true,
   Array[String] $domains = [$title],
-  String $owner = ocfletsencrypt,
-  String $group = ssl-cert,
+  String $owner = 'ocfletsencrypt',
+  String $group = 'ssl-cert',
 ) {
   require ocf::ssl::setup
 
@@ -16,7 +16,6 @@ define ocf::ssl::bundle(
     $intermediate_source = 'puppet:///modules/ocf/ssl/lets-encrypt.crt'
     $cert_path = "/var/lib/lets-encrypt/certs/${title}/cert.pem"
     $key_path = "/var/lib/lets-encrypt/certs/${title}/privkey.pem"
-    $fullchain_path = "/var/lib/lets-encrypt/certs/${title}/fullchain.pem"
     $cert_source = "file://${cert_path}"
     $key_source = "file://${key_path}"
 
@@ -40,12 +39,6 @@ define ocf::ssl::bundle(
         ensure => symlink,
         links  => manage,
         target => $cert_path,
-        mode   => '0644';
-
-      "/etc/ssl/private/${title}.fullchain":
-        ensure => symlink,
-        links  => manage,
-        target => $fullchain_path,
         mode   => '0644';
 
       "/etc/ssl/private/${title}.intermediate":
--- a/modules/ocf/manifests/ssl/bundle.pp
+++ b/modules/ocf/manifests/ssl/bundle.pp
@@ -16,6 +16,7 @@ define ocf::ssl::bundle(
     $intermediate_source = 'puppet:///modules/ocf/ssl/lets-encrypt.crt'
     $cert_path = "/var/lib/lets-encrypt/certs/${title}/cert.pem"
     $key_path = "/var/lib/lets-encrypt/certs/${title}/privkey.pem"
+    $fullchain_path = "/var/lib/lets-encrypt/certs/${title}/fullchain.pem"
     $cert_source = "file://${cert_path}"
     $key_source = "file://${key_path}"
 
@@ -39,6 +40,12 @@ define ocf::ssl::bundle(
         ensure => symlink,
         links  => manage,
         target => $cert_path,
+        mode   => '0644';
+
+      "/etc/ssl/private/${title}.fullchain":
+        ensure => symlink,
+        links  => manage,
+        target => $fullchain_path,
         mode   => '0644';
 
       "/etc/ssl/private/${title}.intermediate":
--- a/modules/ocf/manifests/ssl/bundle.pp
+++ b/modules/ocf/manifests/ssl/bundle.pp
@@ -1,12 +1,16 @@
 define ocf::ssl::bundle(
   Boolean $use_lets_encrypt = true,
   Array[String] $domains = [$title],
+  String $owner = ocfletsencrypt,
+  String $group = ssl-cert,
 ) {
   require ocf::ssl::setup
 
   if $use_lets_encrypt {
     ocf::ssl::lets_encrypt::dns { $title:
       domains => $domains,
+      owner   => $owner,
+      group   => $group,
     }
 
     $intermediate_source = 'puppet:///modules/ocf/ssl/lets-encrypt.crt'
@@ -41,6 +45,7 @@ define ocf::ssl::bundle(
         source => $intermediate_source,
         mode   => '0644';
     }
+
   } else {
     # TODO: Remove this branch once we are confident enough that using Let's
     # Encrypt certs is working well and is sustainable
--- a/modules/ocf/manifests/ssl/default.pp
+++ b/modules/ocf/manifests/ssl/default.pp
@@ -2,9 +2,9 @@
 # /etc/ssl/private/$cert_name.{key,crt,bundle,intermediate}
 
 class ocf::ssl::default(
-  String $owner = ocfletsencrypt,
-  String $group = ssl-cert,
-){
+  String $owner = 'ocfletsencrypt',
+  String $group = 'ssl-cert',
+) {
   # Attempt to collect all domains for a host to include in a SSL certificate.
   # The '@' record needs to be handled in a special case, since
   # "@.ocf.berkeley.edu" and "@.ocf.io" are both not valid domains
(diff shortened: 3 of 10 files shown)


modules/ocf_postgres/manifests/init.pp
```diff
@@ -0,0 +1,23 @@
+class ocf_postgres {
+  class { 'postgresql::server':
+    postgres_password => hiera('postgres::root'),
+    #                      type    db       usr srcaddr   auth
+    ipv4acls          => ['hostssl sameuser all 0.0.0.0/0 md5'],
+    ipv6acls          => ['hostssl sameuser all ::/0 md5'];
```
jvperrin: Could we use something more secure than md5 here? I think specifying `scram-sha-256` would be better, since using `md5` means using either `md5` or `scram-sha-256` but we don't really want to use md5 anyway.

Also, hostssl is probably fine, but we need to add a cert to this machine using the LE stuff if that's the case, otherwise connections won't be trusted correctly.

The IP ranges here might be better off as restricted to the subnet of the OCF (and localhost?) since that gives an extra layer if protection against outside connections if our external firewall doesn't work (unless you're planning on making this available to outside the subnet?). There's a nice option for that called `samenet` apparently. Alsocording to the postgresql docs, `::0/0` is supposed to be the IPv6 catch-all, but I'm not sure that extra 0 matters or not)

modules/ocf_postgres/manifests/init.pp
```diff
@@ -0,0 +1,23 @@
+class ocf_postgres {
+  class { 'postgresql::server':
+    postgres_password => hiera('postgres::root'),
+    #                      type    db       usr srcaddr   auth
```
jvperrin: I'd probably put a link to the postgresql docs on this (https://www.postgresql.org/docs/current/static/auth-pg-hba-conf.html), since they are much more useful and give nice descriptions of each option and what they mean.

modules/ocf_postgres/manifests/init.pp
abizer: I don't think `scram-sha-256` works before Postgres 10, unfortunately, and we're on 9.6.

I've added the ssl cert stuff to puppet, but Postgres needs the key to be owned by `postgres` or `root`, where it is currently owned by `ocfletsencrypt`. I'll try to find a solution to this.

I was planning on leaving it open to the world as I think OCF members ought to be able to use our db services outside of the context of a webapp running on our infra, like I was using OCF mysql as a backend for storing analysis data from research.

abizer:
@jvperrin any final considerations?

modules/ocf/manifests/ssl/bundle.pp
```diff
@@ -37,10 +42,17 @@
         target => $cert_path,
         mode   => '0644';
 
+      "/etc/ssl/private/${title}.fullchain":
```
jvperrin: The bundle at `/etc/ssl/private/${title}.bundle` should have the same contents as this:

```
jvperrin@lethe:~$ sudo diff /etc/ssl/private/lethe.ocf.berkeley.edu.bundle /etc/ssl/private/lethe.ocf.berkeley.edu.fullchain
43a44
>
```

(except some whitespace changes apparently)

Maybe we should not be using concat and could just link to the fullchain cert instead? Seems cleaner and I think we can remove incommon support now anyway.

Actually, wouldn't using the bundle file fix your permissions issues anyway? I do like that the owner/group can be passed in now, but I don't think we'd need to use it here given the bundle by default is owned by `root:ssl-cert`

modules/ocf/manifests/ssl/setup.pp
```diff
@@ -2,9 +2,14 @@
 # To build and deploy private keys, use the `ocf::ssl::bundle` type.
 # (Or include `ocf::ssl::default` if you just want the default one that includes
 # the FQDN and all aliases of a server.)
+
 class ocf::ssl::setup {
   package { 'ssl-cert':; }
 
+  user { 'ocfletsencrypt':
+    groups => ['ssl-cert', 'sys'],
```
jvperrin: I'd add some dependency here to make sure the user exists (to make sure ldap is available) before changing groups here, otherwise I'm worried that puppet will create this user from scratch, and then it'll get shadowed by ldap (or even worse, this user could shadow an ldap user)

modules/ocf_postgres/manifests/init.pp
```diff
@@ -0,0 +1,35 @@
+class ocf_postgres {
+
```
jvperrin: nit: I'd remove this blank line to follow what most of our other modules do

hieradata/nodes/lethe.yaml
```diff
@@ -0,0 +1,4 @@
+classes:
+  - ocf_postgres
+
+ocf::ssl::default::owner: root
```
kpengboy: If this is a hard requirement of postgres, it would be preferable for the `ocf_postgres` class to set this setting. Remind me, is this class being included from elsewhere?

modules/ocf/manifests/ssl/default.pp
```diff
@@ -1,6 +1,10 @@
 # Provides the key, certificate, and Let's Encrypt CA certificate bundle at
 # /etc/ssl/private/$cert_name.{key,crt,bundle,intermediate}
-class ocf::ssl::default {
+
```
kpengboy: Nit: why the blank line?

modules/ocf/manifests/ssl/default.pp
```diff
@@ -1,6 +1,10 @@
...
+class ocf::ssl::default(
+  String $owner = ocfletsencrypt,
+  String $group = ssl-cert,
```
kpengboy: ibid

modules/ocf/manifests/ssl/lets_encrypt/dns.pp
```diff
@@ -1,5 +1,7 @@
 define ocf::ssl::lets_encrypt::dns(
   Array[String] $domains = [$::fqdn],
+  String $owner = ocfletsencrypt,
+  String $group = ssl-cert,
```
kpengboy: ibid

modules/ocf/manifests/ssl/lets_encrypt/dns.pp
```diff
@@ -23,7 +25,7 @@
     # This exec can be notified to get it to run dehydrated again, even if the
     # cert will not expire soon.
     command     => '/usr/bin/dehydrated --cron --privkey /etc/ssl/lets-encrypt/le-account.key',
-    user        => ocfletsencrypt,
+    user        => $owner,
```
kpengboy: You're sure this is safe to run as root, right?

hieradata/nodes/lethe.yaml
abizer: the reason I had to put it in hiera is because of this: https://puppet.com/docs/puppet/5.3/lang_classes.html#include-like-behavior

> Include-like behavior relies on external data and defaults for class parameter values, which allows the external data source to act like cascading configuration files for all of your classes.

otherwise I'd need to do 
```puppet
ocf::ssl::default {
  owner => 'root',
}
```

I'm not attached to either approach

modules/ocf/manifests/ssl/setup.pp
keur: @jvperrin would making ldap a hard requirement, eg `provider => ldap`, prevent this?

modules/ocf/manifests/ssl/default.pp
```diff
@@ -1,6 +1,10 @@
...
+){
```
keur: nit: add a space

modules/ocf/manifests/ssl/setup.pp
jvperrin: Hmmm, it might do, I haven't looked into `provider`s, but that could work

modules/ocf/manifests/ssl/lets_encrypt/dns.pp
jvperrin: This should be, but it's not great, I think we can just change this to use the constructed bundle made by concat and that'll fix all of this cert stuff essentially, since the bundle file at least has the correct permissions.

modules/ocf_backups/manifests/pgsql.pp
```diff
@@ -0,0 +1,14 @@
+class ocf_backups::pgsql {
+  include ocf::packages::postgres
+
+  file {
+    '~/.pgpass':
+      source    => 'puppet:///private/backups/.pgpass',
```
abizer: this will actually need to be done in the `ocf_postgres` module, as this will add it locally on hal instead of on the remote machine

ja5087: in ocf_postgres should this be linked to '/opt/share/.pgpass' then? since it's the home folder of ocfbackups on lethe

abizer: in `ocf_postgres/manifests/init.pp` you probably want to add a rule like

```puppet
file {
  '/opt/share/.pgpass': 
    source => 'puppet:///private/pgpass',
    mode   => '0400',
    owner  => 'ocfbackups',
    show_diff => false,
}
```
or something like that, with some explanatory comments

modules/ocf_postgres/manifests/init.pp
```diff
@@ -0,0 +1,44 @@
...
+    postgres_password => hiera('postgres::rootpw'),
+    # https://www.postgresql.org/docs/current/static/auth-pg-hba-conf.html
+    ipv4acls          => ['hostssl sameuser all 0.0.0.0/0 md5'],
+    ipv6acls          => ['hostssl sameuser all ::/0 md5'],
+  }
+
+  file {
+    # copies proper .pgpass file for ocfbackups to authenticate on backup
+    '/opt/share/.pgpass':
+      source    => 'puppet:///private/backups/.pgpass',
```
kpengboy: Note that the correct path for the source should actually be `puppet:///private/backups/pgpass` (without the dot).

modules/ocf/manifests/ssl/setup.pp
abizer: I think it may be easiest to just say `forcelocal => false`? the ldap puppet provider appears to require information on connecting to LDAP in puppet.conf.

modules/ocf/manifests/ssl/bundle.pp
abizer: while it works for the certs itself, unfortunately, the key file has the same restrictive rules, so managing the files themselves still seems necessary :(

PR #369, opened 2018-09-14
Conf builder

keur:
So basically we have passwords duplicated all over our puppet private share. This is not great when one needs to rotate passwords or add a new host/password. Instead of that, we should have each password once in `private.yaml`, pull that in, and create config files from a template.

This is just one of the possible implementations. The nice thing about this is it creates a config file on the fly and resolves all the passwords in hiera. The downside is that it's constrained only to Windows INI format (good for most our python things) and could become a bit verbose. A positive is that it the `layout` hash is self-describing and will give staffers perusing our code more insight into the infrastructure.

For those not on root staff, I've provided and example using the `slackbridge.conf` that gets put into the slackbridge docker container. It will look like this:

```
[irc]
nickserv_pass=password

[slack]
token=password
user=password
```

If we want to proceed with this, I'll move the other docker secrets here along with the create and ircbot configs. One thing I should mention is that I think we should use this sparingly—only when there is serious password duplication going on.

EDIT: No mesos agents are on my environment, but I've tested this and it does work.

### DIFF
--- a/modules/ocf/functions/lookup_wrap.pp
+++ b/modules/ocf/functions/lookup_wrap.pp
@@ -1,8 +1,6 @@
 function ocf::lookup_wrap($item) {
-  if String($item) =~ /^[a-zA-Z_:]*$/ {
-    lookup(String($item))
-  }
-  else {
-    $item
+  $item ? {
+    String  => lookup($item),
+    default => $item,
   }
 }
--- a/modules/ocf/functions/lookup_wrap.pp
+++ b/modules/ocf/functions/lookup_wrap.pp
@@ -1,0 +1,8 @@
+function ocf::lookup_wrap($item) {
+  if $item =~ /^[a-zA-Z_:]*$/ {
+    lookup($item)
+  }
+  else {
+    $item
+  }
+}
--- a/modules/ocf/functions/lookup_wrap.pp
+++ b/modules/ocf/functions/lookup_wrap.pp
@@ -1,6 +1,6 @@
 function ocf::lookup_wrap($item) {
-  if $item =~ /^[a-zA-Z_:]*$/ {
-    lookup($item)
+  if String($item) =~ /^[a-zA-Z_:]*$/ {
+    lookup(String($item))
   }
   else {
     $item
--- a/modules/ocf/manifests/conf.pp
+++ b/modules/ocf/manifests/conf.pp
@@ -3,16 +3,19 @@ define ocf::conf(
   $owner = 'root',
   $group = 'root',
   $mode = '0400',
+  $show_diff = false,
+  $force = true,
 ) {
 
   $builder = $layout.tree_each.map |$h| { [ $h[0], ocf::lookup_wrap($h[1]) ] }
   $config = Hash($builder, 'hash_tree')
 
   file { $title:
-    ensure  => file,
-    content => template('ocf/conf.erb'),
-    mode    => $mode,
-    owner   => $owner,
-    group   => $group,
+    ensure    => file,
+    content   => template('ocf/conf.erb'),
+    mode      => $mode,
+    owner     => $owner,
+    group     => $group,
+    show_diff => $show_diff,
   }
 }
--- a/modules/ocf/manifests/conf.pp
+++ b/modules/ocf/manifests/conf.pp
@@ -1,0 +1,18 @@
+define ocf::conf(
+  $layout = {},
+  $owner = 'root',
+  $group = 'root',
+  $mode = '0400',
+) {
+
+  $builder = $layout.tree_each.map |$h| { [ $h[0], ocf::lookup_wrap($h[1]) ] }
+  $config = Hash($builder, 'hash_tree')
+
+  file { $title:
+    ensure  => file,
+    content => template('ocf/conf.erb'),
+    mode    => $mode,
+    owner   => $owner,
+    group   => $group,
+  }
+}
--- a/modules/ocf/manifests/configbuilder.pp
+++ b/modules/ocf/manifests/configbuilder.pp
@@ -17,5 +17,6 @@ define ocf::configbuilder(
     owner     => $owner,
     group     => $group,
     show_diff => $show_diff,
+    force     => $force,
   }
 }
--- a/modules/ocf/manifests/configbuilder.pp
+++ b/modules/ocf/manifests/configbuilder.pp
@@ -1,4 +1,4 @@
-define ocf::conf(
+define ocf::configbuilder(
   $layout = {},
   $owner = 'root',
   $group = 'root',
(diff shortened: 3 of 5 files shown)


modules/ocf/functions/lookup_wrap.pp
```diff
@@ -0,0 +1,8 @@
+function ocf::lookup_wrap($item) {
+  if String($item) =~ /^[a-zA-Z_:]*$/ {
```
keur: This is a little hacky. I'd like to fail harder here since we are dealing with passwords. It's a little tricky because all the values that get passed in are hashes, even the keys.

kpengboy: I would match on type instead:
```puppet
function ocf::lookup_wrap($item) {
  $item ? {
    String  => lookup($item),
    default => $item,
  }
}
```

modules/ocf/templates/conf.erb
```diff
@@ -0,0 +1,6 @@
+<% @config.each do |section, passwords| -%>
+[<%= section %>]
+<% passwords.each do |title, value| -%>
+<%= title %>=<%= value %>
+<% end -%>
```
kpengboy: A blank line separating sections would be a little more human-friendly.

keur:
@kpengboy, hmm. Not a bad idea. The biggest issue with a `fileserver.conf` is now docker containers are being mounted with passwords they don't even use.

kpengboy:
@keur we would only need to mount in passwords that are actually used.

modules/ocf/manifests/conf.pp
```diff
@@ -0,0 +1,21 @@
+define ocf::conf(
```
abizer: I would name this something more descriptive, perhaps "ocf::configbuilder" or something like that?

keur:
> s/f/fig/

got it :P

modules/ocf/manifests/conf.pp
```diff
@@ -0,0 +1,21 @@
...
+  $layout = {},
```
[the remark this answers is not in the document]
keur: I don't like `content` because that has it's own meaning in puppet... What about `structure`?

modules/ocf/manifests/configbuilder.pp
```diff
@@ -0,0 +1,22 @@
+define ocf::configbuilder(
+  $layout = {},
+  $owner = 'root',
+  $group = 'root',
+  $mode = '0400',
+  $show_diff = false,
+  $force = true,
+) {
+
+  $builder = $layout.tree_each.map |$h| { [ $h[0], ocf::lookup_wrap($h[1]) ] }
```
jvperrin: Would it be possible to just have the layout that is passed in already have this format? Enforcing it has a type of `Hash[String, Hash[String, String]]` I believe should do the trick.

modules/ocf/manifests/configbuilder.pp
```diff
@@ -0,0 +1,22 @@
+define ocf::configbuilder(
```
jvperrin: Maybe `config_builder` instead given that's the convention we (generally) use for other class names (underscores to separate words).

We do have some exceptions like `ocf::hostkeys`, `ocf::autologout`, or `ocf::crondeny`, but our newer stuff is generally named with underscore separators (like manifests in `ocf_mesos`, `ocf::firewall`, and `ocf::ssl`)

modules/ocf/manifests/configbuilder.pp
```diff
@@ -0,0 +1,22 @@
...
+  $layout = {},
+  $owner = 'root',
+  $group = 'root',
+  $mode = '0400',
+  $show_diff = false,
+  $force = true,
```
jvperrin: I think adding something here to validate passwords would be good, because we currently validate redis (and other) passwords in [a bunch](https://github.com/ocf/puppet/blob/785371b1461b7d69541fa95d92dc2e523502d0ad/modules/ocf_admin/manifests/create/redis.pp#L17) of [different](https://github.com/ocf/puppet/blob/785371b1461b7d69541fa95d92dc2e523502d0ad/modules/ocf_broker/manifests/init.pp#L18) [places](https://github.com/ocf/puppet/blob/785371b1461b7d69541fa95d92dc2e523502d0ad/modules/ocf_ocfweb/manifests/dev_config.pp#L7-L10). It would be better to have somewhere more unified to validate things coming from hiera if we have some restriction on what the value can actually be.

modules/ocf/functions/lookup_wrap.pp
```diff
@@ -0,0 +1,6 @@
+function ocf::lookup_wrap($item) {
+  $item ? {
+    String  => lookup($item),
```
jvperrin: I'm a little concerned that this doesn't allow for string literals, because I feel like those would be pretty useful to be able to use. Maybe use `lookup()` with a fallback of the item itself instead, so like `lookup($item, {'default_value': $item})`? Also, arrays can be passed as the argument to `lookup`, so leaving open support for that would be cool.

My main concern with the alternative I've mentioned is that it would be easy to remove some thing from hiera and have it fail silently, so there's some balance to be struck here between usability (actually being able to use strings) and reliability/predictability (not having this silently break with hiera changes).

Thoughts?

modules/ocf_mesos/manifests/secrets.pp
```diff
@@ -33,6 +33,20 @@
...
+    layout => {
+      'irc'   => {
+        'nickserv_pass' => 'ocf::slackbridge::nickserv_pass',
+      },
+      'slack' => {
+        'token' => 'ocf::slackbridge::token',
+        'user'  => 'ocf::slackbridge::user',
+      },
+    },
+  }
```
jvperrin: Nice, this looks pretty clean

modules/ocf/functions/lookup_wrap.pp
jvperrin: Actually, the regex you had here before was decent, because then you can try to check if it's a hiera-like value or not. Maybe we can make a custom type ([type alias](https://puppet.com/docs/puppet/5.5/lang_type_aliases.html) for a regex) or something? That seems like it could be pretty neat

kpengboy: I am -1 on guessing whether an `$item` is supposed to be a hiera password or literal string. It seems like there would be too many edge cases. Better would be to have the user call `lookup` manually.

jvperrin: I'd agree on that, guessing seems like not the way to go, and adding `lookup` when needed seems much simpler. Then this wrapper wouldn't even be needed either, which I think would be nicer overall.

kpengboy: Well it'd still be helpful to have a wrapper to verify password validity nonetheless

PR #373, opened 2018-09-24
show firefox quit confirmation dialog

cg505:
~Note: this is untested since I don't have root and can't run puppet-trigger. If anyone wants to test you can switch to my env (cooperc).~


cg505:
Tested and working.

PR #374, opened 2018-09-25
Modify enforcer to always read TEACOPIES

dkess:
This should fix our current print copy woes (ie the bug where people can print "infinite" pages by increasing the amount of "copies"). Based on my tests, the "page number is already multiplied by copy count" comment is inaccurate.


dkess:
Will test this tonight after-hours. If all goes well, I will merge it.

PR #375, opened 2018-10-02
only check for vms when we actually have libvirtd running

cg505:
Avoids the current puppetspam about vms failing when virsh is installed but the system daemon isn't actually running.

Would be good to test on a machine that has vms and a machine that doesn't.

### DIFF
--- a/modules/ocf_kvm/facts.d/vms
+++ b/modules/ocf_kvm/facts.d/vms
@@ -2,6 +2,7 @@
 import json
 import os.path
 import subprocess
+import sys
 
 
 def run_command(command_args):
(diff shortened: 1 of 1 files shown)


cg505:
The check given seems to more robust, for instance in the case that the socket is actually in some other directory. On the other hand, checking for the file would probably be faster.

modules/ocf_kvm/facts.d/vms
```diff
@@ -11,7 +11,9 @@ def run_command(command_args):
     return list(filter(None, output))
 
 
-if __name__ == '__main__' and os.path.isfile('/usr/bin/virsh'):
+if (__name__ == '__main__' and
+    os.path.isfile('/usr/bin/virsh') and
+        subprocess.call(('/bin/systemctl', 'is-active', '--quiet', 'libvirtd')) == 0):
```
jvperrin: Indentation here seems kinda strange, I'd probably do something like this, with the added benefit that you can add comments about why these checks actually exist:

```
if __name__ == '__main__':
    # Only check for VMs if virsh is actually available
    if not os.path.isfile('/usr/bin/virsh'):
        return

    # Don't check for VMs if libvirtd isn't running
    if subprocess.call(('/bin/systemctl', 'is-active', '--quiet', 'libvirtd')) != 0:
        return

    [...]
```

cg505: Not sure what's up with the indentation to be honest, it was automatically done by flake8 iirc. I like you approach better.

PR #376, opened 2018-10-03
Add sourcegraph service

jvperrin:
I'm currently running `jaws` and `monsoon` on my puppet env to get sourcegraph working (and sourcegraph is pinned to `jaws`), but once this is merged I should be able to unpin that and have it run anywhere with the correct directory for temporary data.

I also think we should set up the python code intelligence server as another service on port 10013: https://about.sourcegraph.com/docs/code-intelligence/install-manual#python

### DIFF
--- a/modules/ocf_mesos/manifests/secrets.pp
+++ b/modules/ocf_mesos/manifests/secrets.pp
@@ -31,6 +31,16 @@ class ocf_mesos::secrets {
       purge     => true,
       force     => true,
       show_diff => false;
+
+    # Create a couple scratch directories for sourcegraph to use for temporary
+    # data storage
+    #
+    # This isn't great, but I think it's better than creating a whole new repo
+    # for sourcegraph and this temporary directory only stores session
+    # information and cloned repos anyway
+    ['/opt/share/docker/sourcegraph', '/opt/share/docker/sourcegraph/redis']:
+      ensure => directory,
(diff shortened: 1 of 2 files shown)


modules/ocf_mesos/manifests/secrets.pp
```diff
@@ -31,6 +31,16 @@
...
       force     => true,
       show_diff => false;
+
+    # Create a couple scratch directories for sourcegraph to use for temporary
+    # data storage
+    #
+    # This isn't great, but I think it's better than creating a whole new repo
+    # for sourcegraph and this temporary directory only stores session
+    # information and cloned repos anyway
+    ['/opt/share/docker/sourcegraph', '/opt/share/docker/sourcegraph/redis']:
```
abizer: does souregraph run its own redis in the container? can it use gridlock?

jvperrin: It does run its own redis, we could maybe use gridlock, but sourcegraph doesn't seem to support redis authentication from what I can tell: https://about.sourcegraph.com/docs/config/external-database/#using-your-own-redis-server-with-sourcegraph (and it only stores ephemeral data)

abizer: I see. Probably not worth the effort, then.

PR #378, opened 2018-10-04
remove follow_symlinks option from sshfs

cg505:


cg505:
We can mount public_html at the linked directory, or we could just mount public_html in the home directory. Either way should be somewhat trivial. Also, if you really like symlinks to be followed and find it useful, you can just add another mountpoint with the option enabled from .desktoprc.

abizer:
we as staff can certainly do some desktoprc to re-enable symlinks, but the average user, probably not. it would probably be better to use desktoprc to mount a new sshfs that doesn't follow_symlinks for your own use rather than changing it globally unless a suitable solution for `public_html` can be implemented at the same time as making this change.

cg505:
To be clear, I think it is important for public_html to work regardless of what you do or do not put in .desktoprc. I was just referring to the case where the use of other symlinks was also desired. @abizer are you okay with separately mounting the public_html directory at the link destination as a solution?

kpengboy:
NB: we prefer rebasing feature branches here over merging in master

cg505:
will rebase

cg505:
After discussion with abizer, I am closing this PR. The problem is that there is no way to use sshfs to mount ~/public_html on its own from tsunami, since it is a symlink. It is also not possible to mount the symlink destination directly, using pam_mount, because there's no way to construct to the `/u/user` pattern without patching. We could add a manual call to mount.fuse to the login initialization scripts, but this is a non-trivial change, and better for a separate PR if desired. For now, I will just add my own sshfs mount without this option.

PR #379, opened 2018-10-04
add VS Code to desktops

cg505:


modules/ocf/manifests/packages/vscode.pp
```diff
@@ -0,0 +1,7 @@
+class ocf::packages::vscode {
+  class { 'ocf::packages::vscode::apt':
```
abizer: perhaps this is a good opportunity to refactor the way we include third-party repositories in our code, instead of duplicating this pattern multiple times for every type of package we want to include from an upstream source instead of the debian repositories.

maybe we can make a single `ocf::packages::third_party` with all the `apt::source` declarations and declared as `stage => first`, and then just add the `code` (and others) packages to `ocf_desktop::packages`.

keur: rip vscode is really the time refactor :((

kpengboy: Agree. The one drawback is that there would be increased potential for forgetting to remove unneeded third party repositories.

cg505: Agree, but this is fodder for a separate PR.

modules/ocf/manifests/packages/vscode/apt.pp
```diff
@@ -0,0 +1,18 @@
...
+  }
+
+  apt::source { 'vscode':
+    architecture => 'amd64',
+    location     => 'http://packages.microsoft.com/repos/vscode',
+    release      => 'stable',
+    repos        => 'main',
+    include      => {
+      src => false
+    },
```
asottile: isn't this the default?

modules/ocf/manifests/packages/vscode/apt.pp
```diff
@@ -0,0 +1,18 @@
+# Include VS Code apt repo.
+class ocf::packages::vscode::apt {
+  apt::key { 'vscode':
+    id     => 'BC528686B50D79E339D3721CEB3E94ADBE1229CF',
+    server => 'keyserver.ubuntu.com',
+  }
+
+  apt::source { 'vscode':
+    architecture => 'amd64',
```
asottile: doesn't actually matter here but you can do:

```puppet
    architecture => $::architecture,
```

cg505: I think this makes more sense since they only package for amd64 (in case we ever get RISC V desktops lmao)

PR #382, opened 2018-10-05
Add saslauthd

dkess:
This PR adds saslauth and enables the plain authentication mechanism. It follows the instructions in https://security.stackexchange.com/a/69031 and lays the groundwork for LDAP authentication integrations in services like Grafana, Discourse, and potentially many others.

This lets LDAP auth pass passwords to Kerberos.

### DIFF
--- a/modules/ocf_ldap/files/sasl2-slapd
+++ b/modules/ocf_ldap/files/sasl2-slapd
@@ -1,2 +1,2 @@
-mech_list:      GSSAPI plain
+mech_list:      GSSAPI
 pwcheck_method: saslauthd
--- a/modules/ocf_ldap/files/sasl2-slapd
+++ b/modules/ocf_ldap/files/sasl2-slapd
@@ -1,2 +1,2 @@
-mech_list:      GSSAPI
+mech_list:      GSSAPI plain
 pwcheck_method: saslauthd
--- a/modules/ocf_ldap/manifests/init.pp
+++ b/modules/ocf_ldap/manifests/init.pp
@@ -120,4 +120,17 @@ class ocf_ldap {
         action => 'accept',
       };
   }
+
+  package { ['sasl2-bin']:; }
+  augeas { '/etc/default/saslauthd':
+    context => '/files/etc/default/saslauthd',
+    changes => [
+      'set MECHANISMS \'"kerberos5"\'',
+      'set START \'"yes"\'',
+    ],
+    require => Package['sasl2-bin'],
+  }
+  service { 'saslauthd':
+    ensure => running,
+  }
 }
--- a/modules/ocf_ldap/manifests/init.pp
+++ b/modules/ocf_ldap/manifests/init.pp
@@ -130,7 +130,4 @@ class ocf_ldap {
     ],
     require => Package['sasl2-bin'],
   }
-  service { 'saslauthd':
-    ensure => running,
-  }
 }


modules/ocf_ldap/manifests/init.pp
```diff
@@ -120,4 +120,17 @@
...
+  augeas { '/etc/default/saslauthd':
+    context => '/files/etc/default/saslauthd',
+    changes => [
+      'set MECHANISMS \'"kerberos5"\'',
+      'set START \'"yes"\'',
+    ],
+    require => Package['sasl2-bin'],
+  }
+  service { 'saslauthd':
+    ensure => running,
```
jvperrin: Generally this is an antipattern with puppet, instead you'd want to just make sure the service is enabled to start on boot and leave it at that. Otherwise you can get pretty long puppet runs or the system won't work on boot unless puppet has run, stuff like that

modules/ocf_ldap/files/sasl2-slapd
```diff
@@ -1,2 +1,2 @@
-mech_list:      GSSAPI
+mech_list:      GSSAPI plain
```
jvperrin: I'm still not convinced this is the correct way to do this, since I feel like passwords should get passed through using SASL to Kerberos without this if they're set up correctly, and having plaintext passwords seems kinda scary.

modules/ocf_ldap/manifests/init.pp
dkess: I think it's safe to delete this line, since the `START="yes"` line in defaults should make the service start on boot.

modules/ocf_ldap/files/sasl2-slapd
dkess: I don't believe this actually stores passwords in plaintext, or in a second place at all. The "plaintext" password is getting passed to saslauthd, which is configured to send it to Kerberos. I don't think this is any more insecure than sending passwords to death over https, which is what the rest of our services do.

chriskuehl:
Do we really need this? Why can't we just use basic auth instead of SASL auth? I really don't think we need to support *three* different ways to auth to LDAP...

dkess:
The problem with basic auth for Grafana is that it makes it impossible to add an anonymous user. Also, not all applications have clean basic auth support. Discourse currently does it in a somewhat hacky way.

I am not sure what you mean by three ways. Right now, the only way to "authenticate" is to talk to Kerberos, this simply gives LDAP the ability to talk to Kerberos as well.

chriskuehl:
LDAP can already talk to kerberos on your behalf, see rt#458. Or has something changed on that front?

chriskuehl:
The test command from that ticket still seems to work (`ldapsearch -x -D uid=ckuehl,ou=People,dc=OCF,dc=Berkeley,dc=EDU -W uid=ckuehl`).

dkess:
That test command only works because firestorm has been on my puppet environment, which contains this change. I've reverted it back to production, and it no longer works.

daradib:
Personally I like basic auth because failed auth attempts never hit the application, which may be more buggy than the reverse proxy handling basic auth (nginx). Of course that prevents anonymous access by design. And it doesn't work for non-HTTP services.

Kerberos authentication doesn't perform authorization to deny sorried users. I don't think LDAP passthrough does/did either?

Alternatively SASL auth with `MECHANISMS="pam"` would follow the pam_access rules to deny sorried users, but you wouldn't be able to use gssapi with pam, only plain.

daradib:
Also, typically Grafana proxies requests to the backend datasource (influxdb, prometheus), so the anonymous user can use this http endpoint (e.g., `/api/datasources/proxy/1/query`) to run all sorts of queries, regardless of what dashboards are visible to them. So don't keep sensitive info in the database and use read-only credentials.

dkess:
I'm not so worried about sorried users, since I expect most services that use this to be staff-only (such as Grafana). If we put Mastodon on this, I would be willing to revisit the question of sorrying, since that would include all OCF users.

With regard to Grafana and Prometheus endpoints-- right now we don't have any access control over Prometheus queries. However that question is unrelated to Grafana and LDAP auth. The point of allowing staffers to log in as a user (instead of read-only anonymous) is to let them create and experiment with dashboards.

modules/ocf_ldap/files/sasl2-slapd
dkess: Just did some more investigation and you are actually right-- this line is not needed. I will delete it.

dkess:
After doing some investigation, it looks like plain auth was not actually necessary for LDAP passthrough. I've removed that line. However, saslauthd seems to be necessary-- perhaps this daemon (or something with equivalent functionality) was installed and running on the firestorm of 2013, and was lost when the LDAP configuration changed format (we no longer have a `slapd.conf` since the fields originally in that file are now in LDAP itself).

Either way, the test LDAP command and the Grafana LDAP plugin seem to work while the service is running, and don't work when I manually stop the service.

dkess:
Gonna merge this now, but if anyone has more concerns about security please reach out to me on IRC. I definitely want to make sure this is done right, and at this point I'm reasonably confident that this is being done correctly.

PR #384, opened 2018-10-08
let lightning access puppet-ca api endpoint

abizer:
after upgrading to puppet 6, this helpful message appeared in the logs
when trying to run `puppetserver ca list`:

```
2018-10-07 18:17:58,784 ERROR [qtp1785629719-71] [p.t.a.rules] Forbidden request: lightning.ocf.berkeley.edu(2607:f140:8801:0:0:0:1:21) access to /puppet-ca/v1/certificate_statuses/any_key (method :get) (authenticated: true) denied by rule 'puppetlabs deny all'.
```

this additional rule should help fix that by adding:

```
{
    "allow": [
        "lightning.ocf.berkeley.edu"
    ],
    "match-request": {
        "method": [
            "get",
            "post",
	    "put",
	    "delete",
        ],
        "path": "^/puppet-ca/v1/(?:certificate|certificate_status)",
        "query-params": {},
        "type": "regex"
    },
    "name": "allow-puppetserver-cli",
    "sort-order": 998
}
```
to `/etc/puppetlabs/puppetserver/conf.d/auth.conf`

### DIFF
--- a/modules/ocf_puppet/manifests/puppetserver.pp
+++ b/modules/ocf_puppet/manifests/puppetserver.pp
@@ -1,4 +1,4 @@
-class ocf_puppet::puppetmaster {
+class ocf_puppet::puppetserver {
   package {
     ['puppetserver', 'puppet-lint', 'augeas-tools']:;
   }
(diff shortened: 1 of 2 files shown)


modules/ocf_puppet/manifests/puppetserver.pp
```diff
@@ -35,6 +35,17 @@
     notify               => Service['puppetserver'],
   }
 
+  puppet_authorization::rule { 'allow-puppetserver-cli':
```
jvperrin: I'd add a comment above this that the point here is for the puppet master to be able to access puppet information about itself so that `puppetserver ca list` works (hence the `$::fqdn` in `allow`), since that wasn't clear to me when first reading this and I think it could confuse others too.

modules/ocf_puppet/manifests/puppetserver.pp
```diff
@@ -35,6 +35,17 @@
...
+    match_request_path   => '/puppet-ca/v1/(?:certificate|certificate_status)',
+    match_request_type   => 'regex',
+    match_request_method => ['get', 'post', 'put', 'delete'],
+    allow                => $::fqdn,
+    sort_order           => 998,
```
jvperrin: Why does this need to be specified? The default is `200`, does this need to be after the default?

abizer: otherwise, earlier default rules to allow unauthenticated access to the /certificate endpoint match first, which breaks either the puppetserver or the client depending on the flavor of the request

jvperrin: Ok, that makes sense, this is fine then

PR #385, opened 2018-10-09
add keyboard shortcut for whiskermenu

glee-:
relevant code on line 22

Allows a user to press super to open whiskermenu, mimicking the user experience of Windows a bit more.


glee-:
thanks keur. slimmed the file down to 16 lines of code.

abizer:
smh keur taking away github contribution karma from gleeb

cg505:
This PR breaks Super-\<arrow key\> shortcuts for moving windows around. @jvperrin actually had this binding in the past but [removed it](https://github.com/ocf/puppet/commit/fd590d6a4427efc2da11b7a452ab19851412910e). I don't use default bindings anyway, so this doesn't affect me, but I think we should revert.

PR #386, opened 2018-10-09
Split up modules and add alertmanager

dkess:
This PR splits the Prometheus module into the "server" and "alertmanager." It also adds the alertmanager, with a working configuration that sends emails to the mon@ mailing list for "low priority" alerts.

It also adds a basic LowToner alert, which replicates the current cronjob script.


modules/ocf_prometheus/manifests/alertmanager.pp
```diff
@@ -0,0 +1,40 @@
+class ocf_prometheus::alertmanager {
+  ocf::firewall::firewall46 {
+    '995 allow prometheus alertmanager to send on SMTP port':
+      opts   => {
+        chain       => 'PUPPET-OUTPUT',
+        proto       => 'tcp',
+        destination => ['anthrax'],
```
abizer: we shouldn't hardcode the server's names if we can use the service CNAMEs instead

modules/ocf_prometheus/manifests/server.pp
```diff
@@ -0,0 +1,152 @@
...
+
+  cron { 'gen-prometheus-printers':
+    command => '/usr/local/bin/gen-prometheus-printers /var/local/prometheus-printers.json',
+    minute  => '0',
+    require => File['/usr/local/bin/gen-prometheus-printers'];
+  }
+
+  exec { 'gen-prometheus-printers-initial':
+    command => '/usr/local/bin/gen-prometheus-printers /var/local/prometheus-printers.json',
+    creates => '/var/local/prometheus-printers.json',
```
abizer: this task seems common enough that a better pattern than this would exist for it - is there a way we can consolidate these `exec`s in a nicer way?

modules/ocf_prometheus/manifests/alertmanager.pp
```diff
@@ -0,0 +1,40 @@
+class ocf_prometheus::alertmanager {
+  ocf::firewall::firewall46 {
+    '995 allow prometheus alertmanager to send on SMTP port':
```
kpengboy: We may want to consider consolidating this rule and the one in `ocf_jenkins`.

modules/ocf_prometheus/manifests/alertmanager.pp
```diff
@@ -0,0 +1,40 @@
...
+      opts   => {
+        chain       => 'PUPPET-OUTPUT',
+        proto       => 'tcp',
+        destination => ['anthrax'],
+        dport       => 25,
+        uid         => 'alertmanager',
```
kpengboy: This depends on the `alertmanager` user as written.

modules/ocf_prometheus/manifests/server.pp
kpengboy: Make a defined resource?

modules/ocf_prometheus/manifests/alertmanager.pp
keur: i also don't think an array is necessary here?

modules/ocf_prometheus/manifests/server.pp
keur: Shouldn't the cron job have a dependency on the initial script running first?

keur: also why does the cron job require the `/usr/local/bin/gen-prometheus-printers` but not the `exec` command?

keur: Something like this seems a lot more sane. I disagree that a `defined` type is necessary unless this pattern repeats in even more places. So far two places isn't enough to warrant the abstraction.

```
file {
    '/usr/local/bin/gen-prometheus-printers':
      source => 'puppet:///modules/ocf_prometheus/gen-prometheus-printers',
      mode   => '0755';
  } ->
 exec { 'gen-prometheus-printers-initial':
    command => '/usr/local/bin/gen-prometheus-printers /var/local/prometheus-printers.json',
    creates => '/var/local/prometheus-printers.json',
  } ->
 cron { 'gen-prometheus-printers':
    command => '/usr/local/bin/gen-prometheus-printers /var/local/prometheus-printers.json',
    minute  => '0',
  }
```

modules/ocf_prometheus/manifests/server.pp
```diff
@@ -0,0 +1,152 @@
...
+      },
+      {
+        'job_name'        => 'node',
+        'scrape_interval' => '10s',
+        'scrape_timeout'  => '5s',
+
+        'file_sd_configs' => [
+          {
+            'files'            => [ '/var/local/prometheus-nodes.json' ],
+            'refresh_interval' => '1h',
```
keur: refresh interval of an hour here, but a cron job that updates the file every minute..?

modules/ocf_prometheus/manifests/server.pp
```diff
@@ -0,0 +1,152 @@
...
+      },
+      {
+        'job_name'        => 'printer',
+        'scrape_interval' => '30s',
+        'scrape_timeout'  => '5s',
+
+        'file_sd_configs' => [
+          {
+            'files'            => [ '/var/local/prometheus-printers.json' ],
+            'refresh_interval' => '1h',
```
keur: ibid

modules/ocf_prometheus/manifests/alertmanager.pp
```diff
@@ -0,0 +1,40 @@
...
+        action      => 'accept',
+      },
+      before => undef,
+  }
+
+  class { '::prometheus::alertmanager':
+    version       => '0.15.2',
+
+    extra_options => "--cluster.advertise-address=${::ipaddress}:9094 --web.external-url=\"https://prometheus.ocf.berkeley.edu/alertmanager\"",
```
keur: wow this is ugly. i don't blame you, but the designers of this public module.

modules/ocf_prometheus/manifests/server.pp
```diff
@@ -0,0 +1,152 @@
...
+      }
+    ]
+  }
+
+  $cname = $::host_env ? {
+    'dev'  => 'dev-prometheus',
+    'prod' => 'prometheus',
+  }
+
+  apache::vhost {
```
keur: maybe move this to it's own puppet file?

modules/ocf_prometheus/manifests/alertmanager.pp
dkess: The `alertmanager` user is created by puppet-prometheus. Is it necessary to re-declare it here? I'd rather not write something that conflicts with what is written there.

modules/ocf_prometheus/manifests/server.pp
dkess: If this pattern ever repeats a third time (which may actually happen in the future) I'd be willing to look into ways to cut down on the repetition. But for now I'll just take keur's suggestion of using the arrows.

modules/ocf_prometheus/manifests/alertmanager.pp
dkess: Is it really necessary to add more layers of abstraction here? I don't think this repetition is so bad, especially since it's clear what's happening here.

dkess:
Sorry for the delay on this. Thanks to everyone for the feedback, I think I've responded to all of it. I also made some changes to the printer alerts-- the toner alert now looks at a sliding median (which will smooth out the bug where printers sometimes report -2 toner), and I've added a new alert that fires when printers go offline.

modules/ocf_prometheus/manifests/alertmanager.pp
kpengboy: It should have a Requires

dkess: Does it matter if this runs before the `alertmanager` user is created though? I don't think it would cause an error.

kpengboy: I think it would, because there would be no UID to which to map that user.

dkess: Fair point, I've added the requires.

PR #388, opened 2018-10-10
Stopping the keyring prompt

dongkyunc:
If I see "Enter your password to unlock your keyring" one more time istfg I'm going to go nuts


dkess:
We might need to reboot desktops for this to take effect, but it definitely works.  Merging

PR #392, opened 2018-10-13
migrate NFS to new filehost server dataloss

abizer:
also make it a mesos agent (for create) and a ntp peer

### DIFF
--- a/hieradata/nodes/dataloss.yaml
+++ b/hieradata/nodes/dataloss.yaml
@@ -3,14 +3,14 @@ classes:
   - ocf_mesos::slave
   - ocf_mesos::secrets
 
-ocf_filehost::storage_device: '/dev/md/nfs'
-
 ocf::packages::ntp::master: true
 ocf::packages::ntp::peers:
   - hal.ocf.berkeley.edu
   - pandemic.ocf.berkeley.edu
   - jaws.ocf.berkeley.edu
   - riptide.ocf.berkeley.edu
+
+ocf_filehost::storage_device: '/dev/md/nfs'
 
 ocf_mesos::slave::attributes:
   nfs:
--- a/hieradata/nodes/dataloss.yaml
+++ b/hieradata/nodes/dataloss.yaml
@@ -1,0 +1,17 @@
+classes:
+  - ocf_filehost
+  - ocf_mesos::slave
+  - ocf_mesos::secrets
+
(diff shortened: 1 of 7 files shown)


modules/ocf_filehost/manifests/init.pp
```diff
@@ -1,11 +1,43 @@
...
+      ensure  => directory;
   }
 
+  mount {
+    '/opt/homes':
+      ensure  => mounted,
+      atboot  => true,
+      device  => $storage_device,
+      fstype  => 'ext4',
+      options => 'noacl,noatime,nodev,usrquota';
```
kpengboy: If we need to mount NFS read only, will this automatically change mount options?

modules/ocf_filehost/manifests/init.pp
```diff
@@ -1,11 +1,43 @@
...
+
+    '/home':
```
kpengboy: You might want to add an explicit dependency on `Mount['/opt/homes']` here. Otherwise you might find it bind-mounted the wrong directory.

hieradata/nodes/dataloss.yaml
```diff
@@ -0,0 +1,17 @@
+classes:
+  - ocf_filehost
+  - ocf_mesos::slave
+  - ocf_mesos::secrets
+
+ocf_filehost::storage_device: '/dev/md/nfs'
```
kpengboy: Nit: order this after the `ocf` settings?

modules/ocf_filehost/manifests/init.pp
```diff
@@ -1,11 +1,43 @@
...
+    '/etc/exports':
+      source  => 'puppet:///modules/ocf_filehost/exports',
+      require => Package['nfs-kernel-server'];
+
+    '/opt/homes':
+      ensure  => directory;
   }
 
+  mount {
+    '/opt/homes':
```
kpengboy: This needs to depend on `File['/opt/homes']`.

modules/ocf_filehost/manifests/init.pp
abizer: we'd probably mount NFS ro on the clients, right? not on the host?

kpengboy: Looks like in any case puppet won't remount ro -> rw, so this point is moot.

PR #393, opened 2018-10-14
Backport nvidia drivers for HPC from stretch so they install properly.

redplanks:


modules/ocf_hpc/manifests/compute.pp
```diff
@@ -2,7 +2,10 @@
   require ocf_hpc
 
   # install proprietary nvidia drivers
-  package { ['nvidia-driver', 'nvidia-settings', 'nvidia-cuda-toolkit']:; }
+  ocf::repackage {
+    ['nvidia-driver', 'nvidia-settings', 'nvidia-cuda-toolkit']:
```
jvperrin: We aren't currently installing `nvidia-cuda-toolkit` from backports on `corruption` (although we could do), is that package necessary to install from backports?

redplanks: It looks like the [stretch-backports version](https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=nvidia-cuda-toolkit) is newer, so probably good to do, especially if we end up getting one of the newest chipset GPUs in the future.

redplanks: Hmm, on second thought, I think CUDA versions aren't backwards compatible across major semver numbers, so we should hold off on upgrading for now.

PR #396, opened 2018-10-14
nix ocf_rancid

### LINKED ISSUE #395: update or delete ocf_rancid now that we no longer have any Cisco devices on our network
the firewall was replaced by a Palo Alto 5260 (bsecure) and the switch was replaced with an Arista 7050SX, so we should probably either fix rancid to work with the Arista device or get rid of it entirely since we don't really need it anymore

abizer:
we no longer have any cisco devices on our network, and the arista
switches are unmanaged. this module doesn't do much anymore unless
we want to get it working with the arista switches, and are actually
going to configure them to do anything. 

closes #395


jvperrin:
(also archive https://github.com/ocf/rancid)

PR #402, opened 2018-10-20
ocf_puppet: install keychain

kpengboy:
Thoughts?


dkess:
What do we want to use this for?

jvperrin:
@kpengboy All I have for GPG right now is `export GPG_TTY=$(tty)` in my `.zshrc`, and SSH just works, although I did copy my key to my VM.

kpengboy:
@keur on supernova I believe it's already installed through `extrapackages`.

PR #404, opened 2018-10-20
Backport CUDA toolkit, and load nvidia-uvm kernel module on boot.

redplanks:


redplanks:
I'm not sure; it might have been something kuoh did, but I don't remember ever manually loading this.

modules/ocf_hpc/manifests/compute.pp
```diff
@@ -2,11 +2,12 @@
   require ocf_hpc
 
   # install proprietary nvidia drivers and CUDA.
-  ocf::repackage {
-    ['nvidia-driver', 'nvidia-settings']:
+  ocf::repackage { ['nvidia-driver', 'nvidia-settings', 'nvidia-cuda-toolkit']:
```
kpengboy: So we decided to incompatibly bump the CUDA version after all?

modules/ocf_hpc/manifests/compute.pp
```diff
@@ -2,11 +2,12 @@
...
       backport_on => stretch;
+  } -> file { '/etc/modules-load.d/nvidia-uvm.conf':
+    # The nvidia-uvm kernel module, which is needed for CUDA apps, can't be loaded as needed from within Singularity.
+    content => 'nvidia-uvm';
```
kpengboy: Make sure this content ends with an end-of-line character.

modules/ocf_hpc/manifests/compute.pp
redplanks: I tried to downgrade Nvidia drivers on corruption and ended up being forced to upgrade the cuda version :/

kpengboy: Why were you downgrading drivers?

redplanks: Tensorflow wasn't working, so I thought it might have been a bug introduced when we upgraded drivers (though it turned out to be that a kernel module wasn't loaded).

modules/ocf_hpc/manifests/compute.pp
```diff
@@ -2,11 +2,12 @@
   require ocf_hpc
 
   # install proprietary nvidia drivers and CUDA.
-  ocf::repackage {
-    ['nvidia-driver', 'nvidia-settings']:
+  ocf::repackage { ['nvidia-driver', 'nvidia-settings', 'nvidia-cuda-toolkit']:
       backport_on => stretch;
+  } -> file { '/etc/modules-load.d/nvidia-uvm.conf':
+    # The nvidia-uvm kernel module, which is needed for CUDA apps, can't be loaded as needed from within Singularity.
+    content => 'nvidia-uvm\n';
```
kpengboy: This needs to be a double quoted string, or else `\n` won't be interpreted as an escape sequence.

modules/ocf_hpc/manifests/compute.pp
kpengboy: Why can't we downgrade the CUDA version again?

redplanks: I (very unwisely) tried removing the nvidia drivers to downgrade them from the stretch-backports to the stretch version, and ended up with a bunch of broken packages; when I put everything back, I couldn't install CUDA from stretch because of package conflicts with the stretch-backports nvidia drivers. So I'm not sure, to be honest.

abizer: does it really matter if we're using CUDA from stretch-backports instead of stretch? If the newer drivers work we should just stick with them.

redplanks: Our old versions of Tensorflow are compatible with the new CUDA version (9.1), and we can upgrade to the newest versions of TF now, so I don't think we'll have any problems with using the stretch-backports version.

abizer: just upgrade everything to the latest version that works and shipit

kpengboy: But are any HPC users relying on, or having binaries compiled against, CUDA 8? If so we'll break their code.

redplanks: Since corruption is already on CUDA 9, and nobody has complained yet, probably not.

PR #406, opened 2018-10-24
Update to new home page

glee-:
Last day to register to vote in California for this election cycle was 10/22/18. We should no longer remind users to register to vote.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -21,7 +21,7 @@ mesos_slaves:
     - riptide
     - dataloss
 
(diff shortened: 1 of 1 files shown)


abizer:
we should ask them to fill out our feedback survey though - perhaps add the google form in an iframe like we do for the decal signup form.

dkess:
The original homepage was https://www.ocf.berkeley.edu/about/lab/open-source

glee-:
to be merged in with ocf/ocfweb#409

PR #407, opened 2018-10-24
Google doc error

solar464:
A stop gap measure to prevent people from printing from Google Docs until the reason for the formatting errors can be found.

Not sure if there's a better way to tell if a file is a Google Doc than inspecting the title.

### DIFF
--- a/modules/ocf_printhost/files/enforcer
+++ b/modules/ocf_printhost/files/enforcer
@@ -118,12 +118,12 @@ NOTIFY_JOB_ERROR = dedent("""\
 """)
 
 NOTIFY_PROBLEMATIC_FILE_SOURCE = dedent("""\
-        Unfortunately we don't support printing {file_source}.
-        Please download your file as a PDF, then try again.\
+        Unfortunately, printing from {file_source} results in formatting issues.
+        Please download your file as a PDF and print from desktop instead of web browser.\
 """)
 
 PROBLEMATIC_FILE_SOURCE = [
-    ('Google Doc', lambda job: job.doc_name.endswith(' - Google Docs'))
+    ('Google Docs', lambda job: job.doc_name.endswith(' - Google Docs'))
 ]
 
 
--- a/modules/ocf_printhost/files/enforcer
+++ b/modules/ocf_printhost/files/enforcer
@@ -164,6 +164,12 @@ def send_printer_mail(message, job, quo):
 
 def prehook(c, r, job):
     quo = quota.get_quota(c, job.user)
+    import inspect
+    _, _, _, argvals = inspect.getargvalues(inspect.currentframe())
+    with open('~/enforcer_debug.txt', 'w+') as f:
+        for argname in argvals:
+            f.write(argname + ': ' + argvals[argname] + '\n')
+
     if job.pages > quo.daily:
         send_printer_mail(INSUFFICIENT_QUOTA_MESSAGE, job, quo)
         msg = NOTIFY_QUOTA_MESSAGE.format(
--- a/modules/ocf_printhost/files/enforcer
+++ b/modules/ocf_printhost/files/enforcer
@@ -181,7 +181,7 @@ def prehook(c, r, job):
                 file_source=file_source,
             )
             r.publish(job.user, msg)
-            sys.exit(1)
+            sys.exit(255)
 
     if job.pages > quo.daily:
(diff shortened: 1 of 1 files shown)


modules/ocf_printhost/files/enforcer
```diff
@@ -117,6 +117,16 @@ NOTIFY_JOB_ERROR = dedent("""\
...
 """)
 
+NOTIFY_PROHIBITED_FILE_TYPE = dedent("""\
+        I'm afraid I can't let you do that.
+        Please download your file as a PDF before downloading.
+        {file_type} files disagree with my software.\
+""")
+
+PROHIBITED_FILE_TYPES = [
+    ('Google Doc', lambda job: job.doc_name[-14:] == ' - Google Docs')
```
dkess: Use the `endswith` string method

modules/ocf_printhost/files/enforcer
```diff
@@ -117,6 +117,16 @@ NOTIFY_JOB_ERROR = dedent("""\
         Please contact a staff member for assistance.\
 """)
 
+NOTIFY_PROHIBITED_FILE_TYPE = dedent("""\
+        I'm afraid I can't let you do that.
+        Please download your file as a PDF before downloading.
+        {file_type} files disagree with my software.\
```
dkess: I'd reword this to something like, "Unfortunately we don't support printing from Google Docs. Please download your file as PDF, then try again."

modules/ocf_printhost/files/enforcer
```diff
@@ -164,6 +174,16 @@ def send_printer_mail(message, job, quo):
 
 def prehook(c, r, job):
     quo = quota.get_quota(c, job.user)
+
+    # A stop gap until we figure out why some file types aren't printing correctly
+    for file_type, prohibited in PROHIBITED_FILE_TYPES:
```
dkess: Do we expect to have more "prohibited file types"? If not, this kind of abstraction isn't really necessary. Just have an if statement for gdocs.

modules/ocf_printhost/files/enforcer
solar464: Fair enough.

modules/ocf_printhost/files/enforcer
solar464: Abizer mentioned there may be more file types, I figured I'd leave it extendable just in case.

abizer: it stands to reason that Google Docs is not the only thing that fails - does the same occur for printing straight from bCourses or Gmail or something else? this amount of abstraction is hardly enterprise-ready, I think it's safe to plan for a printing-unfriendly future with this. I'd call it `PROBLEMATIC_FILE_SOURCES` or something like that though.

modules/ocf_printhost/files/enforcer
```diff
@@ -117,6 +117,15 @@ NOTIFY_JOB_ERROR = dedent("""\
         Please contact a staff member for assistance.\
 """)
 
+NOTIFY_PROHIBITED_FILE_TYPE = dedent("""\
+        Unfortunately we don't support printing {file_type}.
```
abizer: I'd just say "We've encountered problems printing documents from this source. Please try to download your document as a PDF and print it from the desktop instead of the web browser." or something like that

modules/ocf_printhost/files/enforcer
```diff
@@ -164,6 +173,16 @@ def send_printer_mail(message, job, quo):
...
     quo = quota.get_quota(c, job.user)
+
+    # A stop gap until we figure out why some file types aren't printing correctly
+    for file_source, problematic in PROBLEMATIC_FILE_SOURCE:
+        if problematic(job):
+            msg = NOTIFY_PROBLEMATIC_FILE_SOURCE.format(
+                file_source=file_source,
+            )
+            r.publish(job.user, msg)
+            sys.exit(255)
```
kpengboy: Nit: some exit codes have special meanings. You may wish to consider picking one which would not clash with any of these.

Context: https://www.tldp.org/LDP/abs/html/exitcodes.html#EXITCODESREF

kpengboy:
How long does it take for this notification to pop up after the user prints something?

solar464:
When I tested it, the error notification usually pops up within a a second or two.

cg505:
Is there something keeping this from getting merged?

keur:
This does not work.

PR #410, opened 2018-10-26
retire privileges of alumni staff

abizer:
thank you for your service. if you need these
principals again please reach out to us and we
can add them back.


gnowxilef:
wow, dara is still that active?

PR #411, opened 2018-10-26
Enable TCP BBR congestion control

daradib:
BBR improves throughput over lossy connections, so it would be
useful at least on user-facing servers like death and tsunami.
Linux 4.9+ supports BBR. Our machines run Linux from stretch or
jessie-backports, so they should all support BBR.

### DIFF
--- a/modules/ocf/manifests/networking.pp
+++ b/modules/ocf/manifests/networking.pp
@@ -76,6 +76,14 @@ class ocf::networking(
       require => Package['resolvconf'];
   }
 
+  # Enable TCP BBR congestion control
+  sysctl {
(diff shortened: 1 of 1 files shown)


daradib:
I would suggest testing this out first to see whether it has any undesirable side effects, and if so, we could limit it to death and tsunami. Within our network, we could also try increasing the MTU (jumbo frames), but that's a story for another day.

abizer:
I've put death and tsunami on one of my environments with these changes, let's see what happens

daradib:
Testing in prod! What can possibly go wrong.

abizer:
nothing seems to have gone wrong in the last 6 days

PR #412, opened 2018-10-26
Puppet MUNGE authentication, needed for SLURM.

redplanks:
I've put munge.key into the private share folders for corruption and segfault.

### DIFF
--- a/modules/ocf_hpc/manifests/init.pp
+++ b/modules/ocf_hpc/manifests/init.pp
@@ -41,14 +41,14 @@ class ocf_hpc {
 
   # SLURM uses MUNGE for authentication. Each host needs to have
   # the same munge.key, and have synced clocks.
-  package { 'munge': }
-  -> file { '/etc/munge/munge.key':
+  package { 'munge': } ->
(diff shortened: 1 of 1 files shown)


redplanks:
I did the syntax like `} -> next {` in compute.pp and controller.pp though.

redplanks:
Bump. Which style should it be? In addition, the Puppet [style guide](https://puppet.com/docs/puppet/5.5/style_guide.html#chaining-arrow-syntax) recommends
```
Package['httpd']
-> Service['httpd']
```

abizer:
+1 for following the style guide

jvperrin:
It looks like it [does have a check for that](http://puppet-lint.com/checks/arrow_on_right_operand_line/) and that we have [disabled it](https://github.com/ocf/puppet/blob/2a06900abd36daa8049b4e586cd2c4150101f2fd/.pre-commit-config.yaml#L21). Some discussion about the disable is in https://github.com/ocf/puppet/pull/173.

redplanks:
I'll change the arrow style to match the rest of the repository then.

PR #414, opened 2018-10-30
Allows staff members to highlight themselves on the labmap

NotRyan:
This PR adds a button to the panel of staff desktops that allows staff to highlight their position on the lab map. The default state of this is off. An example of what this looks like is shown below.

desktop
labmap

The intent of this feature is to connect average users who might need assistance with something beyond the scope of opstaff, and staff members in the lab who are interested in helping out. Participation by staff is entirely optional. 

On a technical level, this adds a button to the xfce panel which calls a python script at /opt/stats/staff_visibility.py, which manages the changing icon state and calls /opt/stats/update_flags.sh, which sends an update to https://www.ocf.berkeley.edu/api/session/log, which has been updated to accept this extra data. This updates the ocfstats mysql database. The ocfstats mysql database has had its schema altered to add a 'flags' column to the 'session' table, as well as the corresponding views. Another api page was added to ocfweb at https://www.ocf.berkeley.edu/api/lab/staff, which will display the desktops of any staff who have flagged themselves with the button, with data taken from the ocfstats database. The labmap this api and colors the appropriate desktops purple. 

Let me know if I went against conventions for anything in the code. Also, if someone has a better way to layout the lab map text to display the information, or if you want a different color, feel free to make or suggest the alterations. I'm pretty bad at anything involving UI. All suggestions and criticisms are welcome.


abizer:
Hey rrchan,

This is an impressive effort! I have a few concerns, though. This is a lot of code and behavior changes for a relatively simple feature - would it perhaps be simpler to just always mark location of staff instead of trying to build a visible/invisible flag system? Especially since we already publicly show which staff members are in the lab and at what desktops anyways on ocf.io/stats. Building this flags system is a lot of added complexity if we could instead merely mark everyone coming out of `staff_in_lab_public` as purple on the display.

dkess:
If staffers are busy and not on staff hours, they might not want to be approached by people with questions. So it makes sense that staffers would want to manually turn this on only if they choose.

abizer:
Even if that's the case, a much simpler opt-in system would be putting a text file in ~staff/public_html with a list of usernames of staff who welcome questions, and perhaps a script (or the existing panel button, if necessary) that just does something like

```bash
add="sed -i '/$USER/d' ~staff/lab_visibility"
remove="echo $USER >> ~staff/lab_visibility"

if [ "$1" = "visible" ]; then
  ssh admin "$remove && $add"
else
  ssh admin "$remove"
fi
```
(or something roughly along those lines)

and having ocfweb read that text file like it currently reads hours_temp or staff_hours.yaml

dkess:
I'm definitely not opposed to adding in this kind of functionality, but this should be a slow process of adding in different pieces, rather than one giant set of PRs all going out at once. For example, it looks like you've modified ocfweb to make API calls to your development instance of ocfweb. Instead of having to do that, the API should be finalized first, then you can start working on calling out to the API.

There are a lot of different ways to handle state storage. I'm not sure MySQL + ocfweb API is the right way to do it at all since it's adding a fair amount of complexity. @abizer and I just had a long discussion about this in #rebuild so this is definitely a larger discussion.

I'm also not sure how the panel works. There's probably a better way to handle this that doesn't involve running a script to modify a config file but I'd have to look into it more.

Though once again, thanks for this-- I appreciate the work you've done for a really cool feature.

NotRyan:
> Hey rrchan,
> 
> This is an impressive effort! I have a few concerns, though. This is a lot of code and behavior changes for a relatively simple feature - would it perhaps be simpler to just always mark location of staff instead of trying to build a visible/invisible flag system? Especially since we already publicly show which staff members are in the lab and at what desktops anyways on ocf.io/stats. Building this flags system is a lot of added complexity if we could instead merely mark everyone coming out of `staff_in_lab_public` as purple on the display.

Thanks for the feedback Abizer. This was my initial idea, but staffers I'd spoken to about it told me that people would be opposed to an always-on thing like that, since many staffers use the lab to do their own work or study.

> Even if that's the case, a much simpler opt-in system would be putting a text file in ~staff (symlinked into public_html) with a list of usernames of staff who welcome questions, and perhaps a script (or the existing panel button, if necessary) that just does something like...

Yeah, that would definitely be a simpler solution. I did it through the ocfstats mysql database since the existing labmap works through that, and I figured it'd be best to keep all the lab stats, which I'd consider this to be part of, in the same place,.

NotRyan:
> I'm definitely not opposed to adding in this kind of functionality, but this should be a slow process of adding in different pieces, rather than one giant set of PRs all going out at once. For example, it looks like you've modified ocfweb to make API calls to your development instance of ocfweb. Instead of having to do that, the API should be finalized first, then you can start working on calling out to the API.

You're right, I probably should have atomized this.

> There are a lot of different ways to handle state storage. I'm not sure MySQL + ocfweb API is the right way to do it at all since it's adding a fair amount of complexity. @abizer and I just had a long discussion about this in #rebuild so this is definitely a larger discussion.

I've elaborated upon my reasoning in an earlier post. You both have far more experience with these kinds of things than I do; I trust whatever you guys settle on.

> I'm also not sure how the panel works. There's probably a better way to handle this that doesn't involve running a script to modify a config file but I'd have to look into it more.
I certainly hope there is. My panel solution feels kind of hacky, but I found xfce's panel and its documentation to be very limited in scope, so this is what I came up with.

abizer:
Sure, it's effectively the same thing.

keur:
@dkess @abizer an optin text file is surely a downgrade from what I see here... I think rrchan has made the right call here. Though there is a `staff_in_lab_public` database which would be a more reasonable place to add this flag. Further, we could have `ocflib` and `create` obey the same functionality.

modules/ocf_desktop/files/skel/config/xfce4/panel/launcher-21/switch.desktop
```diff
@@ -0,0 +1,14 @@
...
+Icon=user-invisible
+StartupNotify=false
+Terminal=false
+Categories=Utility;X-XFCE;X-Xfce-Toplevel;
+OnlyShowIn=XFCE;
+Name=Currently Invisible
+Comment=Toggles your visibility as staff on the lab map
+X-XFCE-Source=file:///usr/share/applications/xfce4-about.desktop
+Visibility=Invisible
+Path=
```
keur: Try removing this line, should work the same.

modules/ocf_desktop/files/skel/config/xfce4/panel/launcher-21/switch.desktop
```diff
@@ -0,0 +1,14 @@
+[Desktop Entry]
+Version=1.0
```
keur: jumping straight to 1.0. bold

modules/ocf_desktop/manifests/stats.pp
```diff
@@ -24,6 +24,16 @@
...
       owner  => ocfstats,
       source => 'puppet:///modules/ocf_desktop/stats/update-delay.sh';
+      
+    '/opt/stats/update-flags.sh':
+      mode   => '0555',
+      owner  => ocfstats,
+      source => 'puppet:///modules/ocf_desktop/stats/update-flags.sh';
+      
+    '/opt/stats/staff-visibility.py':
+      mode   => '0555',
```
keur: `o+x` and `g+x` seems like a mistake here. Say i'm lazy and antisocial (which i am), and someone comes in fix their wordpress. Since im lazy and antisocial, I dont want to help. I notice `rrchan` is in the lab, studying hard for a midterm and has his visibility toggled off. I then `ssh` to your computer and toggle you on, and now you have to help fix some broken wordpress install and fail your midterm miserably.

The correct perms should be `544`.

modules/ocf_desktop/files/stats/update-flags.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# state can be "active" "cleanup"
+# "active" implies a current logged-in session
+# "cleanup" triggers closing of the session
+# other files to look at: session-cleanup and session-setup in /etc/lightdm
+
+CUR_USER=$(who | awk '$NF == "(:0)" { print $1 }')
```
keur: Some impressive `awk`ing, but `whoami` is much simpler here

modules/ocf_desktop/files/stats/update-flags.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
keur: We usually have these flags set in our bash scripts
```bash
set -euo pipefail
```

modules/ocf_desktop/manifests/stats.pp
```diff
@@ -24,6 +24,16 @@
       mode   => '0555',
       owner  => ocfstats,
       source => 'puppet:///modules/ocf_desktop/stats/update-delay.sh';
+      
+    '/opt/stats/update-flags.sh':
+      mode   => '0555',
```
keur: `ibid` on permissions. especially here you use `who` instead of `whoami` in the which further hurts the AC

modules/ocf_desktop/files/stats/staff-visibility.py
```diff
@@ -0,0 +1,27 @@
...
+import os
+import re
+
+LAUNCHER_SETTINGS_PATH = os.path.expanduser('~/.config/xfce4/panel/launcher-21/switch.desktop')
+UPDATE_FLAGS_SCRIPT_PATH = os.path.expanduser('/opt/stats/update-flags.sh')
+
+def changePanelState(file_path, icon_name, visibility):    
+    with open(file_path, 'r') as f:
+        content = f.read()
+    content_1 = re.sub(r'((?<=Icon=).*)', icon_name, content)
```
keur: Looks good. It's a shame that `configparser` won't work on regular `ini` files because they don't have sections.

abizer:
`staff_in_lab_public` is a view, so it's immutable. I definitely disagree that adding a new column to the database and mutating a field there is a good solution - especially in a table that is designed to store statistics, in an overwhelmingly append-only manner. If we wanted to retain the use of the database in implementing this feature, it would probably be better to make a new table with (username, is_visible) and have the panel upsert into that table rather than change the existing table. And that basically reduces down to a text file with usernames in it, except without any of the complexity of trying to connect to mysql. Using the database to carry this kind of minute state is definitely the wrong way to go here.

modules/ocf_desktop/files/stats/update-flags.sh
abizer: the original script runs as the `ocfstats` user, so `whoami` wouldn't work. which is why we parse the user logged into the primary display for stats.

keur:
@abizer with your script that ssh's into supernova there's all sorts of race conditions and locking not taken into account that SQL will handle for you.

dkess:
There's an easy way to handle this that doesn't lead to locking issues/cronjobs:

* Make an `available` folder in `~staff/public_html`
* When logging in, remove `~staff/public_html/available/user`
* Toggle the indicator by adding/removing the `~staff/public_html/available/user` file
* If you want access control, turn the `user` file into a directory with restricted permissions

However after some reflection, I suppose using the database for this isn't the worst idea. But this isn't really related to labstats so we should probably make a separate table for it.

keur:
@dkess, your solution is better than one file. but i don't see what benefits this has over a database. actually, after some reflection myself, what about storing this in `redis` instead of `SQL`. It seems suited for something like this.

kpengboy:
My single two cents' worth is this: the system should be opt-out instead of opt-in, and opt-outs should not be possible during your staff hour. After all, one of the basic duties of staff is helping OCF members with use of our services. If staff have to opt in in order to be displayed, I can confidently predict that hardly anyone will ever remember to opt in.

PR #417, opened 2018-11-01
Change paper-genmon text

glee-:
After reading some initial lab survey results, I decided to open up this pr.
Results: https://i.fluffy.cc/zqf3xPJvKFfR9kKF7mFfKFgn6zv9g4xX.png

As we can see in the upper right hand corner, this is quite a big visual change compared to before.

### DIFF
--- a/modules/ocf_desktop/files/xsession/paper-genmon
+++ b/modules/ocf_desktop/files/xsession/paper-genmon
@@ -24,9 +24,11 @@ def main(argv=None):
             semester_quota = quota.semesterly
 
     print('<img>{}</img>'.format(IMAGE_PATH))
-    print('<txt> {} {} left</txt>'.format(daily_quota,
-                                          'page' if daily_quota == 1 else 'pages'))
-    print('<tool>{} remaining this semester</tool>'.format(semester_quota))
+    print('<txt> {} {} left today | {} {} left this semester </txt>'.format(daily_quota,
+                                                                            'page' if daily_quota == 1 else 'pages',
+                                                                            semester_quota,
+                                                                            'page' if semester_quota == 1 else 'pages'
+                                                                            ))
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


modules/ocf_desktop/files/xsession/paper-genmon
```diff
@@ -24,9 +24,11 @@ def main(argv=None):
             semester_quota = quota.semesterly
 
     print('<img>{}</img>'.format(IMAGE_PATH))
-    print('<txt> {} {} left</txt>'.format(daily_quota,
-                                          'page' if daily_quota == 1 else 'pages'))
-    print('<tool>{} remaining this semester</tool>'.format(semester_quota))
+    print('<txt> {} {} left today | {} {} left this semester </txt>'.format(daily_quota,
```
abizer: I'd do something like
```python
PANEL_TEXT="<txt {} {} left today | {} {} left this semester </txt>"
print(PANEL_TEXT.format(...))
```
so the things aren't all the way to the right

dkess:
Maybe change "left this semester" to "semester" so it doesn't take up so much space? Otherwise lgtm though.

modules/ocf_desktop/files/xsession/paper-genmon
```diff
@@ -24,9 +24,11 @@ def main(argv=None):
...
+                                                                            'page' if daily_quota == 1 else 'pages',
+                                                                            semester_quota,
+                                                                            'page' if semester_quota == 1 else 'pages'
+                                                                            ))
```
keur: nit: fix this whitespace

abizer:
I actually like that this information is very visible in the status bar since you added this panel last year - it has grown on me, and especially with the upcoming theme change, I think we can tastefully add more information to the taskbar and it would be beneficial to the users without being 'clutter' - for example, displaying today's lab hours in the task bar as well.

PR #418, opened 2018-11-08
Remove rt-ops list

dkess:
I'm almost certain this isn't used anywhere, since I've repurposed the `operations` RT queue to automatically email the right people.

After this is merged, I'll also delete this group from Google Admin.


dkess:
Don't remember exactly what I did but it was something along those lines.

PR #420, opened 2018-11-09
Add CUPS monitoring and alerting

dkess:
This lays the groundwork for monitoring CUPS, and adds some important alerts.

It works by periodically writing to a metrics file on whiteout, and including that file in `node_exporter` via the textfile collector.

We alert on inconsistent configurations of CUPS printer rotations. I think I was able to cover all erroneous cases, though it's possible I'm missing something. See the mon@ mailing list for examples of these in action.

You'll notice the cron->creates->file pattern is back. This is at least the third time it appears, so suggestions are welcome on how to cut down this repetition.

### DIFF
--- a/modules/ocf_printhost/files/monitor-cups
+++ b/modules/ocf_printhost/files/monitor-cups
@@ -25,6 +25,7 @@ def main():
     conn = cups.Connection()
     for cups_class, printers in conn.getClasses().items():
         for printer in printers:
+            # class is a reserved keyword, so we have to pass it via dictionary
             classes.labels(**{'class': cups_class, 'printer': printer}).set(1)
 
     write_to_textfile(sys.argv[1], registry)
--- a/modules/ocf_printhost/files/monitor-cups
+++ b/modules/ocf_printhost/files/monitor-cups
@@ -1,0 +1,33 @@
+#!/usr/bin/env python3
+"""
+Collects metrics from CUPS and writes them to a text file in the Prometheus
+metrics format.
+"""
+
+import sys
+
+import cups
+from prometheus_client import CollectorRegistry
+from prometheus_client import Gauge
+from prometheus_client import write_to_textfile
+
+
+def main():
+    registry = CollectorRegistry()
+
+    classes = Gauge(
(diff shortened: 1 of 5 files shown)


modules/ocf_printhost/files/monitor-cups
```diff
@@ -0,0 +1,33 @@
...
+        'cups_class',
+        'Existence of printer on CUPS class',
+        ['class', 'printer'],
+        registry=registry,
+    )
+
+    conn = cups.Connection()
+    for cups_class, printers in conn.getClasses().items():
+        for printer in printers:
+            classes.labels(**{'class': cups_class, 'printer': printer}).set(1)
```
abizer:
```suggestion
            classes.labels(class=cups_class, printer=printer).set(1)
```
?

modules/ocf_printhost/manifests/monitor.pp
```diff
@@ -0,0 +1,25 @@
...
+      source => 'puppet:///modules/ocf_printhost/monitor-cups',
+      mode   => '0755';
+  } ->
+  exec { 'monitor-cups-initial':
+    command => '/usr/local/bin/monitor-cups /var/local/prometheus/cups.prom',
+    creates => '/var/local/prometheus/cups.prom',
+  } ->
+  cron { 'monitor-cups':
+    command => '/usr/local/bin/monitor-cups /var/local/prometheus/cups.prom',
+    # Run every minute
```
abizer: is this the default?

modules/ocf_printhost/files/monitor-cups
dkess: Unfortunately `class` is a reserved term, so this causes a syntax error. I can add a comment for this.

modules/ocf_printhost/manifests/monitor.pp
dkess: Yup, not specifying a field sets it to `*`, so if we don't specify any it defaults to every minute.

dkess:
Going to merge this now and make a separate PR for your suggestion.

PR #423, opened 2018-11-14
rename ocfdev to ocfapphost

abizer:
"ocfdev" is kind of a bad name for this group, it implies it's some kind of development thing when in reality it's an enrollment mechanism for apphosting. So, we should rename the group. If this gets merged I'll rename the cn/dn in LDAP as well.

### DIFF
--- a/hieradata/nodes/vampires.yaml
+++ b/hieradata/nodes/vampires.yaml
@@ -1,6 +1,6 @@
 classes:
     - ocf_apphost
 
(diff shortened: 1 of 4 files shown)


abizer:
I think the rename should be atomic - from my reading it's just piping an LDIF into ldapmodify with `changetype: moddn`, I think

abizer:
I made the duplicate group - `ldapsearch -x can=ocfapphost`

PR #424, opened 2018-11-15
Bootstrap kubernetes master and worker nodes

keur:
Using the puppetlabs-kubernetes module, add functionality to
boostrap a host as either a kubernetes master or a kubernetes
worker node. Note that adding additional master nodes will require
running puppetlabs kubetool on an OCF host to generate new etcd
keys and certs, and the generated file must be added to the
puppet private share in the kubernetes directory.

With this configuration we have successfully deployed the OCF
templates container.


modules/ocf_kubernetes/manifests/worker.pp
```diff
@@ -0,0 +1,19 @@
...
+# is necessary, so worker nodes can be trivially scaled
+# with the small requirement that kubernetes::worker
+# class is added.
+class ocf_kubernetes::worker {
+  include ocf::packages::docker_kube
+  include ocf::packages::kubernetes
+
+  class { 'kubernetes':
+    worker        => true,
+    # puppetlabs-kubernetes is a little confused here
```
abizer: confused about what?

modules/ocf/manifests/packages/docker_kube.pp
```diff
@@ -0,0 +1,10 @@
+class ocf::packages::docker_kube {
```
abizer: I'd spell out the full name

modules/ocf_kubernetes/manifests/worker.pp
keur: good catch. this comment was erroneously copied.

hieradata/common.yaml
```diff
@@ -64,7 +64,12 @@ kubernetes::kubernetes_package_version: 1.12.2-00
...
 kubernetes::etcd_version: 3.2.25
 kubernetes::container_runtime: docker
+# This is the file used to bootstrap flannel, which facilitates cluster networking.
+# It is one of the first things deployed by the master via kubectl create.
 kubernetes::cni_network_provider: https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml
+# This is the virtual network that flannel uses by default.
+# Changing this would require also deploying a modified version of kube-flannel.yml,
+# as it is hardcoded in the version they distribute.
 kubernetes::cni_pod_cidr: 10.244.0.0/16
 kubernetes::etcd_initial_cluster: hozer-75=https://169.229.226.75:2380
```
dkess: We should probably name this VM

modules/ocf/manifests/packages/kubernetes/apt.pp
```diff
@@ -0,0 +1,13 @@
...
+  apt::key { 'kubernetes':
+    id      => '54A647F9048D5688D7DA2ABE6A030B21BA07F4FB',
+    source  => 'https://packages.cloud.google.com/apt/doc/apt-key.gpg',
+    require => Package['apt-transport-https'],
+  }
+
+  apt::source { 'kubernetes':
+    location => 'https://apt.kubernetes.io',
+    repos    => 'main',
+    release  => 'kubernetes-xenial',
```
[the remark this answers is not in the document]
keur: True... I really should've checked the `dists`. I will fix this (in the next iteration of kubernetes features)[#427]

PR #427, opened 2018-11-25
Kubernetes ingress

keur:


modules/ocf_kubernetes/files/ingress_vhost.yaml
```diff
@@ -0,0 +1,13 @@
+# Add Ingress Rules to route by Host Header
```
keur: Forgot about this. This file should not be here... These should be in a separate repo with our kubernetes manifests. That said, I'll leave it here for now since it's a good reference to see what the `ingress-nginx` controller does.

keur:
Maybe the loadbalancer and ingress stuff should go in its own directory?

modules/ocf_kubernetes/manifests/master.pp
```diff
@@ -41,4 +39,14 @@
...
   }
+
+  class { 'ocf_kubernetes::ingress':
+    require => Class['kubernetes'];
+  }
+
+  file {
+    '/etc/profile.d/kubeconfig.sh':
+      mode    => '0755',
+      content => 'export KUBECONFIG=/etc/kubernetes/admin.conf';
```
kpengboy: Should have EOL character

modules/ocf_kubernetes/manifests/ingress.pp
```diff
@@ -0,0 +1,32 @@
...
+      ensure => directory,
+      mode   => '0700';
+    '/etc/kubernetes/manifests/ingress/ingress-expose.yaml':
+      content => template('ocf_kubernetes/ingress/ingress_expose.yaml.erb'),
+      mode    => '0644';
+  }
+
+  # Add ingress-nginx to the cluster
+  exec { 'init-ingress':
+    command => 'kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/mandatory.yaml',
```
kpengboy: You will want to make sure this is only executed once

modules/ocf_kubernetes/manifests/ingress.pp
```diff
@@ -0,0 +1,32 @@
+class ocf_kubernetes::ingress {
+
+  $kubernetes_workers_ipv4 = lookup('kubernetes::workers_ipv4')
+
+  file {
+    default:
+      require =>  Package['kubeadm', 'kubectl'];
+    '/etc/kubernetes/manifests/ingress':
```
kpengboy: Please add blank lines

modules/ocf_kubernetes/manifests/loadbalancer.pp
```diff
@@ -0,0 +1,38 @@
...
+  haproxy::backend { 'kubernetes-backend':
+    options => {
+      option         => [
+        'forwardfor'
+      ],
+      'mode'         => 'http',
+      'balance'      => 'roundrobin',
+      'http-request' => 'set-header X-Forwarded-Port %[dst_port]',
+      # TODO: this does nothing until we add TLS
+      'http-request' => 'add-header X-Forwarded-Proto https if { ssl_fc }',
```
kpengboy: Is this not going to clobber the entry above?

modules/ocf_kubernetes/manifests/ingress.pp
keur: Kubernetes will ignore it if the manifest doesn't change.

dkess: Still I think we should pick a tag instead of running directly from their master branch

modules/ocf_kubernetes/manifests/loadbalancer.pp
```diff
@@ -0,0 +1,38 @@
...
+    }
+  } ->
+  haproxy::balancermember { 'kubernetes-ingress-worker':
+    listening_service => 'kubernetes-frontend',
+    ports             => ['80'],
+    ipaddresses       => $kubernetes_workers_ipv4,
```
dkess: I wonder if there's a way to do this without having to manually specify all the workers. Could we use Kubernetes internal IPs in combination with some sort of service discovery mechanism to know which ones are up?

keur: Not that I currently know of. A puppet `fact` could also work, but I don't see anything horribly wrong with having the worker IPs in an array.

modules/ocf_kubernetes/manifests/ingress.pp
```diff
@@ -0,0 +1,32 @@
...
+  # Add ingress-nginx to the cluster
+  exec { 'init-ingress':
+    command => 'kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/mandatory.yaml',
+    require => Package['kubectl'];
+  }
+
+  # Set up a NodePort service so all kubernetes workers
+  # are running an instance of ingress-nginx.
+  exec { 'expose-ingress':
+    command   => 'kubectl apply -f /etc/kubernetes/manifests/ingress/ingress-expose.yaml',
```
dkess: Are you sure this will actually run? Does it need the `KUBECONFIG` environment variable?

modules/ocf_kubernetes/templates/ingress/ingress_expose.yaml.erb
```diff
@@ -0,0 +1,28 @@
...
+    - <%= ip %>
+    <% end %>
+  ports:
+    - name: http
+      port: 80
+      targetPort: 80
+      protocol: TCP
+    - name: https
+      port: 443
+      targetPort: 443
```
dkess: If we're doing HTTPs through haproxy (which we probably should), then this isn't necessary.

modules/ocf_kubernetes/manifests/loadbalancer.pp
keur: One sets the `X-Forwarded-Port` header while the other adds the `X-Forwarded-Proto` header. They are different headers so I don't think anything gets clobbered? Tbh `X-Forwarded-Port` isn't very important right now because all our services are HTTP.

keur:
Also, I will move kubernetes workers out of `common.yaml` in a future PR.

modules/ocf_kubernetes/manifests/ingress.pp
```diff
@@ -0,0 +1,33 @@
...
+      mode   => '0700';
+
+    '/etc/kubernetes/manifests/ingress/ingress-expose.yaml':
+      content => template('ocf_kubernetes/ingress/ingress_expose.yaml.erb'),
+      mode    => '0644';
+  }
+
+  # Add ingress-nginx to the cluster
+  exec { 'init-ingress':
+    command => "kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/nginx-${nginx_version}/deploy/mandatory.yaml",
```
abizer: for clarity perhaps split this into 
```
command => 'kubectl apply -f ${ingress_nginx_url}`
```
and define `$ingress_nginx_url` below `$nginx_version`?

modules/ocf_kubernetes/manifests/package/first_stage.pp
```diff
@@ -1,4 +1,4 @@
-class ocf::packages::kubernetes::apt {
+class ocf_kubernetes::package::first_stage {
```
abizer: I'd switch back, `apt` indicates better than it has to do with installing the package than `first_stage`

modules/ocf_kubernetes/templates/ingress/ingress_expose.yaml.erb
```diff
@@ -0,0 +1,29 @@
...
+  name: ingress-nginx
+  namespace: ingress-nginx
+  labels:
+    app.kubernetes.io/name: ingress-nginx
+    app.kubernetes.io/part-of: ingress-nginx
+spec:
+  type: NodePort
+  externalIPs:
+    <% @kubernetes_workers_ipv4.each do |ip| -%>
+  - <%= ip %>
```
abizer: perhaps this should be indented one level?

modules/ocf_kubernetes/manifests/loadbalancer.pp
```diff
@@ -0,0 +1,38 @@
...
+      #'redirect' =>  'scheme https code 301 if !{ ssl_fc }',
+    };
+  } ->
+  haproxy::backend { 'kubernetes-backend':
+    options => {
+      option         => [
+        'forwardfor'
+      ],
+      'mode'         => 'http',
+      'balance'      => 'roundrobin',
```
abizer: will this work with sessions?

modules/ocf_kubernetes/manifests/loadbalancer.pp
```diff
@@ -0,0 +1,38 @@
+class ocf_kubernetes::loadbalancer {
+  include ::haproxy
+
+  $kubernetes_workers_ipv4 = lookup('kubernetes::workers_ipv4')
+  $kubernetes_worker_nodes = lookup('kubernetes::worker_nodes')
+
+  haproxy::frontend { 'kubernetes-frontend':
+    mode    => 'http',
+    bind    => {
+      "${::ipaddress}:80"   => [],
```
jvperrin: Can this just bind to all IPs instead so it listens on IPv6 too? Looks like using `0.0.0.0` or just leaving it empty makes it listen on all which I think is what you want here.

hieradata/common.yaml
```diff
@@ -79,3 +79,10 @@ kubernetes::kube_api_advertise_address: "%{::ipaddress}"
 kubernetes::api_server_count: 1
 kubernetes::install_dashboard: true
 kubernetes::controller_address: 169.229.226.75:6443
+kubernetes::nginx_version: '0.21.0'
+kubernetes::workers_ipv4:
+    - 169.229.226.73
+    - 169.229.226.74
```
jvperrin: You shouldn't need to hardcode the IPs here, you can look them up with something like this: https://github.com/ocf/puppet/commit/5133f8cf0ecf231bcef204fa1b51b53b6368bc00 (using the `ldap_attr` function since each host has its IP in LDAP)

modules/ocf_kubernetes/manifests/loadbalancer.pp
jvperrin: I do agree that updating the IPs each time is a pain, and even having a list of workers is a bit annoying, but we do that currently for the marathon/mesos setup, so I think it's probably fine to get it to a feature parity level at least.

modules/ocf_kubernetes/manifests/package/first_stage.pp
jvperrin: Yeah, I'd agree on this

modules/ocf_kubernetes/manifests/loadbalancer.pp
keur: I think so. This is just roundrobining to ingress nginx, which will handle [session affinity](https://github.com/kubernetes/ingress-nginx/tree/master/docs/examples/affinity/cookie)

hieradata/common.yaml
keur: ooooh. yay thanks

PR #430, opened 2018-12-01
add support for per-interface network config files and bonding

abizer:
with this change we replace `/etc/network/interfaces` in favor of per-interface network configuration in `/etc/network/interfaces.d`. This makes for cleaner config files and clearer network configuration logic in puppet.

Furthermore, we add a bond interface type, allowing us to take advantage of LACP on our switch, giving us 20Gb/s of in-network bandwidth that gracefully falls to 10Gb/s if one of the links fails.

this leaves us with a couple variations of network configuration, all of which I have tested:
1. default, i.e. no bond or bridge, e.g. a regular desktop or VM. Tested on hozer-77 and outbreak. 
2. bond0 as the primary interface, tested on dev-fallingrocks
3. br0 as the primary interface, tested on hozer-77
4. br0 as the primary interface with bond0 as the bridged interface: tested on hal and scurvy


modules/ocf/manifests/networking.pp
```diff
@@ -12,61 +13,75 @@
...
 
   # network configuration
   file {
     '/etc/network/interfaces':
-      content => template('ocf/networking/interfaces.erb');
+      source => 'puppet:///modules/ocf/networking/interfaces';
+    '/etc/network/interfaces.d':
+      ensure  => directory,
+      recurse => true,
+      purge   => true;
```
dkess: Are you sure `purge` will work here? The files aren't actually in the directory, they get filled in by templates elsewhere. Hopefully this won't delete then re-add them on each puppet run.

abizer: the docs mention that `purge` will remove all unmanaged files. in my testing it doesn't appear to delete and re-add so I assume it's intelligent enough to not recreate them every time.

PR #431, opened 2018-12-01
fish bangbang

Sillu129092:
Hello world! :dagger:


dkess:
Agree, don't think this is necessary. It's fine to modify your personal config, but let's keep the defaults minimal.

PR #433, opened 2018-12-04
Run apt update on every puppet run.

redplanks:


modules/ocf/manifests/apt.pp
```diff
@@ -2,7 +2,13 @@
   package { ['aptitude', 'imvirt', 'apt-transport-https', 'lsb-release']:; }
 
   class { '::apt':
-    purge => {
+    # Run apt update at every Puppet run.
+    update => {
+      frequency => 'always',
+      loglevel  => 'debug',
```
abizer: we want debug logs?

redplanks: By default, the log level would be notify, which according to the docs causes puppet to report the update as a significant change.

redplanks:
Just tested firewalling off mirrors from segfault; the Puppet run took 256 seconds but was a success still.

redplanks:
Adding the primary mirrors doesn't work as intended, as apt update tries to fetch from both the primary debian mirror and fallingrocks. I can't seem to find anything on having one mirror be fallback from another, so this may not be feasible.

abizer:
it's fine if apt update pulls from both primary and secondary. it needs to know what is available at each. it should automatically fall back to the Debian mirrors if it can't pull packages from our mirrors during apt upgrade or package install, so that's how it becomes a fallback.

redplanks:
If one/both of the mirrors is down, apt update will fail, causing the rest of the puppet run to fail.

keur:
We could just add a cron job to do this once a day?

jvperrin:
Wait, so you said when our mirror was down that the run succeeded but was slow. That sounds perfectly fine to me, but when you add more mirrors as backups then it fails for some reason? Am I not understanding that correctly?

I don't think you'd want to be using ftp for the backup mirror either, http seems like a better (or more standard) choice there.

Also, @keur has a point that just having a cronjob every 15/30 minutes or something would be enough to make ocflib be updated within an hour (or half an hour with a 15 minute cronjob) of pushing if that's your aim.

jvperrin:
Bump on this.

I think going with a cronjob-based method is probably better for this? Then it's non-blocking for the puppet run at least but keeps the index updated for upgrading ocflib when puppet does run, for instance. Thoughts? Objections?

redplanks:
I think there's already a daily cronjob actually? My reason initially for this was so that ocflib hours/quota changes would be updated faster, but if we're moving away from hardcoding those in ocflib I don't think we should do this anymore, because of potential rootspam.

jvperrin:
Ok, that sounds fine to me, I'll close this for now then and it can be reopened if it's something we want to do again in the future.

PR #436, opened 2018-12-05
Add kubectl to extrapackages

dkess:
This would pull in the kubernetes apt repo on all machines that use extrapackages, but I think it's worth it since kubectl could end up being used a lot (for instance, by staffers on supernova). Thoughts?


dkess:
That's not completely true. This will be installed on Jenkins, which has a different mechanism for setting `KUBECONFIG` (PR for this is incoming).

For supernova, I'd like to have a separate way of setting up `KUBECONFIG` to fit with OCF auth.

PR #437, opened 2018-12-08
Add kubernetes jenkins auth

dkess:
* We add an authentication file (`static-tokens.csv.erb`) which contains the Jenkins token password
* We add an authorization file (`abac.jsonl`) which allows Jenkins to do anything
* We mount these in the apiserver Docker container, and configure the apiserver to use them
* As a bonus, we make the cert valid for `kubernetes.ocf.berkeley.edu`.

### DIFF
--- a/modules/ocf_kubernetes/files/abac.jsonl
+++ b/modules/ocf_kubernetes/files/abac.jsonl
@@ -1,0 +1,1 @@
+{"apiVersion": "abac.authorization.kubernetes.io/v1beta1", "kind": "Policy", "spec": {"group":"ocfdeploy",     "namespace": "*",              "resource": "*",         "apiGroup": "*"                   }}
--- a/modules/ocf_kubernetes/manifests/master.pp
+++ b/modules/ocf_kubernetes/manifests/master.pp
@@ -20,21 +20,66 @@ class ocf_kubernetes::master {
   $etcd_archive = "etcd-v${etcd_version}-linux-amd64.tar.gz"
   $etcd_source  = "https://github.com/etcd-io/etcd/releases/download/v${etcd_version}/${etcd_archive}"
 
+  class { 'ocf_kubernetes::package::first_stage':
+    stage => first,
+  }
+
+  # Passwords for the static token file
+  # https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-token-file
+  $ocf_jenkins_deploy_token = lookup('kubernetes::jenkins_token')
+
+  file {
+    '/etc/kubernetes-ocf':
+      ensure  => directory,
+      recurse => true,
+      purge   => true;
+
+    '/etc/kubernetes-ocf/static-tokens.csv':
+      content   => template('ocf_kubernetes/static-tokens.csv.erb'),
+      mode      => '0400',
+      show_diff => false;
+
+    '/etc/kubernetes-ocf/abac.jsonl':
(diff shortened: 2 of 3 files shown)


modules/ocf_kubernetes/files/abac.jsonl
```diff
@@ -0,0 +1 @@
+{"apiVersion": "abac.authorization.kubernetes.io/v1beta1", "kind": "Policy", "spec": {"group":"ocfdeploy",     "namespace": "*",              "resource": "*",         "apiGroup": "*"                   }}
```
jvperrin: Could you remove the extra whitespace in this file (or even better, format it more nicely across lines if that's possible)

modules/ocf_kubernetes/manifests/master.pp
```diff
@@ -20,21 +20,66 @@
...
+  $ocf_jenkins_deploy_token = lookup('kubernetes::jenkins_token')
+
+  file {
+    '/etc/kubernetes-ocf':
+      ensure  => directory,
+      recurse => true,
+      purge   => true;
+
+    '/etc/kubernetes-ocf/static-tokens.csv':
+      content   => template('ocf_kubernetes/static-tokens.csv.erb'),
```
jvperrin: Instead of making this a whole template, you could just make it an inline content block (like https://github.com/ocf/puppet/blob/34d1a1fe8f850dac0f008b7521d8dbaedddf09d3/modules/ocf_mesos/manifests/slave.pp#L15 or something) since it's just a line. Either way is fine, but it might just be a bit cleaner to avoid having a whole separate file for a single line of content.

I'd say this depends on if you think you'd need to add multiple lines to that file or how complicated that file is likely to get.

modules/ocf_kubernetes/manifests/master.pp
```diff
@@ -20,21 +20,66 @@
   $etcd_archive = "etcd-v${etcd_version}-linux-amd64.tar.gz"
   $etcd_source  = "https://github.com/etcd-io/etcd/releases/download/v${etcd_version}/${etcd_archive}"
 
+  class { 'ocf_kubernetes::package::first_stage':
+    stage => first,
+  }
```
jvperrin: Why was this added? What is it needed for?

modules/ocf_kubernetes/manifests/master.pp
dkess: I'd like to leave open the possibility of adding more to this file.

modules/ocf_kubernetes/files/abac.jsonl
dkess: JSONL is supposed to be a format similar to CSV, and each entry is supposed to be one line. I'll remove the whitespace though because nothing needs to line up yet.

dkess:
I wasn't sure how to add the secret token to Jenkins through Puppet, so I added it in the web interface for managing secrets. I also installed the [Kubernetes CLI](https://wiki.jenkins.io/display/JENKINS/Kubernetes+CLI+Plugin) Jenkins plugin through the web interface. Let me know if there's a better way to do this.

modules/ocf_kubernetes/manifests/master.pp
dkess: Hmm, this might be an artifact of earlier changes. I'll see if it's safe to delete this.

PR #438, opened 2018-12-08
Add ocf::systemuser resource

keur:
Setting system => true is something we have shown time and time again
that we are not able to remember, creating annoying ldap conflicts.
Having a custom resource for this will make it much more apparent. The
resource is simple. It proxies all the arguments passed to opts to the
system resource, but ensures that system is set to true and that the
user is in the group 'sys'.

### DIFF
--- a/modules/ocf/manifests/systemuser.pp
+++ b/modules/ocf/manifests/systemuser.pp
@@ -1,19 +1,20 @@
 define ocf::systemuser(
   $ensure = present,
   $opts = {},
-  $groups = [],
 ){
 
-  $groups_real = $groups ? {
-    Array   => $groups + ['sys'],
-    String  => [$groups, 'sys'],
+  $groups_real = $opts['groups'] ? {
+    Array   => $opts['groups'] + ['sys'],
+    String  => [$opts['groups'], 'sys'],
     default => ['sys'],
   }
+  $updated = { 'groups' =>  $groups_real }
+  # Rightmost hash takes precedence on merge
+  $updated_opts = $opts + $updated
 
   user { $title:
     ensure => $ensure,
     system => true,
-    groups => $groups_real,
     *      => $opts,
   }
 }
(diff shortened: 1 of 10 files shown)


modules/ocf/manifests/systemuser.pp
```diff
@@ -0,0 +1,20 @@
...
+    default => ['sys'],
+  }
+  $updated = { 'groups' =>  $groups_real }
+  # Rightmost hash takes precedence on merge
+  $updated_opts = $opts + $updated
+
+  user { $title:
+    ensure => $ensure,
+    system => true,
+    *      => $opts,
```
dkess: Should this be `updated_opts`?

modules/ocf_mail/manifests/init.pp
```diff
@@ -12,15 +12,15 @@
     subscribe => Class['ocf::ssl::default'],
   }
 
-  user { 'ocfmail':
+  ocf::systemuser { 'ocfmail':
     ensure  => present,
```
dkess: Is this line still necessary?

modules/ocf/manifests/systemuser.pp
jvperrin: This breaks when `ensure` or `system` are also passed in as part of `$opts` with something like `Server Error: Evaluation Error: The attribute 'ensure' has already been set`, even if the passed in values are the same as the ones already specified here.

Can the `ensure` just be passed directly through instead of special-cased and `system` validated with an `assert` at the top of the module and some error raised if it's set? (that or removing it from the opts passed through with `$opts - 'system'` should also work, but then it'll fail silently, which I think is probably worse for something like this which is essentially operator error when it does happen)

modules/ocf/manifests/systemuser.pp
```diff
@@ -0,0 +1,20 @@
+define ocf::systemuser(
+  $ensure = present,
+  $opts = {},
```
jvperrin: Instead of doing this, you could make it only support the attributes we actually use and it'll look a little nicer on the user end of it since you won't have to specify `opts` when using it, but that's probably not a big deal and will lead to a bunch of params being repeated in this file. Anyway, there's a trade-off to be had there.

keur:
Closing because bad. See #672

keur:
Closing because bad. See #672

PR #440, opened 2018-12-08
Kubernetes ssl

keur:


modules/ocf_kubernetes/manifests/master/loadbalancer/ssl.pp
```diff
@@ -0,0 +1,20 @@
+class ocf_kubernetes::master::loadbalancer::ssl(
+  $server_name,
+) {
+
+  $cnames = ldap_attr_multi($server_name, 'dnsCname')
+
+  $cnames.each |String $alias| {
+    $service_fqdn = "${alias}.ocf.berkeley.edu"
+    ocf::ssl::bundle { $service_fqdn:
+      domains => [$service_fqdn, "${alias}.ocf.io"],
```
abizer: does this not get handled for you by the default bundle anyways? IMO it would be preferable to not hard code .ocf.berkeley.edu or .ocf.io in a deep child class, though it's unlikely to ever change, it seems like something that should be configured at a higher level of abstraction.

modules/ocf_kubernetes/manifests/master/loadbalancer/ssl.pp
```diff
@@ -0,0 +1,20 @@
...
+    }
+  }
+
+  file { '/etc/ssl/ocf-certs.txt':
+    mode    => '0400',
+    owner   => 'haproxy',
```
keur: These won't exist until haproxy is provisioned

keur:
Anyone is encouraged to visit [templates to see that sweet sweet encryption](https://dev-templates.ocf.berkeley.edu)

modules/ocf_kubernetes/manifests/master/loadbalancer/ssl.pp
dkess: Can we make this dependent on the haproxy package?

keur: yes

PR #443, opened 2018-12-17
remove hal from mesos agents

abizer:
hal is really slow and marathon seems to time out rather often
trying to start containers on it. this causes the entire deploy to fail.
we have enough other agents for removing hal to not be an issue.

we'll also need to manually remove the configs/secrets that cause hal to register as an agent so I'll do that manually if this gets approved

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -16,7 +16,6 @@ mesos_masters:
 
 mesos_slaves:
     - jaws
-    - hal
     - pandemic
     - riptide
     - dataloss
--- a/hieradata/nodes/hal.yaml
+++ b/hieradata/nodes/hal.yaml
@@ -1,8 +1,6 @@
 classes:
     - ocf_backups
     - ocf_kvm
-    - ocf_mesos::slave
-    - ocf_mesos::secrets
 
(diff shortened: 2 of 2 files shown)


jvperrin:
Is hal really that slow? Like riptide had problems starting containers when it had the USB root drive, but I'm surprised that hal is too slow for this. That being said, hal seems to be quite slow to log in to, but I'm pretty sure that's due to [systemd-logind issues](https://serverfault.com/questions/707377/slow-ssh-login-activation-of-org-freedesktop-login1-timed-out).

jvperrin:
Yeah, having just installed a couple debugging tools on hal through apt (`sysbench` and `hdparm`), it was painfully slow to install them. I would blame file I/O, since the cpu and memory benchmarks I did weren't _that_ bad, although they were a fair amount worse than jaws, as expected.

dkess:
`hal` definitely seems to be using disk a lot-- see "Disk Utilization" in our metrics compared to `riptide`: https://grafana.ocf.berkeley.edu/d/Ckeu23piz/node-exporter-server-metrics?orgId=1&var-node=hal&var-node=riptide

Perhaps we should look into why this is the  case?

PR #445, opened 2018-12-18
Include project name in ftpsync config file.

dkess:
The ftpsync script expects the config file to be in this format in certain circumstances. It seems to have broken the Kali push sync, see rt#7886.

I believe this will fix the issue, but someone please correct me if I'm wrong. `fallingrocks` is currently on my env, and has this change. If this is merged I will manually remove the files with the old name.


dkess:
Gonna wait for the Kali Linux people to confirm that their script is working before merging this. (Or I'll merge Saturday night if we don't hear from them)

PR #446, opened 2018-12-19
Mount secrets on kubernetes workers

dkess:
This basically copies the Marathon secrets file into Kubernetes.

One departure from how this is handled in Marathon: in Marathon, we labeled certain slaves as "holding secrets" because some we didn't want some slaves (desktops) to mount them. I'm not including that here, since I don't expect us to use the desktops as Kubernetes workers. If we do end up deciding to do that, we can make a separate PR for labeling workers.

### DIFF
--- a/modules/ocf_kubernetes/manifests/secrets.pp
+++ b/modules/ocf_kubernetes/manifests/secrets.pp
@@ -1,0 +1,45 @@
+class ocf_kubernetes::secrets {
+  # The way we manage secrets in Kubernetes is that we put them in the Puppet
+  # private share under "docker", which is made available to the Kubernetes
+  # workers via a special file share, "private-docker".
+  #
+  # We then mount individual directories (e.g. for the RT service, we might
+  # mount /opt/share/docker/secrets/rt into the container).
+  #
+  # Since inside the container, our processes typically run as "nobody" (or
+  # another unprivileged user), these secrets need to be world-readable.
+  # That's fine inside Docker, but outside Docker, we don't want random users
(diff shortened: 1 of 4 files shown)


modules/ocf_kubernetes/manifests/secrets.pp
```diff
@@ -0,0 +1,45 @@
+class ocf_kubernetes::secrets {
```
keur: please create a worker directory and put this file there: `ocf_kubernetes::worker::secrets` 

Did this for the master with ingress stuff

modules/ocf_kubernetes/manifests/worker/secrets.pp
```diff
@@ -0,0 +1,45 @@
...
+  # 0644/0755 on children. Oh well.
+  #
+  # Only some Kubernetes workers have this class. Some (e.g. desktops) don't,
+  # and so won't run jobs that require secrets.
+  file {
+    '/opt/share/docker':
+      ensure => directory,
+      mode   => '0700';
+
+    '/opt/share/docker/secrets':
```
abizer: it would be nice if we could rename this directory to something more semantically meaningful, like `/opt/share/kubernetes/secrets/` or something like that.

dkess: We use the same name for Marathon. I think it's fine since it matches up with what we call the folder in the private share (which is also `docker`).

abizer:
I know, I'm suggesting we change it as we are migrating away from marathon.

PR #447, opened 2018-12-19
migrate puppetdb to managed postgres server

abizer:
my network connection is too tenuous right now for me to actually test this but it was working for a period of time during the original postgres testing run

### DIFF
--- a/modules/ocf_puppetdb/manifests/init.pp
+++ b/modules/ocf_puppetdb/manifests/init.pp
@@ -1,30 +1,18 @@
 class ocf_puppetdb {
-  class { 'puppetdb::database::postgresql':
-    # Only listen locally since puppetdb is the only database client and it is
-    # a local database
-    listen_addresses    => 'localhost',
+  class { 'puppetdb':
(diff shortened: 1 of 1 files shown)


abizer:
there's still some weirdness where Java apparently refuses to accept that the Lets Encrypt cert is trusted and it takes an inordinately long amount of time to get the server to restart with the regular on-host postgres. The java version we have and java certfile both claim to trust Lets Encrypt so I'm not sure what's going on.

ja5087:
I'm missing a lot of context on this but can anyone give a rundown of why this is better? Trying to see if it's worth finishing or we should close it eventually.

jvperrin:
I think it's just one fewer postgres server to configure, since we don't need both a local postgres on thunder and a shared one on lethe. That being said, we haven't really had any troubles that I know of with the one on thunder and it's got ephemeral data anyway, so probably not a big deal to keep it around instead.

nikhiljha:
We probably no longer need this, so closing it :)

PR #448, opened 2018-12-19
Add IRC cert reloading cronjob

dkess:
* IRC server now listens on port 6667 for localhost
* Creates a new oper user that has permission to reload certs
* Cronjob script that logs in as this oper and reloads the cert

We need to do the 6667 because cert reloading seems to not work if the operator itself is connected via SSL.

### DIFF
--- a/modules/ocf_irc/files/reload-ssl.sh
+++ b/modules/ocf_irc/files/reload-ssl.sh
@@ -1,17 +1,18 @@
 #!/bin/bash
+set -euo pipefail
 
 # Log into the IRC server and reload the SSL cert.
 
 # Use a random nick to ensure no collisions with users on the server
-username=$(pwgen --secure --no-numerals 16 1)
+nick=$(pwgen --secure --no-numerals 16 1)
 oper_pass=$(<"$1")
 
 (
-echo NICK $username
-echo USER ocfletsencrypt 0 0 :Lets Encrypt cert reloading script
+echo NICK $nick
+echo USER ocf-cert-reload 0 0 :Lets Encrypt cert reloading script
 sleep 2
-echo OPER ocfletsencrypt "$oper_pass"
+echo OPER ocf-cert-reload "$oper_pass"
 sleep 2
 echo REHASH -ssl
 echo QUIT
-) | telnet 127.0.0.1 6667
+) | nc 127.0.0.1 6667
--- a/modules/ocf_irc/files/reload-ssl.sh
+++ b/modules/ocf_irc/files/reload-ssl.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# Log into the IRC server and reload the SSL cert.
+
+# Use a random nick to ensure no collisions with users on the server
+username=$(pwgen --secure --no-numerals 16 1)
+oper_pass=$(<"$1")
+
+(
+echo NICK $username
+echo USER ocfletsencrypt 0 0 :Lets Encrypt cert reloading script
+sleep 2
+echo OPER ocfletsencrypt "$oper_pass"
+sleep 2
+echo REHASH -ssl
+echo QUIT
+) | telnet 127.0.0.1 6667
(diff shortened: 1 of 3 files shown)


modules/ocf_irc/files/reload-ssl.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
```
kpengboy: Perhaps we could consider turning on `set -u` to enforce `$1` being set.

modules/ocf_irc/files/reload-ssl.sh
```diff
@@ -0,0 +1,17 @@
...
+(
+echo NICK $username
+echo USER ocfletsencrypt 0 0 :Lets Encrypt cert reloading script
+sleep 2
+echo OPER ocfletsencrypt "$oper_pass"
+sleep 2
+echo REHASH -ssl
+echo QUIT
+) | telnet 127.0.0.1 6667
```
kpengboy: Would `nc` work here?

modules/ocf_irc/manifests/ircd.pp
```diff
@@ -6,6 +6,15 @@
     enable    => true,
     require   => Package['inspircd'],
     subscribe => Class['ocf::ssl::default'],
+  } ->
+  cron { 'reload-irc-cert':
+    # We can't use chronic here to suppress output because telnet will return an
+    # "error" code of 1 if the server disconnects, which it will for this
+    # script.
```
kpengboy: Using `nc` would avoid this issue, it seems.

modules/ocf_irc/manifests/ircd.pp
```diff
@@ -29,5 +38,14 @@
 
     '/etc/inspircd/inspircd.motd':
       source  => 'puppet:///modules/ocf_irc/ircd.motd';
+
+    '/usr/local/bin/reload-ssl.sh':
+      source => 'puppet:///modules/ocf_irc/reload-ssl.sh',
+      mode   => '0755';
+
+    '/etc/inspircd/reload_pass':
+      content   => $passwords['cert-reload-pass'],
```
kpengboy: Nit: this file should have an EOL character

modules/ocf_irc/files/reload-ssl.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Log into the IRC server and reload the SSL cert.
+
+# Use a random nick to ensure no collisions with users on the server
+username=$(pwgen --secure --no-numerals 16 1)
```
abizer: if I'm understanding this right, would it be possible to use the "ocf" admin user that appears to already exist according to inspircd.conf.erb?

dkess: I'll rename this variable to "nick" so it's more clear. If the "admin" user is online, using the ocf nick, this script will fail. So we set our nick to something random, but our "user" to the proper name, so it'll always work.

dkess:
Thanks @abizer and @kpengboy for your comments. I believe I addressed all of them, but feel free to un-resolve anything or add more.

modules/ocf_irc/manifests/ircd.pp
```diff
@@ -6,6 +6,12 @@
     enable    => true,
     require   => Package['inspircd'],
     subscribe => Class['ocf::ssl::default'],
+  } ->
+  cron { 'reload-irc-cert':
+    command => 'chronic /usr/local/bin/reload-ssl.sh /etc/inspircd/reload_pass',
+    hour    => 0,
+    minute  => 0,
```
jvperrin: I don't think this should need to run every day, we renew certs when they have 30 days left, so could this just run weekly or something? Doesn't really matter too much though

dkess: sure

PR #449, opened 2018-12-20
better check for when the letsencrypt cert needs renewal

### LINKED ISSUE #343: Retry obtaining certs if domains don't match
One problem right now with the Let's Encrypt changes (in #337) is that if new domains are added in LDAP, they likely won't be in DNS yet, a request will be made for a new cert, it'll fail, and won't be retried, since it's been [refreshed once](https://github.com/ocf/puppet/blob/06b9244683d6396a053aa9dc0b25e42b37f27491/modules/ocf/manifests/ssl/lets_encrypt/dns.pp#L27) and that's it.

If the domains requested don't match the cert, or maybe if the previous cert request failed, we should retry the cert request, since giving up isn't a good option. I'm not sure what would be the best option here, one quick and dirty trick that comes to mind is to make a lockfile of sorts, so that if the cert fails to get obtained, a future run can check if the file exists, and schedule a run. I'd hope there's something cleaner than this, maybe a custom fact that shows the domains in each cert by parsing `openssl x509 -in cert.pem -text` output?

cg505:
specifically, make sure to check that the cert that we have will also contain all the requested alternate names

also just make the cert checking code more robust in general imo

resolves #343

### DIFF
--- a/modules/ocf/facts.d/le-cert-info
+++ b/modules/ocf/facts.d/le-cert-info
@@ -1,12 +1,22 @@
 #!/usr/bin/env python3
 import json
 import os
+import sys
 from datetime import datetime
 
-from cryptography import x509
-from cryptography.hazmat.backends import default_backend
-
 LETSENCRYPT_CERT_DIR = '/var/lib/lets-encrypt/certs/'
+
+
+def print_info(data):
+    print('le_cert_info=' + json.dumps(data))
+
+
+try:
+    from cryptography import x509
+    from cryptography.hazmat.backends import default_backend
+except ImportError:
+    print_info({})
+    sys.exit()
 
 cert_data = {}
 
@@ -17,7 +27,7 @@ for title in os.listdir(LETSENCRYPT_CERT_DIR):
     except FileNotFoundError:
         continue
     except PermissionError as e:
-        print('try running this command as root or ocfletsencrypt')
+        print('try running this command as root or ocfletsencrypt', file=sys.stderr)
         raise e
 
     cert = x509.load_pem_x509_certificate(pem_data, default_backend())
@@ -27,4 +37,4 @@ for title in os.listdir(LETSENCRYPT_CERT_DIR):
         'days_to_expiration': (cert.not_valid_after - datetime.today()).days,
     }
 
-print('le_cert_info=' + json.dumps(cert_data))
+print_info(cert_data)
--- a/modules/ocf/facts.d/le-cert-info
+++ b/modules/ocf/facts.d/le-cert-info
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+import json
+import os
+from datetime import datetime
+
+from cryptography import x509
+from cryptography.hazmat.backends import default_backend
+
+LETSENCRYPT_CERT_DIR = '/var/lib/lets-encrypt/certs/'
+
+cert_data = {}
+
+for title in os.listdir(LETSENCRYPT_CERT_DIR):
+    try:
+        with open(os.path.join(LETSENCRYPT_CERT_DIR, title, 'cert.pem'), 'rb') as pem_file:
+            pem_data = pem_file.read()
+    except FileNotFoundError:
+        continue
+    except PermissionError as e:
+        print('try running this command as root or ocfletsencrypt')
+        raise e
+
+    cert = x509.load_pem_x509_certificate(pem_data, default_backend())
+    san = cert.extensions.get_extension_for_oid(x509.oid.ExtensionOID.SUBJECT_ALTERNATIVE_NAME)
+    cert_data[title] = {
+        'cert_names': san.value.get_values_for_type(x509.DNSName),
+        'days_to_expiration': (cert.not_valid_after - datetime.today()).days,
+    }
+
+print('le_cert_info=' + json.dumps(cert_data))
--- a/modules/ocf/manifests/ssl/lets_encrypt/dns.pp
+++ b/modules/ocf/manifests/ssl/lets_encrypt/dns.pp
@@ -10,24 +10,30 @@ define ocf::ssl::lets_encrypt::dns(
     content => "${join($domains, ' ')} > ${title}",
   }
 
-  # Only run the dehydrated command to renew the cert if it is old enough (30
-  # days = 2592000 seconds left until it expires). dehydrated does this check
-  # itself, but this should help with puppet error spam a bit (for instance
-  # if Let's Encrypt is down for any reason) and mean that services can
-  # subscribe to this class to reload after new certs are obtained since this
-  # won't refresh unless dehydrated is re-run for some reason.
-  exec { "check ${title} cert expiration":
-    command => '/bin/true',
-    unless  => "openssl x509 -checkend 2592000 -noout -in /var/lib/lets-encrypt/certs/${title}/cert.pem",
-    user    => ocfletsencrypt,
-  } ~>
+  # Only run the dehydrated command to renew the cert if it is old enough, or
+  # missing some domains. dehydrated does the former check itself, but this
+  # should help with puppet error spam a bit (for instance if Let's Encrypt is
+  # down for any reason) and mean that services can subscribe to this class to
+  # reload after new certs are obtained since this won't refresh unless
+  # dehydrated is re-run for some reason.
+
+  $parsed_cert_info = parsejson($::le_cert_info)
+
+  $have_cert_info = $title in $parsed_cert_info
+  if $have_cert_info {
+    $cert_expires_soon = ($parsed_cert_info[$title]['days_to_expiration'] + 0) < 30
+    # we subtract out any domains that are in the cert, and see if any are left
+    $cert_has_all_domains = ($domains - $parsed_cert_info[$title]['cert_names']) =~ Array[String, 0, 0]
+  }
+
   exec { "obtain ${title} cert":
     # This exec can be notified to get it to run dehydrated again, even if the
     # cert will not expire soon.
     command     => '/usr/bin/dehydrated --cron --privkey /etc/ssl/lets-encrypt/le-account.key',
     user        => $owner,
-    refreshonly => true,
     require     => Package['dehydrated-hook-ddns-tsig'],
+    # if the cert's chillin, only exec on notification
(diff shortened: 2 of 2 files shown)


modules/ocf/manifests/ssl/lets_encrypt/dns.pp
```diff
@@ -10,24 +10,30 @@
...
+
   exec { "obtain ${title} cert":
     # This exec can be notified to get it to run dehydrated again, even if the
     # cert will not expire soon.
     command     => '/usr/bin/dehydrated --cron --privkey /etc/ssl/lets-encrypt/le-account.key',
     user        => $owner,
-    refreshonly => true,
     require     => Package['dehydrated-hook-ddns-tsig'],
+    # if the cert's chillin, only exec on notification
+    refreshonly => $have_cert_info and !$cert_expires_soon and $cert_has_all_domains,
```
keur: Hmm, not sure if this is the logic we want here. If the first run fails to obtain a cert, I think these resources will be updated while the cert stays out of date. Then the second run, `refreshonly` will be set to true, but that says to only run if dependent resources are updated, which they will not be. 

Someone double check me on this though

abizer: shouldn't it pull this info fresh from the cert on disk every time it runs? if dehydrated fails to acquire the new cert, the next time puppet runs, the fact will return info from the unchanged cert.

modules/ocf/facts.d/le-cert-info
```diff
@@ -0,0 +1,30 @@
...
+            pem_data = pem_file.read()
+    except FileNotFoundError:
+        continue
+    except PermissionError as e:
+        print('try running this command as root or ocfletsencrypt')
+        raise e
+
+    cert = x509.load_pem_x509_certificate(pem_data, default_backend())
+    san = cert.extensions.get_extension_for_oid(x509.oid.ExtensionOID.SUBJECT_ALTERNATIVE_NAME)
+    cert_data[title] = {
```
dkess: Is it possible that there could be multiple entries here with the same title? That could cause problems here.

modules/ocf/facts.d/le-cert-info
```diff
@@ -0,0 +1,30 @@
...
+from datetime import datetime
+
+from cryptography import x509
+from cryptography.hazmat.backends import default_backend
+
+LETSENCRYPT_CERT_DIR = '/var/lib/lets-encrypt/certs/'
+
+cert_data = {}
+
+for title in os.listdir(LETSENCRYPT_CERT_DIR):
```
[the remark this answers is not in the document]
cg505: We are looping through all the different directories (which correspond to $title in the puppet resource). Each directory contains cert.pem which is a link to the most recent. So yes to both?

modules/ocf/manifests/ssl/lets_encrypt/dns.pp
```diff
@@ -10,24 +10,30 @@
...
+  # should help with puppet error spam a bit (for instance if Let's Encrypt is
+  # down for any reason) and mean that services can subscribe to this class to
+  # reload after new certs are obtained since this won't refresh unless
+  # dehydrated is re-run for some reason.
+
+  $parsed_cert_info = parsejson($::le_cert_info)
+
+  $have_cert_info = $title in $parsed_cert_info
+  if $have_cert_info {
+    $cert_expires_soon = ($parsed_cert_info[$title]['days_to_expiration'] + 0) < 30
```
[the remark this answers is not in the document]
cg505: Yeah this is wrong, since I switched to parsejson I don't need to cast anyway.

modules/ocf/facts.d/le-cert-info
```diff
@@ -0,0 +1,30 @@
...
+    try:
+        with open(os.path.join(LETSENCRYPT_CERT_DIR, title, 'cert.pem'), 'rb') as pem_file:
+            pem_data = pem_file.read()
+    except FileNotFoundError:
```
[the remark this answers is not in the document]
cg505: /var/lib/letsencrypt/certs/dirname exists but /var/lib/letsencrypt/certs/dirname/cert.pem doesn't

modules/ocf/manifests/ssl/lets_encrypt/dns.pp
cg505: Refreshonly is just a boolean that says whether or not we should only run the exec when the resource is notified. If the cert is out of date, or doesn't have all the domains, we should run the exec every time (refreshonly = false). Otherwise we should wait for notification (refreshonly = true).

modules/ocf/facts.d/le-cert-info
cg505: No. `title` comes from directory names, which must be unique within the /var/lib/lets-encrypt/certs/ directory

modules/ocf/manifests/ssl/lets_encrypt/dns.pp
cg505: I see... if we get a new config or something, and then fail to obtain a cert, we still won't retry. Not sure how to handle that case.

modules/ocf/manifests/ssl/lets_encrypt/dns.pp
```diff
@@ -10,24 +10,30 @@
...
-  # itself, but this should help with puppet error spam a bit (for instance
-  # if Let's Encrypt is down for any reason) and mean that services can
-  # subscribe to this class to reload after new certs are obtained since this
-  # won't refresh unless dehydrated is re-run for some reason.
-  exec { "check ${title} cert expiration":
-    command => '/bin/true',
-    unless  => "openssl x509 -checkend 2592000 -noout -in /var/lib/lets-encrypt/certs/${title}/cert.pem",
-    user    => ocfletsencrypt,
-  } ~>
+  # Only run the dehydrated command to renew the cert if it is old enough, or
```
jvperrin: I think the comment is a little out of place now that it's separated from the `exec` it was talking about, although I see why you put it here since the checking logic is just below here. Maybe it should be split into two sections, one about checks to perform on the cert and wanting to re-obtain the cert if the cert is old or the domain has changed, and another section about the exec and the puppet spam? Something like that anyway, since some of this feels like it belongs further down.

modules/ocf/manifests/ssl/lets_encrypt/dns.pp
```diff
@@ -10,24 +10,30 @@
...
+  # reload after new certs are obtained since this won't refresh unless
+  # dehydrated is re-run for some reason.
+
+  $parsed_cert_info = parsejson($::le_cert_info)
+
+  $have_cert_info = $title in $parsed_cert_info
+  if $have_cert_info {
+    $cert_expires_soon = $parsed_cert_info[$title]['days_to_expiration'] < 30
+    # we subtract out any domains that are in the cert, and see if any are left
+    $cert_has_all_domains = ($domains - $parsed_cert_info[$title]['cert_names']) =~ Array[String, 0, 0]
```
jvperrin: Nice, this is a pretty cool way of doing this. I wish there was a nicer syntax than `Array[String, 0, 0]` for this, but I don't see any way of doing this same kind of match.

modules/ocf/manifests/ssl/lets_encrypt/dns.pp
jvperrin: I think that's fine, it should be a rare case where we change the config, and definitely a much rarer case than the current system fails on.

The other way you could do this is have some kind of file or other state that indicates a previously failed run and always re-run if the prior run has failed, something like that. It's definitely not ideal to have to do that though.

keur: Yeah, this is an edge case. Maybe just leave a comment and dwai

modules/ocf/facts.d/le-cert-info
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+import json
+import os
+from datetime import datetime
+
+from cryptography import x509
+from cryptography.hazmat.backends import default_backend
```
jvperrin: I'm a little worried actually about using these, which I didn't notice the first time, since they are actually from the third-party `python3-cryptography` package. I think this script just won't run on a bootstrap of a new system until `python3-cryptography` is installed as a dependency of [`python3-paramiko` from `ocf::packages`](https://github.com/ocf/puppet/blob/487324734f4d34d5ae68cb79918ac0455cddc1d7/modules/ocf/manifests/packages.pp#L115).

Anyway, I think it should be ok since it's ok if facts fail to run, but maybe wrap this in a `try` block and if you get an `ImportError`, then just return `{}` from this fact or maybe some better fallback values if you can?

cg505: I feel like if this depends on `python3-cryptography`, than we should add that explicit dependency in a letsencrypt manifest as well in case the `python3-paramiko` dependency gets removed later.

Will add the `try` block.

cg505:
@jvperrin how is this

modules/ocf/manifests/ssl/lets_encrypt/dns.pp
```diff
@@ -5,6 +5,10 @@
 ) {
   require ocf::ssl::lets_encrypt::dns_common
 
+  # we use this package in the le_cert_info fact, to gather information about
+  # whatever certs are currently present
+  package { 'python3-cryptography':; }
```
abizer: you may also want to consider adding it to the preseed file (https://github.com/ocf/puppet/blob/master/modules/ocf_dhcp/files/preseed/buster#L91) since facts will be collected before any manifests run on new hosts

cg505: I thought about this but I don't think it's necessary because the fact won't crash on new hosts (since the import is guarded) and the fact will be empty on new hosts anyway (since they won't have certs yet).

modules/ocf/facts.d/le-cert-info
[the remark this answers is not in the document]
kpengboy: I would put this in a `main()`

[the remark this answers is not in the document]
kpengboy: I would put this in a `main()`

PR #451, opened 2018-12-21
Add Mastodon PersistentStorage

fydai:


modules/ocf_filehost/templates/exports.ebb
```diff
@@ -7,3 +7,7 @@
            dev-ssh(rw,fsid=0,no_subtree_check,no_root_squash) \
            apphost(rw,fsid=0,no_subtree_check,no_root_squash) \
            dev-apphost(rw,fsid=0,no_subtree_check,no_root_squash)
+
```
keur: i think you want the file extension to be erb?

modules/ocf_kubernetes/manifests/master/persistent_volume.pp
```diff
@@ -0,0 +1,19 @@
+class ocf_kubernetes::master::persistent_volume {
+
+  file {
+    default:
+      require => Package['kubeadm', 'kubectl'];
```
keur: nit: While it surely doesn't hurt, I don't think `kubeadm` is required for any of this.

modules/ocf_kubernetes/manifests/master.pp
```diff
@@ -87,4 +91,11 @@
   class { 'ocf_kubernetes::master::ingress::nginx':
     require => Class['kubernetes'],
   }
+
+  class { 'ocf_kubernetes::master::persistent_volume':
+    require => [
+      Class['kubernetes'],
+      File['/etc/profile.d/kubeconfig.sh'],
```
keur: since you are exporting the environment  variable, this requirement is no longer necessary (it also didn't work consistently, for whatever reason)

modules/ocf_filehost/manifests/init.pp
```diff
@@ -5,7 +5,7 @@
 
   file {
     '/etc/exports':
-      source  => 'puppet:///modules/ocf_filehost/exports',
+      content => epp('puppet:///modules/ocf_filehost/exports'),
```
keur: While this works, OCF tends to use the pattern:

`content => template(ocf_filehost/exports.erb)`.

It's much more apparent what is going on.

modules/ocf_filehost/templates/exports.ebb
keur: also, a comment here about what you doing would be nice

modules/ocf_kubernetes/files/mastodon-pv.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: PersistentVolume
+metadata:
+  name: mastodon-nfs-pv
+spec:
+  capacity:
+    storage: 30Gi
+  accessModes:
+    - ReadWriteMany
+  nfs:
+    path: /opt/homes/services/mastodon
```
keur: What if we instead had `/opt/homes/services/kubernetes` and simply made mastodon a subdirectory there? This way, in the future, if we want to add more `nfs` stuff to kubernetes, we can simply create another directory and use one PV that we can r+w to.

modules/ocf_kubernetes/files/mastodon-pv.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: mastodon-nfs-pv
+spec:
+  capacity:
+    storage: 30Gi
```
keur: This is okay for testing, but once we have a server running and people are uploading pictures and such, we probably want to bump this.

modules/ocf_filehost/templates/exports.ebb
```diff
@@ -7,3 +7,7 @@
...
+<% $kubernetes_workers = lookup('kubernetes::worker_nodes') -%>
+<% $worker_exports = $kubernetes_workers.map |$worker| { "${worker}(rw,fsid=0)" } -%>
+/opt/homes/services/mastodon <%= $worker_exports.join(" ") %>
```
keur: might want to `.join("\\ \n")` (that is probably wrong, but you get the point) so the exports file is human readable if we have a bunch of workers.

modules/ocf_filehost/templates/exports.ebb
```diff
@@ -7,3 +7,7 @@
            dev-ssh(rw,fsid=0,no_subtree_check,no_root_squash) \
            apphost(rw,fsid=0,no_subtree_check,no_root_squash) \
            dev-apphost(rw,fsid=0,no_subtree_check,no_root_squash)
+
+<% $kubernetes_workers = lookup('kubernetes::worker_nodes') -%>
```
keur: nit: makes more sense to make the declaration at the top of the file.

modules/ocf_kubernetes/files/mastodon-pv.yaml
abizer: 30GB seems like a lot. In however long we've been running mysql, it's still < 5 GB. This is probably safe.

modules/ocf_kubernetes/files/mastodon-pv.yaml
abizer: I think it's a better pattern to have separation between services, so we can modify them independently should it become necessary, and for general isolation as well.

keur: You can create this separation in the `PhysicalVolumeClaim`. See the mastodon one [here](https://github.com/fydai/mastodon/blob/master/kubernetes/mastodon-pv.yml). This includes access mode, amount of storage used, etc. The PV is just the backing for the storage mapping that  PVCs will request.

modules/ocf_kubernetes/files/mastodon-pv.yaml
keur: fair enough

modules/ocf_kubernetes/files/mastodon-pv.yaml
keur: What I said is wrong. There is a one-to-one mapping between PVs and PVCs, sadly. Instead, I would de-maston-ify this and we can put all future PVs in this file, separated by `--`. This will reduce the `kubectl` code in puppet.

modules/ocf_kubernetes/manifests/worker.pp
```diff
@@ -7,6 +7,8 @@
 # with the small requirement that kubernetes::worker
 # class is added.
 class ocf_kubernetes::worker {
+  package { 'nfs-common': }
```
dkess: Are you sure this is necessary? Isn't all the NFS stuff running in Docker anyways?

fydai: Yes, Kubernetes is mounting the NFS stuff on the workers, it's not being done in Docker.

abizer: mounting NFS inside the container would require advanced privileges IIRC and is recommended against

keur:
@fydai : #452 has been merged

modules/ocf_filehost/manifests/export.pp
```diff
@@ -0,0 +1,14 @@
+define ocf_filehost::export (
```
abizer: `ocf_filehost::nfs_export`

modules/ocf_filehost/manifests/nfs_export.pp
```diff
@@ -0,0 +1,14 @@
+define ocf_filehost::nfs_export (
```
keur: good call on the defined type

modules/ocf_kubernetes/files/persistent-volume.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
```
keur: nit: please name this file `persistent-volume-nfs`

modules/ocf_kubernetes/manifests/master.pp
keur: Reminder: still dont need this line

modules/ocf_kubernetes/manifests/master/persistent_volume.pp
```diff
@@ -0,0 +1,17 @@
+class ocf_kubernetes::master::persistent_volume {
```
keur: good call on putting this in its own file

fydai:
Thank you all so much for looking over my code! Really appreciate it. I will squash the commits tomorrow, to give some more time for review.

PR #452, opened 2018-12-21
Kube apply

keur:
sets your environment variable correct and ensure kubectl is installed


keur:
sorry for all the force pushes... i hate git

dkess:
This is fine, but I'd rather apply these resources by either putting them all in one folder and automatically running kubectl apply on all of them at once, or using the kubernetes repo to run them from Jenkins.

dkess:
Sure, in that case we shouldn't have a separate Puppet declaration for each yaml file. We should have one directory in `files` called "resources" or something, and do a bulk `kubectl apply` on everything there.

Issue #398, opened 2018-10-17
Box.com deprecating WebDAV


keur:
Box.com has announced that they are [deprecating WebDAV support](https://community.box.com/t5/Box-Product-News/Deprecation-WebDAV-Support/ba-p/55684) on January 31st, 2019. Indiana University has a nice page on [alternative ways to connect to Box.com](https://uisapp2.iu.edu/confluence-prd/display/SOICKB/Using+Box+under+Linux). I'm thinking that connecting to box over the `FTPS` protocol will be the best replacement of those listed.

jvperrin:
This seems like a duplicate of #367, although using FTPS is a bit different (compared to rclone or something like that).

PR #453, opened 2018-12-22
Change upload-to-box from dav to ftps

### LINKED ISSUE #367: Rewrite upload_to_box to not use DAV
box.com is deprecating DAV, see https://community.box.com/t5/Box-Product-News/Deprecation-WebDAV-Support/ba-p/55684

rclone has good support for this, but it could be difficult to puppet, since the version currently packaged in Debian doens't support box.com. We'll have to figure out how to get the latest version.

dkess:
Resolves #367 and #398. The script now uses the `lftp` binary to upload files to Box, which lets us get rid of a lot of custom upload code. I also got rid of code related to the progress bar, since `lftp` has its own mechanism for showing progress.

This has been semi-tested with a small folder I made on `hal` (`/opt/backups/scratch/dkessler`). It's accessible at https://berkeley.box.com/s/9nw4saovck21j75xhf1z1nddgmj78scf for those who have access to backups.

### DIFF
--- a/modules/ocf_backups/files/upload-to-box
+++ b/modules/ocf_backups/files/upload-to-box
@@ -105,24 +105,34 @@ def get_access_token(creds):
         return response['access_token']
 
 
-def upload_to_box(creds):
+def upload_to_box(creds, quiet):
     """Uploads the archive to box at the configured location"""
     lftp_cmd = '''
         set ftps:initial-prot ""
         set ftp:ssl-force true
         set ftp:ssl-protect-data true
         open {box_ftp}
-        user ocf@berkeley.edu {passwd}
+        user {email} {passwd}
         mirror -R --parallel={concurr} {archive} {folder}
     '''.format(
         box_ftp=BOX_FTP,
+        email=creds['email'],
         passwd=creds['password'],
         concurr=MAX_CONCURR_UPLOADS,
         archive=ARCHIVE,
         folder=FOLDER,
     )
 
-    subprocess.run(['lftp', '--norc'], input=lftp_cmd.encode()).check_output()
+    if quiet:
+        stdout_pipe = subprocess.DEVNULL
+    else:
+        stdout_pipe = None
+
+    subprocess.run(
+        ['lftp', '--norc'],
+        input=lftp_cmd.encode(),
+        stdout=stdout_pipe,
+    ).check_returncode()
 
 
 def get_folder_id(access_token, folder_name):
@@ -189,12 +199,13 @@ def main(args):
     with open('/opt/share/backups/box-creds.json') as f:
         creds = json.load(f)
 
+    # Used for statistics, shown at the end of the run.
     start_time = time.time()
     file_bytes = sum(os.path.getsize(ARCHIVE + '/' + f) for f in files)
     if not args.quiet:
         print('Uploading {}...'.format(humanize_bytes(file_bytes)))
 
-    upload_to_box(creds)
+    upload_to_box(creds, args.quiet)
 
     # Get a shared link from Box.com's API and share it with certain emails
     access_token = get_access_token(creds)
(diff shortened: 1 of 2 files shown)


modules/ocf_backups/files/upload-to-box
```diff
@@ -164,6 +105,26 @@ def get_access_token(creds):
...
+        mirror -R --parallel={concurr} {archive} {folder}
+    '''.format(
+        box_ftp=BOX_FTP,
+        passwd=creds['password'],
+        concurr=MAX_CONCURR_UPLOADS,
+        archive=ARCHIVE,
+        folder=FOLDER,
+    )
+
+    subprocess.run(['lftp', '--norc'], input=lftp_cmd.encode()).check_output()
```
jvperrin: I think this can just be `subprocess.check_call(['lftp', '--norc'], input=lftp_cmd.encode())` since you're discarding the output anyway, so collecting it isn't really useful.

modules/ocf_backups/files/upload-to-box
```diff
@@ -164,6 +105,26 @@ def get_access_token(creds):
...
 
 
+def upload_to_box(creds):
+    """Uploads the archive to box at the configured location"""
+    lftp_cmd = '''
+        set ftps:initial-prot ""
+        set ftp:ssl-force true
+        set ftp:ssl-protect-data true
+        open {box_ftp}
+        user ocf@berkeley.edu {passwd}
```
jvperrin: I think this email should come from `creds['email']` instead of being hardcoded here (or it should be removed from the private share's creds in favor of having it be a constant in this script).

modules/ocf_backups/files/upload-to-box
dkess: We actually can't use `check_call`, here, since it  doesn't support the `input` keyword.

modules/ocf_backups/files/upload-to-box
```diff
@@ -164,6 +105,36 @@ def get_access_token(creds):
...
 
+def upload_to_box(creds, quiet):
+    """Uploads the archive to box at the configured location"""
+    lftp_cmd = '''
+        set ftps:initial-prot ""
+        set ftp:ssl-force true
+        set ftp:ssl-protect-data true
+        open {box_ftp}
+        user {email} {passwd}
+        mirror -R --parallel={concurr} {archive} {folder}
```
keur: I know @jvperrin did some benchmarking. Did you try seeing if the box.com server supports `zlib` compression with:

```
set ftp:use-mode-z true
set ftp:mode-z-level 6
```

Since you are using mirror, you may also want to consider: 
```
set mirror:parallel-directories true
```

This will do directories in parallel, if possible, since you already pass the `--parallel` flag.

abizer: since the files we are uploading are encrypted, compression will just burn CPU as pseudorandom data is nearly incompressible

keur: ah, good point.

dkess: Don't think `parallel-directories` will do anything either since this just one directory with a bunch of flat files in it.

abizer:
100Mb/s into 1 week gives us a max backup size of about 7.5TB. FTP seems like a simpler protocol compared to WebDAV so I'm curious why it's 3x slower.

Because we're uploading encrypted files anyways we can also consider turning off protocol encryption since the double encryption isn't really necessary. Perhaps that will help improve throughput.

jvperrin:
Webdav is just http requests essentially, so I don't think it's that different in complexity, but I'd agree ftp should be about as fast. Anyway, 7.5 TB should be enough for a while (or we can backup offsite every 2 weeks or something instead).

One issue with turning off encryption is that we still need to authenticate (would this go over plaintext?), and I'm not even sure that FTP is enabled for our account from https://community.box.com/t5/Upload-and-Download-Files-and/Using-Box-with-FTP-or-FTPS/ta-p/26050

PR #454, opened 2018-12-22
add reptyr

keur:
Reasoning for this:

sleep 1000 ; [Ctrl+Z] ; bg ; disown ; tmux ; reptyr $(pidof sleep) 

Since Debian has /proc/sys/kernel/yama/ptrace_scope set to 0
by default, no further configuration is necessary.


dkess:
This should be in `extrapackages`.

dkess:
Yeah I guess this could be useful anywhere, nvmd then.

keur:
squashed because !sux github web editor

PR #456, opened 2018-12-26
Add prometheus textfiles to mirrors healthchecks

dkess:
This lets us do the healthcheck with Prometheus, which would be a bit nicer. The metrics get written to textfiles which in turn get scraped by the node_exporter.

I tested this in production and it looks like it works.

### DIFF
--- a/modules/ocf_mirrors/files/healthcheck
+++ b/modules/ocf_mirrors/files/healthcheck
@@ -16,6 +16,9 @@ from datetime import timedelta
 
 import dateutil.parser
 import requests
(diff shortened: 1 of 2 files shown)


dkess:
This lets us use the following query to see how out-of-date all of our mirrors are: [mirror_updated_upstream - mirror_updated_local](https://prometheus.ocf.berkeley.edu/graph?g0.range_input=1h&g0.expr=mirror_updated_upstream%20-%20mirror_updated_local&g0.tab=1)

(right now it only shows one, but when the cronjobs run, this will work for all mirrors)

PR #460, opened 2018-12-27
Remove redis on supernova

dkess:
This isn't used anymore, so we should be able to safely delete it.


dkess:
Done. Didn't want to mess up the file configs but I'll remove those manually if this PR goes well.

PR #461, opened 2018-12-29
Remove centos-debuginfo mirror

dkess:
This mirror has been "out of date" for weeks now, and apparently it was never supposed to be used, according to https://lists.centos.org/pipermail/centos-mirror/2018-December/011418.html.

I'm pretty sure this PR properly removes the mirror, but please lmk if I missed anything.

### DIFF
--- a/modules/ocf_mirrors/manifests/centos_debuginfo.pp
+++ b/modules/ocf_mirrors/manifests/centos_debuginfo.pp
@@ -11,15 +11,13 @@ class ocf_mirrors::centos_debuginfo {
   ocf_mirrors::monitoring {
     'centos-debuginfo':
       type          => 'unix_timestamp',
-      upstream_host => 'debuginfo.centos.org',
+      ensure        => absent,
       upstream_path => '/',
+      upstream_host => 'debuginfo.centos.org',
       ts_path       => 'TIME';
   }
 
-  ocf_mirrors::timer {
-    'centos-debuginfo':
-      exec_start => '/opt/mirrors/project/centos-debuginfo/sync-archive',
-      hour       => '0/3',
-      minute     => '56';
+  ocf::systemd::timer { 'centos-debuginfo':
+    ensure => absent;
   }
 }
--- a/modules/ocf_mirrors/manifests/centos_debuginfo.pp
+++ b/modules/ocf_mirrors/manifests/centos_debuginfo.pp
@@ -10,8 +10,8 @@ class ocf_mirrors::centos_debuginfo {
 
   ocf_mirrors::monitoring {
     'centos-debuginfo':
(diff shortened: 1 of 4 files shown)


modules/ocf_mirrors/manifests/centos_debuginfo.pp
```diff
@@ -11,15 +11,13 @@
   ocf_mirrors::monitoring {
     'centos-debuginfo':
       type          => 'unix_timestamp',
-      upstream_host => 'debuginfo.centos.org',
+      ensure        => absent,
```
keur: `puppet-lint` should've thrown an error, but ensure should always go first

abizer:
we should be using that endpoint since we run a mirror that's listed in the public mirrorlist.

before removing it we should probably reach out to ask if we can help in any way considering how few people actually mirror this seemingly useful resource.

modules/ocf_mirrors/manifests/init.pp
```diff
@@ -10,7 +10,6 @@
   include ocf_mirrors::archlinux
   include ocf_mirrors::centos
   include ocf_mirrors::centos_altarch
-  include ocf_mirrors::centos_debuginfo
```
abizer: if you remove this include, the ensure => absents will never get run

dkess:
I don't see it mentioned anywhere on their [public mirroring guide](https://wiki.centos.org/HowTos/CreatePublicMirrors). Perhaps they used to require it, but no longer do?

I'll go in IRC and see if I can get any more information.

abizer:
I remember reading somewhere that they made debuginfo a completely separate mirror because it had much higher disk space requirements than their regular mirrors and was necessary to separate it. disk space is also why we only mirror centos 6 and 7 debuginfo and exclude 4 and 5 from the upstream rsync.

abizer:
dkess and I spoke with the centos people on IRC, we're going to discuss this in the mailing list (once they accept our request to join)

dkess:
Email has been sent to the mailing list, and I've seen a some discussion in their IRC channel about this. Hopefully we'll get a response soon.

jvperrin:
Facebook just sent an email about this, we should be able to use `rsync://mirror.facebook.net/centos-debuginfo/` for this.

dkess:
Unfortunately their mirror is just as out of date as ours. However based on conversations in their IRC channel looks like they've been making progress on this.

fydai:
Is there an update on this?

abizer:
Facebook finally set up their mirrors for debuginfo, we should switch our upstream to that

fydai:
Updated PR merged, https://github.com/ocf/puppet/pull/648

PR #462, opened 2018-12-30
Add prune-old-backups script

dkess:
This script removes old backups. Going to wait for a lot of approvals and scrutiny before I even add the code to run this on a cronjob, since this is script has potential to be very destructive. 

I haven't fully tested the script, but I can confirm that a dry run works (displays the correct information), and I can confirm that running `prune_old_backups.delete_items(creds, ['ocf-backup-2016-06-09', 'ocf-backup-2016-07-25'], False)` in an interactive shell works.

### DIFF
--- a/modules/ocf_backups/files/prune-old-backups
+++ b/modules/ocf_backups/files/prune-old-backups
@@ -10,8 +10,9 @@ import textwrap
 
 RETENTION_TIME = datetime.timedelta(days=180) # 6 months
 
-BOX_FTP = 'ftps://ftp.box.com:990'
 BACKUP_PREFIX = 'ocf-backup-'
+BOX_FTP = 'ftps://ftp.box.com:990'
+CREDS_PATH = '/opt/share/backups/box-creds.json'
 
 LFTP_INIT = '''
     set ftps:initial-prot ""
@@ -78,7 +79,7 @@ def is_old_backup(filename):
     return datetime.date.today() - backup_date > RETENTION_TIME
 
 def main(args):
-    with open('/opt/share/backups/box-creds.json') as f:
+    with open(CREDS_PATH) as f:
         creds = json.load(f)
 
     items = set(list_items(creds))
@@ -93,8 +94,8 @@ def main(args):
         print('Items to delete:')
         print(textwrap.indent('\n'.join(sorted(to_delete)), '  '))
 
-    if args.dry_run:
-        return
+    if not args.dry_run:
+        delete_items(to_delete)
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Prune old Box.com backups')
--- a/modules/ocf_backups/files/prune-old-backups
+++ b/modules/ocf_backups/files/prune-old-backups
@@ -102,7 +102,7 @@ def main():
 
 
     if not args.dry_run:
-        delete_items(to_delete)
+        delete_items(creds, to_delete, args.quiet)
 
 if __name__ == '__main__':
     main()
--- a/modules/ocf_backups/files/prune-old-backups
+++ b/modules/ocf_backups/files/prune-old-backups
@@ -1,0 +1,128 @@
+#!/usr/bin/env python3
+
+import argparse
+import datetime
+import json
+import shlex
+import subprocess
+import sys
+import textwrap
+
+RETENTION_TIME = datetime.timedelta(days=180) # 6 months
+
+BOX_FTP = 'ftps://ftp.box.com:990'
+BACKUP_PREFIX = 'ocf-backup-'
+
+LFTP_INIT = '''
+    set ftps:initial-prot ""
+    set ftp:ssl-force true
+    set ftp:ssl-protect-data true
+    open {box_ftp}
+    user {email} {passwd}
+'''
+
+def parse_date(dtstr):
+    """Converts a 'YYYY-MM-DD' formatted string to a datetime.date object."""
+    if len(dtstr) != 10:
+        raise ValueError
+
+    year = int(dtstr[0:4])
+
+    if dtstr[4] != '-':
+        raise ValueError
+
+    month = int(dtstr[5:7])
+
+    if dtstr[7] != '-':
+        raise ValueError
+
+    day = int(dtstr[8:10])
+
+    return datetime.date(year, month, day)
+
+def list_items(creds):
+    """List all items at the root of the Box.com folder."""
+    lftp_cmd = (LFTP_INIT + 'renlist').format(
+        box_ftp=BOX_FTP,
+        email=creds['email'],
+        passwd=creds['password'],
+    )
+
+    p = subprocess.run(
+        ['lftp', '--norc'],
+        input=lftp_cmd.encode(),
+        stdout=subprocess.PIPE,
+    )
+
+    p.check_returncode()
+
+    return p.stdout.decode().splitlines()
+
+def delete_items(creds, items, quiet):
+    lftp_cmd = (LFTP_INIT + 'rm -r {items}').format(
+        box_ftp=BOX_FTP,
+        email=creds['email'],
+        passwd=creds['password'],
+        items=' '.join(shlex.quote(item) for item in items)
+    )
+
+    if quiet:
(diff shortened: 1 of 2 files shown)


modules/ocf_backups/files/prune-old-backups
```diff
@@ -0,0 +1,128 @@
...
+LFTP_INIT = '''
+    set ftps:initial-prot ""
+    set ftp:ssl-force true
+    set ftp:ssl-protect-data true
+    open {box_ftp}
+    user {email} {passwd}
+'''
+
+def parse_date(dtstr):
+    """Converts a 'YYYY-MM-DD' formatted string to a datetime.date object."""
```
abizer: does strptime("%Y-%m-%d") not work here?

modules/ocf_backups/files/prune-old-backups
```diff
@@ -0,0 +1,128 @@
...
+    if not args.quiet:
+        print('Items to keep:')
+        print(textwrap.indent('\n'.join(sorted(items - to_delete)), '  '))
+        print()
+
+        print('Items to delete:')
+        print(textwrap.indent('\n'.join(sorted(to_delete)), '  '))
+
+    if args.dry_run:
+        return
```
abizer: is there supposed to be an else case here to actually delete the items?

modules/ocf_backups/files/prune-old-backups
```diff
@@ -0,0 +1,128 @@
...
+        raise ValueError
+
+    day = int(dtstr[8:10])
+
+    return datetime.date(year, month, day)
+
+def list_items(creds):
+    """List all items at the root of the Box.com folder."""
+    lftp_cmd = (LFTP_INIT + 'renlist').format(
+        box_ftp=BOX_FTP,
```
abizer: I would abstract these out into another function like `lftp_cmd(command, output=subprocess.PIPE)`

modules/ocf_backups/files/prune-old-backups
```diff
@@ -0,0 +1,128 @@
...
+
+    try:
+        backup_date = parse_date(end)
+    except ValueError:
+        return False
+
+    return datetime.date.today() - backup_date > RETENTION_TIME
+
+def main(args):
+    with open('/opt/share/backups/box-creds.json') as f:
```
abizer: put this at the top as `CREDS_PATH` or something?

modules/ocf_backups/files/prune-old-backups
dkess: There are enough differences between the two command calls that I think it more sense to keep them separate. An abstraction here would just make the code more confusing.

modules/ocf_backups/files/prune-old-backups
dkess: Sure, we should probably do that for the `upload-to-box` script as well.

modules/ocf_backups/files/prune-old-backups
dkess: Yeah that's probably better.

modules/ocf_backups/files/prune-old-backups
dkess: lol forgot that part

modules/ocf_backups/files/prune-old-backups
abizer: it looks like the major difference is `renlist` vs `rm -r {items}.format(items=' '.join(shlex.quote...)`? unless I'm missing something?

would 

```python
def lftp_cmd(command, out=subprocess.PIPE):
  cmd = LFTP_INIT.format(...)
  cmd += command
  p = subprocess.run(('lftp', '--norc'), input=cmd, output=out)
  p.check_returncode()
```

work?

dkess: The format args dict has to get passed in both times, but that breaks the abstraction barrier because the caller has to know which template fields to fill in for `LFTP_INIT`.

abizer: I mean both are filling in `box_ftp`, `email`, and `passwd`, and the command format field can be rendered by the caller, right? e.g.

`lftp_cmd('renlist')` and `lftp_cmd('rm -r ' + ' '.join(shlex.quote(item) for item in items))`

dkess: Ok on second thought you're right, since this makes the `list_items` and `delete_items` functions clearer to understand. New commit incoming

modules/ocf_backups/files/prune-old-backups
```diff
@@ -0,0 +1,114 @@
...
+        print()
+
+        print('Items to delete:')
+        print(textwrap.indent('\n'.join(sorted(to_delete)), '  '))
+
+    if not args.dry_run:
+        delete_items(to_delete)
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Prune old Box.com backups')
```
ethanhs: It seems odd (not necessarily bad) to do arg parsing outside of main. I will have to think about whether I like it or not 🤔

modules/ocf_backups/files/prune-old-backups
```diff
@@ -0,0 +1,114 @@
...
+        help='Show which backups would be deleted, but do not delete.'
+    )
+    parser.add_argument(
+        '-q', '--quiet',
+        action='store_true',
+        help='do not output progress or information to stdout',
+    )
+
+    args = parser.parse_args()
+    sys.exit(main(args))
```
ethanhs: `main` doesn't return anything, so I don't see why we are calling `sys.exit` on the result, which will always be `None`...

modules/ocf_backups/files/prune-old-backups
```diff
@@ -0,0 +1,114 @@
...
+    parser = argparse.ArgumentParser(description='Prune old Box.com backups')
+    parser.add_argument(
+        '-n', '--dry-run',
+        action='store_true',
+        help='Show which backups would be deleted, but do not delete.'
+    )
+    parser.add_argument(
+        '-q', '--quiet',
+        action='store_true',
+        help='do not output progress or information to stdout',
```
ethanhs:
```suggestion
        help='Do not output progress or information to stdout.',
```

modules/ocf_backups/files/prune-old-backups
dkess: This is standard for pretty much all of our scripts.

modules/ocf_backups/files/prune-old-backups
dkess: This was a mistake, I'll move it to the main function.

dkess:
Thanks everyone for the feedback so far, I've implemented all of the resolved comments.

modules/ocf_backups/files/prune-old-backups
ethanhs: At least in the scripts I looked at, `main` returns some exit code (which is actually not great, I much prefer explicit `sys.exit`ing).

In other scripts where it does not, this is not used (https://github.com/ocf/puppet/blob/master/modules/ocf_desktop/files/xsession/auto-lock, https://github.com/ocf/puppet/blob/master/modules/ocf_desktop/files/xsession/lab-close-notify). Its pointless to call sys.exit here, as the script will exit with 0 anyway (or it won't, which this won't change).

dkess: Alright fair enough

modules/ocf_backups/manifests/offsite.pp
```diff
@@ -11,6 +11,10 @@
       source => 'puppet:///modules/ocf_backups/upload-to-box',
       mode   => '0755';
 
+    '/opt/share/backups/prune-old-backups':
```
abizer: we need to add a cronjob for this as well, right?

dkess: Yeah, gonna do that before merging. Just didn't want to put it in before the code got properly reviewed, to prevent it from running.

dkess:
Made some minor changes, but I've added the cronjob so this is officially running in prod. I am currently doing a full run of the script and it appears to be working.

PR #467, opened 2018-12-31
Check for ValueError in process-mirrors-logs

dkess:
Got some rootspam about this last night. Probably a small one-off thing and we don't want the whole script to crash because of it.


abizer:
considering this comes out of a well-structured log, that the date doesn't parse is alarming. rather than silencing the error I'd print the entire line of context to stderr before continuuing so we can look at why it was strange

dkess:
We got this error again so I'm gonna merge this

PR #469, opened 2018-12-31
Redirect any SSL aliases too in nginx_proxy

### LINKED ISSUE #464: sg.ocf.io cert is broken
@jvperrin mentioned this some time ago but I thought I'd make a ticket for it

jvperrin:
This just makes sure that if we're making a proxy with SSL that we also listen on SSL for any aliases and redirect them to the canonical domain. This enables domains like http://sg.ocf.io, http://templates.ocf.io, http://grafana.ocf.io, etc. which otherwise wouldn't redirect properly because we have HSTS on for `*.ocf.io`, so the redirect to HTTPS comes before our server is actually contacted.

One assumption made here is that we have a "bundled" certificate for the domain that also contains all aliases within it, which might not be a proper assumption for all proxies.

Funny enough, other servers using nginx_proxy already (sorta) work, for instance https://irc.ocf.io, https://puppet.ocf.io, and https://jenkins.ocf.io work perfectly fine, but only because they fall back to the default vhost in nginx, which happens to be the one listening for SSL and with the correct cert already. However, it doesn't redirect to the canonical domain properly yet, so that would be nice to have.

Closes #464 (for real this time)

### DIFF
--- a/modules/ocf/manifests/nginx_proxy.pp
+++ b/modules/ocf/manifests/nginx_proxy.pp
@@ -42,7 +42,6 @@ define ocf::nginx_proxy(
           # HSTS header
           'Strict-Transport-Security' => 'max-age=31536000',
         }, $headers),
-
         *                => $nginx_options;
 
       "${title}-redirect":
@@ -51,8 +50,6 @@ define ocf::nginx_proxy(
           'return' => "301 https://${server_name}\$request_uri"
         },
 
-        # We have to specify www_root even though we always redirect/proxy
-        www_root          => '/var/www',
         add_header        => $headers,
         *                 => $nginx_options;
     }
@@ -76,8 +73,6 @@ define ocf::nginx_proxy(
             'Strict-Transport-Security' => 'max-age=31536000',
           }, $headers),
 
-          # We have to specify www_root even though we always redirect/proxy
-          www_root          => '/var/www',
           *                 => $nginx_options;
       }
     }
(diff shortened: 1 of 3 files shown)


modules/ocf/manifests/nginx_proxy.pp
```diff
@@ -46,19 +46,41 @@
...
-        # We have to specify www_root even though we always redirect/proxy
-        www_root          => '/var/www',
-
         server_name       => concat([$server_name], $server_aliases),
-
         server_cfg_append => {
           'return' => "301 https://${server_name}\$request_uri"
         },
 
+        # We have to specify www_root even though we always redirect/proxy
```
ethanhs: :(

jvperrin: I actually haven't tried changing this, it's just been copied from place to place. I'll try and remove it and see if it still works, but I somehow doubt it will.

ethanhs: Ah I see. Yeah I mostly was just marking my unhappiness at the situation. Thanks for trying without!

abizer:
nginx::resource::server/location has an `ssl_redirect` parameter, right? would it be possible to use that?

jvperrin:
I don't believe so because that's for redirecting from HTTP -> HTTPS instead of this, which is redirecting from HTTPS to some other HTTPS. I think it's also been renamed to something else in our version of puppet-nginx, but I'll take a closer look once I'm in something that isn't my phone.

modules/ocf/manifests/nginx_proxy.pp
jvperrin: So I tested this, and it worked without this specified, so that's pretty great! Thanks for the idea, I've removed it everywhere else this was specified and tested the redirects/proxying still works.

Issue #471, opened 2019-01-01
mysql root password preseed doesn't work


abizer:
the preseed file we include for debconf (https://github.com/ocf/puppet/blob/master/modules/ocf_mysql/manifests/init.pp#L6) has two problems. First, we're now on mariadb-server-10.1, so the package name for debconf needs to be updated, but more importantly, the mariadb-server-10.1 package templates doesn't appear to support mysql-server/root_password and root_password_again debconf options at all:

```bash
root@e5a1d73f257e:/# sudo debconf-show mariadb-server-10.1                                                                                                                               
  mariadb-server-10.1/nis_warning:
  mariadb-server-10.1/old_data_directory_saved:
  mariadb-server-10.1/postrm_remove_databases: false
```

unless I'm interpreting this incorrectly, we'll need to change the way we preseed the root password and do initial grants.

jvperrin:
From [this Debian bug comment](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=865931#28) it looks like we shouldn't need to preseed it any more since the local root user can log in over a unix socket. We also have the root password set up in the database already so when we migrate to a new host then when we import from backups (or from the current database host), then the root password should already be set, so I don't think we need to preseed or do any fancy puppet config for it.

Thoughts? I'm going to close this for now and delete the preseed file from the private share (we can remake it knowing the root password), I think we can just not do it, but I'm open to feedback, feel free to reopen this.

abizer:
Yeah I don't think this is necessary anymore. Can be deleted and closed.

PR #472, opened 2019-01-01
Enable encryption of MySQL connections

abizer:
```bash
root@dev-maelstrom:~# mysql -u root                                                        
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 2
Server version: 10.1.37-MariaDB-0+deb9u1 Debian 9.6

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [(none)]> show variables like '%ssl%';
+---------------------+-----------------------------------------------------+
| Variable_name       | Value                                               |
+---------------------+-----------------------------------------------------+
| have_openssl        | NO                                                  |
| have_ssl            | YES                                                 |
| ssl_ca              |                                                     |
| ssl_capath          |                                                     |
| ssl_cert            | /etc/ssl/private/dev-maelstrom.ocf.berkeley.edu.crt |
| ssl_cipher          |                                                     |
| ssl_crl             |                                                     |
| ssl_crlpath         |                                                     |
| ssl_key             | /etc/ssl/private/dev-maelstrom.ocf.berkeley.edu.key |
| version_ssl_library | YaSSL 2.4.4                                         |
+---------------------+-----------------------------------------------------+
10 rows in set (0.00 sec)

MariaDB [(none)]>
```

also remove the preseed stuff re: #471


modules/ocf_mysql/manifests/init.pp
```diff
@@ -1,30 +1,31 @@
 class ocf_mysql {
-  file {
-    # preseed root password
-    '/opt/share/puppet/mariadb-server-10.0.preseed':
-      mode      => '0600',
-      source    => 'puppet:///private/mariadb-server-10.0.preseed',
-      show_diff => false;
 
+  include ocf::ssl::default;
```
chriskuehl: maybe use `require` instead of `include` to ensure that the SSL files are in place before the service is configured/started (or add an explicit dependency later on)

otherwise i think the initial puppet run may fail?

modules/ocf_mysql/manifests/init.pp
```diff
@@ -1,30 +1,31 @@
...
+
+  user { 'mysql':
+    groups => ['ssl-cert'];
+  }
```
chriskuehl: what creates the `mysql` user? the package install? if so, this should probably depend on the package, or it might fail on the first puppet run

modules/ocf_mysql/manifests/init.pp
```diff
@@ -1,30 +1,31 @@
...
+
+  file {
     '/root/.my.cnf':
       mode      => '0600',
       source    => 'puppet:///private/root-my.cnf',
       show_diff => false;
   }
 
   class { 'ocf::packages::mysql_server':
```
keur: nit: it would make more sense to have this be the first thing in the file after ssl is pulled in. most things are going to have a dep on it

modules/ocf_mysql/manifests/init.pp
keur: You probably want `system => true`

modules/ocf_mysql/manifests/init.pp
keur: nit: it would make more sense to have this be the first thing in the file after ssl is pulled in. most things are going to have a dep on it too

modules/ocf_mysql/templates/99-ocf.cnf
```diff
@@ -1,6 +1,10 @@
 [mysqld]
 bind-address = ::
 
+# enable encrypted connections
+ssl-cert=/etc/ssl/private/<%= @fqdn %>.crt
+ssl-key=/etc/ssl/private/<%= @fqdn %>.key
+
```
jvperrin: You don't need `ssl-ca` to be specified here too? https://mariadb.com/kb/en/library/securing-connections-for-client-and-server/#configuration

PR #473, opened 2019-01-02
Add puppet certs to ocfweb dev config

dkess:
This commit lets all machines get access to the ocfweb puppet cert, and makes it part of the ocfweb dev config, so users testing on supernova (or staffvms) can use them.

### DIFF
--- a/modules/ocf_puppet/manifests/puppetserver.pp
+++ b/modules/ocf_puppet/manifests/puppetserver.pp
@@ -51,6 +51,18 @@ class ocf_puppet::puppetserver {
     notify               => Service['puppetserver'],
   }
 
+  # let anyone get the ocfweb certs
+  puppet_authorization::rule { 'ocfweb-cert':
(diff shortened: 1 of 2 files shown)


modules/ocf_puppet/manifests/puppetserver.pp
```diff
@@ -51,6 +51,18 @@
     notify               => Service['puppetserver'],
   }
 
+  # let anyone get the ocfweb certs
+  puppet_authorization::rule { 'ocfweb-cert':
+    match_request_path   => '^/puppet/v3/file_(content|metadata)s?/private-docker/ocfweb/puppet-certs$',
+    match_request_type   => 'regex',
+    match_request_method => ['get', 'post'],
+    allow                => '*.ocf.berkeley.edu',
```
kpengboy: Are we concerned about tsunami, desktops, etc. having access to this?

dkess: I don't think users would actually be able to access this, since they'd still need the Puppet private key.

PR #478, opened 2019-01-06
create-encrypted-backup: add --no-tty option

dkess:
We've been getting some "cannot open '/dev/tty': No such device or address" errors from cron lately, so this could fix it. Unfortunately it's hard to test because it only happens over cron, however this seems like a promising solution.


abizer:
did we update gpg recently?

PR #480, opened 2019-01-07
Use systemd timers instead of cron for healthchecks

dkess:
We've been getting an increased amount of rootspam recently because we are running the healthcheck at 5m frequency instead of 6 hours. We want to know if a mirror endpoint is down, but we don't need a root@ email if an endpoint is down for a few scrapes.

Since systemd timers log instead of send email alerts, they make sense for this. Now the healthcheck fails "silently" but if it hasn't been updated for 3 hours, we alert. Then we can check the logs to see what the alert is for.

### DIFF
--- a/modules/ocf_mirrors/manifests/timer.pp
+++ b/modules/ocf_mirrors/manifests/timer.pp
@@ -4,6 +4,7 @@ define ocf_mirrors::timer(
   $day = '*',
   $month = '*',
   $year = '*',
+  $type = 'sync',
   $exec_user = 'mirrors',
   $exec_start = '',
   $environments = {},
--- a/modules/ocf_mirrors/templates/systemd/sync.timer.erb
+++ b/modules/ocf_mirrors/templates/systemd/sync.timer.erb
@@ -1,5 +1,5 @@
 [Unit]
-Description=Sync timer for <%= @title %>
+Description=<% if @type == "sync" %>Sync<% elsif @type == "monitor" %>Monitoring<% end %> timer for <%= @title %>
 After=network.target
 Wants=network.target
 
--- a/modules/ocf_mirrors/templates/systemd/sync.timer.erb
+++ b/modules/ocf_mirrors/templates/systemd/sync.timer.erb
@@ -1,5 +1,5 @@
 [Unit]
-Description=<% if @type == "sync" %>Sync<% elsif @type == "monitor" %>Monitoring<% end %> timer for <%= @title %>
+Description=<%= @type %> timer for <%= @title %>
 After=network.target
 Wants=network.target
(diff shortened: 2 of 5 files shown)


modules/ocf_mirrors/templates/systemd/sync.timer.erb
```diff
@@ -1,5 +1,5 @@
 [Unit]
-Description=Sync timer for <%= @title %>
+Description=<% if @type == "sync" %>Sync<% elsif @type == "monitor" %>Monitoring<% end %> timer for <%= @title %>
```
keur: I'd just set the `type` variable accordingly and do `Description=<% @type %>`. If not just split the conditionals over multiple lines for readability.

jvperrin: Yeah I'd just do `Description=<%= @type %> timer for <%= title %>`, it makes enough sense and doesn't involved logic like this.

modules/ocf_mirrors/manifests/timer.pp
```diff
@@ -4,6 +4,7 @@
   $day = '*',
   $month = '*',
   $year = '*',
+  $type = 'sync',
```
jvperrin: Maybe make this an Enum type so that it can't be given some other value that doesn't exist?

Not that it really matters though, this is only used in the description of the timer anyway.

modules/ocf_prometheus/files/rules.d/mirror.rules.yml
```diff
@@ -1,3 +1,6 @@
+# TODO: add an annotation explaining that these alerts can be investigated by
+# reading {{ $labels.project }}.service and {{ $labels.project }}-health.service
+# on fallingrocks
```
jvperrin: How long would this take to actually add? Could this be done instead of adding it as a `TODO`?

It sounds to me like this is a pretty simple change, so that why I'm asking this, but let me know if this is more complicated or if it's not feasible to do for some reason.

modules/ocf_mirrors/manifests/timer.pp
dkess: Yeah I think I'll just leave it as a string for now

modules/ocf_prometheus/files/rules.d/mirror.rules.yml
dkess: I wanted to give myself more time to think about a convention for alert annotations, but I should do this now because I'll probably forget :P

dkess:
Thanks for the feedback. Gonna merge this now but feel free to leave more comments and I can fix them in another commit.

PR #481, opened 2019-01-10
Add automatic docker image removal and cleanup to ocf_docker.

redplanks:
This stops the Docker registry while garbage collection is happening, as if an image is being uploaded as GC is happening, GC could delete one if its layers. If the registry goes down as a push happens, or if Puppet restarts the registry while GC is happening, that might be bad, but hopefully that is rare enough to not happen (too often).

Script has been tested with dry run flags, and output looks good.

### DIFF
--- a/modules/ocf_docker/files/garbage-collect
+++ b/modules/ocf_docker/files/garbage-collect
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -euo pipefail
+
+# Delete all images older than 2 months.
+deckschrubber -repos 9999 -month 2
+
+systemctl stop docker-registry
+docker run --rm \
+    -v /var/lib/registry:/var/lib/registry \
+    registry:2 \
+    registry garbage-collect /etc/docker/registry/config.yml
+
+# Always make sure the registry is started, no matter what fails.
+trap 'systemctl start docker-registry' EXIT
--- a/modules/ocf_docker/manifests/deckschrubber.pp
+++ b/modules/ocf_docker/manifests/deckschrubber.pp
@@ -1,0 +1,20 @@
+class ocf_docker::deckschrubber {
+  package { 'deckschrubber': }
+
+  file { ['/opt/docker', '/opt/docker/registry']:
+    ensure => directory,
+    mode   => '0755',
+  } ->
+  file { '/opt/docker/registry/garbage-collect':
+    source  => 'puppet:///modules/ocf_docker/garbage-collect',
+    mode    => '0755',
+    require => Package['deckschrubber'],
+  } ->
+  cron { 'registry-gc':
+    command  => '/opt/docker/registry/garbage-collect',
+    user     => 'root',
+    monthday => 1,
+    hour     => 7,
+    minute   => 0,
+  }
+}
--- a/modules/ocf_docker/manifests/init.pp
+++ b/modules/ocf_docker/manifests/init.pp
@@ -3,6 +3,8 @@ class ocf_docker {
   include ocf::packages::docker
   require ocf::ssl::default
 
+  include ocf_docker::deckschrubber
+
   class { 'nginx':
     manage_repo  => false,
     confd_purge  => true,


modules/ocf_docker/manifests/deckschrubber.pp
```diff
@@ -0,0 +1,20 @@
...
+  package { 'deckschrubber': }
+
+  file { ['/opt/docker', '/opt/docker/registry']:
+    ensure => directory,
+    mode   => '0755',
+  } ->
+  file { '/opt/docker/registry/garbage-collect':
+    source  => 'puppet:///modules/ocf_docker/garbage-collect',
+    mode    => '0755',
+    require => Package['deckschrubber'],
```
keur: This script also requires `File['/var/lib/registry:/var/lib/registry']` and `File['/etc/docker/registry/config.yml']` to exist.

modules/ocf_docker/manifests/deckschrubber.pp
```diff
@@ -0,0 +1,20 @@
...
+  } ->
+  cron { 'registry-gc':
+    command  => '/opt/docker/registry/garbage-collect',
+    user     => 'root',
+    monthday => 1,
```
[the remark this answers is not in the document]
abizer: weekly seems good to me, the only host on which this operates is biohazard anyways

modules/ocf_docker/manifests/deckschrubber.pp
redplanks: You're right about `/var/lib/registry`. The config file is the default one included in the container though.

modules/ocf_docker/manifests/deckschrubber.pp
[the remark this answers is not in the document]
redplanks: I picked one month to minimize the downtime the registry has, but once per week sounds ok to me as well.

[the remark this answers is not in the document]
jvperrin: I'd agree with once per week, every day seems too frequent because it won't even clean up much after that long and it takes the registry out of service, and every month seems too infrequent since one failure could mean a whole month of images piles up.

modules/ocf_docker/manifests/init.pp
```diff
@@ -2,6 +2,7 @@
   include ocf::firewall::allow_web
   include ocf::packages::docker
   require ocf::ssl::default
+  include ocf_docker::deckschrubber
```
jvperrin: I'd probably just include this above the ssl thing to make this look more uniform, but it doesn't really make a difference, I just think it would look a little nicer :P

modules/ocf_docker/files/garbage-collect
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -euo pipefail
+
+# Always make sure the registry is started, no matter what fails.
+trap 'systemctl start docker-registry' EXIT
```
jvperrin: Nice, this is something I've never actually used in bash, but have seen around. This is a pretty good use of it :)

modules/ocf_docker/files/garbage-collect
```diff
@@ -0,0 +1,14 @@
...
+
+# Delete all images older than 2 months.
+deckschrubber -repos 9999 -month 2
+
+systemctl stop docker-registry
+docker run --rm \
+    -v /var/lib/registry:/var/lib/registry \
+    registry:2 \
```
jvperrin: This is not a comment on this PR as much as a general one, but I do wonder how we should manage versioning our running third-party docker images, I mean we just pin this to `2` but it won't pull any new versions unless it's shut down, the local cache is cleaned, and it's started up again. This could be done automatically, but then it has the potential to break on some unexpected upgrade change. It could be pinned to a more specific version like `2.7`, but then it needs to be bumped like [marathon-lb](https://github.com/ocf/puppet/blob/487324734f4d34d5ae68cb79918ac0455cddc1d7/modules/ocf_mesos/files/master/load_balancer/ocf-lb.service#L7) does. Altogether, I'm not sure what's the best path here.

(Yes, this is doing the garbage collection part, not actually managing the service, but I thought I'd dump this thought somewhere)

modules/ocf_docker/manifests/deckschrubber.pp
```diff
@@ -0,0 +1,22 @@
+class ocf_docker::deckschrubber {
+  package { 'deckschrubber': }
+
+  file { ['/opt/docker', '/opt/docker/registry']:
```
jvperrin: I'd probably put this in `/opt/share/puppet/` since we tend to put other scripts there, like [some backup stuff](https://github.com/ocf/puppet/blob/487324734f4d34d5ae68cb79918ac0455cddc1d7/modules/ocf_backups/manifests/offsite.pp#L6-L16), [some puppet group script](https://github.com/ocf/puppet/blob/487324734f4d34d5ae68cb79918ac0455cddc1d7/modules/ocf/manifests/groups.pp#L3-L8), and [other random scripts](https://github.com/ocf/puppet/blob/487324734f4d34d5ae68cb79918ac0455cddc1d7/modules/ocf_desktop/manifests/xsession.pp#L13-L27), etc. It's also already created for you in `ocf::puppet`, so you can just require it as a resource when placing the script there to make it a little cleaner if you'd like.

modules/ocf_docker/manifests/deckschrubber.pp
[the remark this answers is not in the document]
kpengboy: On the other hand though, in the past we have only needed to clean the docker registry once every several months. I personally would prefer cleaning on a monthly basis since it's less frequent downtime.

modules/ocf_docker/files/garbage-collect
redplanks: Pinning it to a more specific version would probably be better. I was thinking maybe setting an environment variable in biohazard using hiera to control which image version is used, and then bumping that hiera variable. But maybe that's unnecessary, and the versions can just be changed in the script and the systemd unit file.

PR #482, opened 2019-01-10
keyserver: redirect http to https

abizer:
@keur was right on this, since HKP operates on its own port,
it's ok for the automatic redirect to occur. The web interface
isn't used by clients, so the 80->443 redirect probably won't
break them.

### DIFF
--- a/modules/ocf_keyserver/templates/sks-web-nginx.erb
+++ b/modules/ocf_keyserver/templates/sks-web-nginx.erb
@@ -1,7 +1,12 @@
 server {
-
     listen 80 default_server;
     listen [::]:80 default_server;
+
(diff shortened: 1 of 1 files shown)


dkess:
Perhaps we should be using a set of [recommended HTTPS settings](https://mozilla.github.io/server-side-tls/ssl-config-generator/)? Or do we have a snippet somewhere that we can re-use?

abizer:
@dkess I added most of the settings you recommended. I'm not going to add HSTS because it breaks HKPS on the *.berkeley.edu domain names, but the *.ocf.io ones are pre-seeded for HSTS anyways, so it should be fine.

PR #483, opened 2019-01-13
Kubernetes secrets

fydai:
Add a kubernetes-secrets puppetserver export and apply it on the kubernetes masters.


modules/ocf_puppet/manifests/puppetserver.pp
```diff
@@ -39,6 +39,18 @@
     notify               => Service['puppetserver'],
   }
 
+  # Allow kubernetes secrets to access kubernetes secrets
+  puppet_authorization::rule { 'kubernetes-secrets':
+    match_request_path   => '^/puppet/v3/file_(content|metadata)s?/kubernetes-secrets$',
+    match_request_type   => 'regex',
+    match_request_method => ['get', 'post'],
```
abizer: post?

fydai: I copied it from the docker secrets, should it be removed there too?

modules/ocf_kubernetes/manifests/master/secrets.pp
```diff
@@ -0,0 +1,18 @@
+class ocf_kubernetes::master::secrets {
+
+  file {
+    default:
+      require => Package['kubectl'];
+
+    '/etc/ocf-kubernetes/secrets':
+      ensure  => 'directory',
+      source  => 'puppet:///kubernetes-secrets',
```
keur: err... is this the right location of the file? if this is meant to point somewhere in the private share, you should use the `kubernetes` directory that we setup.

modules/ocf_kubernetes/manifests/master/secrets.pp
```diff
@@ -0,0 +1,18 @@
+class ocf_kubernetes::master::secrets {
+
+  file {
+    default:
```
keur: we usually only use `default` when there is more then one file being created in the same resource definition

modules/ocf_puppet/manifests/puppetserver.pp
keur: masters probably don't need `post` to add secrets. we would do that manually

modules/ocf_puppet/manifests/puppetserver.pp
```diff
@@ -39,6 +39,18 @@
     notify               => Service['puppetserver'],
   }
 
+  # Allow kubernetes secrets to access kubernetes secrets
+  puppet_authorization::rule { 'kubernetes-secrets':
+    match_request_path   => '^/puppet/v3/file_(content|metadata)s?/kubernetes-secrets$',
```
keur: Why do we need content and metadata? I imagine one is for `get` and the other for `post`, though we probably don't need to `post` from masters, and mentioned in a previous comment

modules/ocf_kubernetes/manifests/master/secrets.pp
```diff
@@ -0,0 +1,18 @@
...
+      require => Package['kubectl'];
+
+    '/etc/ocf-kubernetes/secrets':
+      ensure  => 'directory',
+      source  => 'puppet:///kubernetes-secrets',
+      recurse => true,
+      mode    => '0700';
```
dkess: We should probably specify `purge => true` here as well.

fydai:
This is the best solution for now I guess, so reopening.

PR #485, opened 2019-01-13
mirrors: reorganize projects into their own directory

abizer:
this is a no-op change that just moves the project manifests into their own directory for organizational purposes


modules/ocf_mirrors/manifests/init.pp
```diff
@@ -1,32 +1,32 @@
 class ocf_mirrors {
-  include ocf::ssl::default
+  require ocf::ssl::default
```
dkess: What does this do?

abizer: https://github.com/ocf/puppet/pull/472#discussion_r244648872

PR #487, opened 2019-01-14
add the etc puppet share and put it on all hosts

dkess:
Here's my first candidate for this, based on recent discussion in IRC.

I realize there are other ways of doing this-- most notably, we could have Jenkins populate `/opt/shares/etc` instead of making it a `vscrepo`, but my worry with such a solution is that it means a newly provisioned Puppetserver would have an empty `etc` share, which is very bad.  This makes sure we always have something valid being put on all hosts.

However I'm still not sure if I like this solution. It doesn't seem so far off from just putting the `vcsrepo` on every host, and this PR adds another (albeit minor) layer of complexity to that.

Thoughts?

### DIFF
--- a/modules/ocf/manifests/etc.pp
+++ b/modules/ocf/manifests/etc.pp
@@ -1,0 +1,11 @@
+class ocf::etc {
+  file { '/etc/ocf':
+    ensure  => directory,
+    source  => 'puppet:///etc',
+    owner   => root,
+    group   => root,
+    mode    => '0755',
+    purge   => true,
+    recurse => true,
+  }
+}
--- a/modules/ocf/manifests/init.pp
+++ b/modules/ocf/manifests/init.pp
@@ -2,6 +2,7 @@ class ocf {
   include ocf::apt
   include ocf::auth
   include ocf::autologout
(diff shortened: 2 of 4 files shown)


dkess:
I mean, we're still relying on Github to do this, it's just happening on one host instead of all the others...

dkess:
Also I tested this on supernova and it works so an authorization rule isn't necessary.

abizer:
that's a significant improvement as far as I'm concerned

modules/ocf/manifests/etc.pp
```diff
@@ -0,0 +1,10 @@
+class ocf::etc {
+  file { '/etc/ocf':
+    ensure  => directory,
+    source  => 'puppet:///etc',
+    owner   => root,
+    group   => root,
+    mode    => '0755',
+    recurse => true,
```
jvperrin: I'd add `purge => true` on this too so that unmanaged files are deleted, since they wouldn't be with this current setup.

dkess:
I was thinking of doing something similar to the DNS repo, where Jenkins will trigger a puppet run on lightning on every push.

PR #488, opened 2019-01-14
examine-mail-log: increase alert threshold from 45 to 75 emails per day

abizer:
since we've been getting this alert a lot more often recently, perhaps we can increase it. It may also be worthwhile to export some of these metrics to prometheus/grafana and use that for alerting as well.


abizer:
sending username info to something relatively public like grafana would probably also be a privacy issue. instead perhaps we may be able to just send aggregate counts of emails sent per day?

dkess:
That's possible, which is why I used the word "reliably". I still think having a specialized script is fine for this, while also keeping a count of total emails sent from postfix.

PR #489, opened 2019-01-14
motd: reintroduce OS version shaming

abizer:


dkess:
Perhaps we should wait a bit before merging this?

abizer:
yeah this was mostly memery

jvperrin:
Yeah, I'd definitely wait on this too for like a year or something, probably like a 6 months after release? Shruuug

PR #490, opened 2019-01-16
packages: update brave

keur:
The Brave version on our desktop has been notifying us that the current
version is out-of-date. Firstly, the package was renamed from brave to
brave-browser. The upstream apt repository to add changed, along with
the gpg signing key. Brave recommends pulling the key from their server,
but instead I uploaded the key to our keyserver and fetch it from there.

The configuration for brave and the configuration options have also
completely changed. Instead of making modifications to
`~/.config/brave/session-store-1`, we now put configuration options in
`~/.config/BraveSoftware/Brave-Browser/Default/Preferences`. Since brave
is built on top of chromium, that means they inherit chromium's style of
asking to be the default browser, via a timestamp. With a small hack we
circumvent this. In the chromium code they will consider your config
option 'unparseable' if you pass a negative value to the timestamp, thus
it never asks to be the default browser.

The offending code is here:
https://github.com/chromium/chromium/blob/2ca8c5037021c9d2ecc00b787d58a31ed8fc8bcb/chrome/browser/ui/startup/default_browser_prompt.cc#L90-L91. Get fucked, Google :)


modules/ocf/manifests/packages/brave.pp
```diff
@@ -7,17 +7,35 @@
...
+  }
 
   # TODO: Change this to a global config
   # if/when Brave adds support for it.
   file {
-    '/etc/skel/.config/brave':
-      ensure  => directory,
-      require => [File['/etc/skel/.config'], Package['brave']];
+    default:
+      require =>  [
```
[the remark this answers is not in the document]
keur: unsure what you mean by this? chain the files, or chain the package to the file resource?

modules/ocf/manifests/packages/brave.pp
```diff
@@ -7,17 +7,35 @@
...
+        File['/etc/skel/.config'],
+        Package['brave-browser'],
+      ];
+
+    ['/etc/skel/.config/BraveSoftware/',
+    '/etc/skel/.config/BraveSoftware/Brave-Browser/',
+    '/etc/skel/.config/BraveSoftware/Brave-Browser/Default']:
```
kpengboy: Nit: indent these so quotes align

modules/ocf/manifests/packages/brave.pp
```diff
@@ -7,17 +7,35 @@
     stage =>  first,
   }
 
-  package { 'brave':; }
+  # New install instructions, per https://brave-browser.readthedocs.io/en/latest/installing-brave.html#linux
+  package { 'brave':
+    ensure => purged,
+  }
```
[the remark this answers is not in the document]
kpengboy: Or if you do keep these statements, you should mark them as temporary.

modules/ocf/manifests/packages/brave/apt.pp
```diff
@@ -1,13 +1,13 @@
 # Include Brave apt repo.
 class ocf::packages::brave::apt {
   apt::key { 'brave':
-    id     => '5AB4CA7AC2F93FE3EBD6082D6ED19DBB448EEE6C',
-    server => 'keyserver.ubuntu.com',
+    id     => 'D8BAD4DE7EE17AF52A834B2D0BB75829C2D4E821',
+    server => 'pgp.ocf.berkeley.edu',
```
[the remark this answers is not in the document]
kpengboy: Lol should we make that the default for the apt::key resource?

modules/ocf/templates/brave/Preferences.erb
```diff
@@ -0,0 +1,15 @@
...
+    "default_browser_infobar_last_declined": "-1",
+    "has_seen_welcome_page": true,
+    "show_home_button": true
+  },
+  "homepage": "https://ocf.berkeley.edu",
+  "homepage_is_newtabpage": false,
+  "session": {
+    "restore_on_startup": 4,
+    "startup_urls": [
+      "https://ocf.berkeley.edu/"
```
kpengboy: idem

modules/ocf/templates/brave/session-store-1.erb
```diff
@@ -1,23 +0,0 @@
...
-      "general.startup-mode": "homePage",
-      "privacy.autofill-enabled": false,
-      "security.fullscreen.content": "alwaysAllow",
-      "security.passwords.active-password-manager": "Unmanaged",
-      "shutdown.clear-all-site-cookies": true,
-      "shutdown.clear-autocomplete-data": true,
-      "shutdown.clear-autofill-data": true,
-      "shutdown.clear-cache": true,
-      "shutdown.clear-downloads": true,
-      "shutdown.clear-history": true
```
cg505: don't we still want some of these? are the equivalents in the new settings file?

keur: yes i want some (all) of these because i chose every setting here. these are no longer applicable and a lot don't have equivalents. the new settings GUI is also much harder to navigate. if you find a setting worth switching on lmk and i can add it. 

most the defaults have user privacy (default: ask user permission in all cases) in mind so it should be fine. the only thing i can think of adding that i've seen is clearing cookies on browser exit. disabling brave from asking to be the default browser is in the new config, there is no password manager because i opted not to install `brave-keyring`, so that should also take care of the autocomplete stuff.

we can have brave clear it's cache, downloads, history, etc on shutdown if we run it with the `--incognito` flag, but then the start screen is purple, so we don't get our homepage. it also seems a bit useless since `~/` will be deleted on user suspend, and even with all that stuff enabled, rootstaff could still get all this data if they poked around while someone was running a browser session.

PR #492, opened 2019-01-17
pgp: add sks CPU resource limit

keur:
CPUQuota's were added back in systemd 213, and it appears that stretch's
kernel is compiled with support for this. When our keyserver is peering
with other keyservers, the CPU cores pin to 100%, thus rendering search
on the website useless. This ensures that sks can only pin to 50% on
each core.


keur:
Note: if we want to impose a limit exactly how many CPUs it can use that can be done in `docker`, though I'd prefer not to tie code to the hardware on the machine unless absolutely necessary.

modules/ocf_keyserver/files/sks-keyserver.service
```diff
@@ -11,6 +11,7 @@ ExecStart=/usr/bin/docker run --rm \
     -p 127.0.0.1:11372:11371/tcp \
     --name sks-keyserver \
     zhusj/sks
+CPUQuota=50%
```
kpengboy: Have you verified that this quota will be forwarded to the Docker container?

keur: `cgroups` are hierarchical so I don't imagine this will be an issue. Once I test it I will leave a follow up comment though.

chriskuehl: The `docker run` command is just sending commands to the Docker daemon over the network, your container's processes are not a descendant of `docker run`. I would be surprised if this works.

keur: I see. Thanks for the comments. I think the newest commit addresses this.

chriskuehl:
> This ensures that sks can only pin to 50% on each core.

Doesn't this actually limit it to 50% across all cores? The docs suggest this is 50% total -- you can verify by running something like `systemd-run -p CPUQuota=25% -- python3.6 -c 'import os; os.fork(); list(range(1000000000))'` and observing that each process only gets ~12.5% CPU.

keur:
@chriskuehl lmao cant believe you just fork bombed me out of RAM

keur:
@chriskuehl Since I ran out of RAM before I could even open `htop`, I wrote a script that only uses CPU to test the same thing. On all the cores, it does use 25% CPU. If I use `cpu_count // 2` instead it still only uses half on the cores it was given. So it appears to be per core.

```python
import multiprocessing as mp

def loop_forever():
    while True:
        pass

p = mp.Pool(processes=mp.cpu_count())
for _ in range(mp.cpu_count()):
    p.apply_async(loop_forever)
p.close()
p.join()
```

chriskuehl:
I'm not sure I understand what you're trying to achieve, but I'm pretty sure it is not per-core but instead a limit across all cores.

```bash
ckuehl@dev45-uswest1adevc:/tmp/tmp.XCWQcRdy02$ cat test.py
#!/usr/bin/env python3.7
import os, sys
for _ in range(int(sys.argv[1]) - 1):
    os.fork()
while True:
    pass

$ sudo systemd-run -p CPUQuota=50% -- ./test.py 1
Running as unit run-r4fcc68bcf61c4cd4a0631745dec6e1ee.service.
# [Watch htop, there is one process using 50% CPU]

$ sudo systemd-run -p CPUQuota=50% -- ./test.py 2
Running as unit run-r4fcc68bcf61c4cd4a0631745dec6e1ee.service.
# [Watch htop, there are two processes using 25% CPU]

$ sudo systemd-run -p CPUQuota=50% -- ./test.py 3
Running as unit run-r4fcc68bcf61c4cd4a0631745dec6e1ee.service.
# [Watch htop, there are four processes using 12.5% CPU]
```

The systemd docs say the same thing:
```
CPUQuota=

Assign the specified CPU time quota to the processes executed. Takes a
percentage value, suffixed with "%". The percentage specifies how much CPU time
the unit shall get at maximum, relative to the total CPU time available on one
CPU. Use values > 100% for allotting CPU time on more than one CPU. This
controls the "cpu.max" attribute on the unified control group hierarchy and
"cpu.cfs_quota_us" on legacy. For details about these control group attributes,
see cgroup-v2.txt and sched-bwc.txt.
```

chriskuehl:
Also, won't capping the CPU just cause the process which is consuming all CPU to take longer to complete, thus impacting the service even more?

keur:
Well the webserver is running on the same machine, and when the key server is trying to peer you can't even execute a search query. Peering is not a particularly time-sensitive operation either. We could try giving the machine more cores.

abizer:
sks is single threaded, giving it more cores isn't going to help - I'm not sure the problem is one of peering as it is. I've sent keys to the server and searched for them using the HKP interface and had no problems. Maybe we should turn debugging up to 3 and see if there's anything in the logs.

jvperrin:
So now that the VM this is running on has 4 cores and sks is single threaded in the first place, do we need this any more? As far as I understand it, it had only one core before, so sks was taking all of that core and not giving any for the web server, but adding more cores has essentially fixed that issue. Is that correct?

keur:
@jvperrin It had 2 cores last night, and the webserver was still unable to run after I uploaded a key. Maybe something else was happening at the same time and I attributed it to sks?

abizer:
I added 4 cores 8GB of RAM to build the database, because that requires a lot of resources. running the server on its own was going fine on 2c/4gb until keur added his key last night, which perhaps might have been anomalous?

keur:
@cg505 has also experienced this, on a separate day.

chriskuehl:
> Well the webserver is running on the same machine, and when the key server is trying to peer you can't even execute a search query.

Isn't the webserver (not nginx, the thing actually generating the pages) running inside the same container? Giving the container fewer CPUs to work with seems like it will make things strictly worse if so.

abizer:
it's not the peering that's the problem - apparently there's a specific key that someone uploaded to the public network as a PoC for some bug that is propagating through the network because of gossipping. attempting to merge that key causes the server to crash and requires a restart.

PR #394, opened 2018-10-14
Change desktop theme to Arc

ckwok2021:
Changed the desktop and the greeter theme to Arc.

PR #493, opened 2019-01-19
Add preliminary support for desktops on buster

jvperrin:
I need to test this again sometime since this is still based off my work from last September, but I believe this should more or less work for buster desktops. I held off on submitting it then because it was messy with jessie, stretch, and buster all included, but it's nicer and cleaner now with just stretch/buster. I'm pleased to see that a few packages (`mssh`, `fcitx-hangul`, `pdfchain`, probably more) have been added since I tested out buster then, which is good news for us.

Last I tested, there were also issues with logging on to a desktop running on buster, so some more testing definitely has to be done there and see if we can narrow down if it is a bug or something we've misconfigured somehow. If it is indeed a bug, now is a great time to fix the issue before [the soft freeze next month](https://wiki.debian.org/DebianBuster#Release_and_updates). Logins were definitely fine on a buster VM, but something with lightdm or whatever could cause different effects on desktops for logins.

4.19 kernel, woohoo! :D So close to 4.20, but nope, 4.19 is the LTS, so it'll likely stay there (unfortunately for the memes).

~One step I'm missing here is a new desktop background, but we're already planning on updating that in #394, so I think foregoing it for now (or maybe for good) is a better move.~

~Note that there's still a TODO to only allow the lightdm access where needed (on desktops) instead of everywhere, so I'd like to address that before actually merging this (and retest on a desktop/VM of course).~


modules/ocf/manifests/extrapackages.pp
```diff
@@ -219,6 +216,30 @@
...
+        # instead and buster has PHP 7.3:
+        # http://php.net/manual/en/migration71.deprecated.php
+        'php-mcrypt',
+
+        # pkpgcounter comes from pykota, which is unmaintained upstream, so we
+        # need to find a replacement or forward port it for buster. We had this
+        # same problem with pykota from wheezy to jessie but with pykota as a
+        # broader package. Looks like pkpgcounter is finally joining pykota and
+        # we should move off it somehow and find some other way to count pages.
+        'pkpgcounter',
```
dkess: Where is this used by us? A Sourcegraph search gave me nothing.

jvperrin: Hmmm, yeah I can't find anything either, but I thought it was pretty integral to the printing pipeline. Then again, I'm not sure why it's in `ocf::extrapackages` then and not in just `ocf_printhost`...

I'll try and do some more digging

jvperrin: Ok yeah I think we're just counting the pages using postscript now with [ocf_printhost/files/enforcer-pc](https://github.com/ocf/puppet/blob/master/modules/ocf_printhost/files/enforcer-pc), so that's great. We can totally remove this then, I don't believe it's used anywhere. Thanks for pointing that out!

modules/ocf_desktop/manifests/init.pp
```diff
@@ -7,7 +7,9 @@
   include ocf::packages::pulse
   include ocf::packages::vscode
 
-  include ocf_mesos::slave
```
kpengboy: Are we still intending to use desktops as media slaves at all, given that we're moving to Kubernetes?

dkess: I don't plan to do this for Kubernetes.

kpengboy: *mesos

modules/ocf/manifests/extrapackages.pp
```diff
@@ -114,15 +114,13 @@
     'php-curl',
     'php-gd',
     'php-mbstring',
-    'php-mcrypt',
     'php-mysql',
     'php-sqlite3',
     'php-soap',
     'php-xml',
     'php-zip',
     'php7.0-dba', # This isn't available as php-dba unfortunately
```
fydai: php-dba is avaliable in buster now

cg505:
i believe there is another series of changes in `ocf/buster-desktop` that perhaps should be incorporated into this PR

modules/ocf_desktop/manifests/init.pp
kpengboy: Might as well completely get rid of this code then

jvperrin: It's completely gone as of #680 anyway

modules/ocf/manifests/extrapackages.pp
jvperrin: Bleh it's a virtual package though, which means puppet will try and install it on every run :/

jvperrin: Ah I see you figured that out anyway, given https://github.com/ocf/puppet/commit/87e1a30e9d20e840029c986540f1b276d179e5ce

jvperrin:
@cg505 Good point, I didn't notice that branch since it was made after I originally opened this review. I've merged in any changes on that branch that aren't already here (cherry-picked a couple commits).

To be clear, my goal with this PR is to make it so that buster desktops are supported alongside the existing stretch ones, so any packages that have naming differences, etc. should have conditionals in puppet (which is a different approach from the `buster-desktop` branch from what I can tell). This allows for gradually switching over desktops, starting with eruption for example, then switching a few more, then a whole row, then multiple rows, and the full lab if everything goes smoothly. This worked well in the past since it means that if something is found that is broken, it doesn't affect the whole lab and can be fixed on just the desktops running the newer version.

modules/ocf_desktop/manifests/packages.pp
```diff
@@ -34,21 +34,39 @@
...
     # notifications
     ['libnotify-bin', 'notification-daemon']:;
-    # performance improvements
-    ['preload']:;
     # security tools
     ['scdaemon']:;
     # utilities
     ['wakeonlan']:;
     # Xorg
-    ['xclip', 'xsel', 'xserver-xorg', 'xscreensaver', 'freerdp-x11']:;
```
ethanhs: I have used rdesktop but personally prefer freerdp. Freerdp2-x11 is available in both stretch backports and buster, so perhaps we can use that?

jvperrin: Ah, good idea, yeah that should work ok for both

jvperrin:
Ok, this is ready for review now and tested on a buster desktop (albeit a virtual one over VNC, but it works).

jvperrin:
Fair enough, I'd also like to test it on an existing stretch desktop to make sure it doesn't break anything there, since I haven't done that yet.

jvperrin:
This has been tested on a couple real desktops (`surge` and `blizzard`) and also on a desktop running stretch, so I think it's ready to merge.

PR #497, opened 2019-01-23
labstats: remove low-toner-alert

dkess:
This is no longer needed, since we monitor the printers with Prometheus.


dkess:
The Prometheus alerting system is less prone to false positives ("-2 toner") since it takes the median over a time period

PR #498, opened 2019-01-23
puppet: allow all hosts to access the etc volume

dkess:
This will stop the puppetspam errors from hosts that don't have permission.


jvperrin:
lol how did this work before then?

dkess:
I only tested it on supernova

jvperrin:
But why did it work on supernova?

dkess:
lol not sure

PR #502, opened 2019-01-24
Revert "Preventing keyring popup on brave"

keur:
Reverts ocf/puppet#500

While this does stop the keyring, Brave will now ask you to save your password on every page.


keur:
Sorry about that. Should've tested the original PR before merging. `gnome-keyring` is no longer on desktops, so this is no longer necessary.

PR #509, opened 2019-01-26
Fix superkey map

glee-:
amend the feature proposed in #385


dkess:
What exactly does this do? Does it re-enable super+arrow key window snapping?

glee-:
it allows the user to press the super to open whiskermenu, as well as super+arrow for window snapping

cg505:
@glee- actually can we do some weirder key combo like `Super_L=Control_R|Alt_L|Shift_R`? Turns out I have a lot of key combos that use left control + left alt + left shift and these all now break and/or open the menu.

glee-:
Thanks for the feedback. Ctrl_L/R don't seem to be able to be differentiated in the configs.
Hopefully nobody uses `ctrl + alt + shift_r`

PR #510, opened 2019-01-26
ocf_kubernetes: Allow ocfroot to use kubectl

keur:
Give ocfroot read access to /etc/kubernetes/admin.conf, which has the
client cert required to authenticate as a "kubernetes-admin". Now
ocfroot can administer the cluster without becoming the root user.

Create a smaller helper script called 'ocf-kubectl' that sets the
KUBECONFIG environment variable and proxies the rest of the arguments to
kubectl.

### DIFF
--- a/modules/ocf_kubernetes/manifests/master.pp
+++ b/modules/ocf_kubernetes/manifests/master.pp
@@ -95,6 +95,27 @@ class ocf_kubernetes::master {
     },
   }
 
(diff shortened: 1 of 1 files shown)


dkess:
I guess this is fine temporarily, but I'd rather have a mechanism where `ocfroot` users are forced to use their Kerberos root credential to get these privileges. Also the mechanism should be able to work from supernova.

keur:
This was a temporary fix for #539. No longer necessary.

PR #511, opened 2019-01-26
puppet: change etc volume to only include configs

dkess:
Anyone have thoughts on this? I think it would be nicer.

This would mean, for example, that `hours.yaml` would be located at `/etc/ocf/hours.yaml` instead of `/etc/ocf/configs/hours.yaml`. The other files don't really need to be on every host.


keur:
@dkess confirmed not to be paying attention during BoD

PR #512, opened 2019-01-27
some useful shorturls + bit of sp19 cleanup

abizer:
> some of these like the family and infra ones can be re-added
when we have those ready, but we may want to consider, for example,
making the infra one a canonical shorturl to the most up-to-date
slide deck on how the infra works.


dkess:
Re: calendar URLs. Is there a plan to document these somewhere or otherwise write them down? Otherwise they'll never get used.

Re: removing old slide decks: I get why we want to do this, but I'd rather not break the links to old slides in case people reading through the announce archives want to go back and read them. Perhaps a better solution to this would be to introduce some categorization of shorturls, so we don't have to worry about "clutter"?

abizer:
re: categorization, let's just get this shorturls thing done once and for all: https://github.com/ocf/ircbot/pull/94. we've spent too much time bikeshedding it and not actually implementing something that works for us.

PR #513, opened 2019-01-29
etc: add force and nobackup options

dkess:
`backup` is definitely required, but `force` is required too since we need to remove residual directories from the old state of the `etc` folder on hosts.

Should `force => true` stay on as an option though? Seems like it should, I can't really think of how it could harm us, though I'd like to exercise caution here since the word "force" is scary...

### DIFF
--- a/modules/ocf/manifests/etc.pp
+++ b/modules/ocf/manifests/etc.pp
@@ -7,5 +7,7 @@ class ocf::etc {
     mode    => '0755',
     purge   => true,
     recurse => true,
+    backup  => false,
+    force   => true,
   }
 }


dkess:
Results of testing on supernova: without the `force` option, this does nothing because it doesn't want to delete the old files. With the `force` option, everything seems to work properly.

jvperrin:
Please also remove the permission setting part, they shouldn't have execute permissions when not necessary (although they are all yaml files right now so that doesn't especially matter)

dkess:
Done in 8a1db266e4302059132cebc150161649fe013039

PR #514, opened 2019-02-01
add class for specifying additional custom apt repositories

cg505:
This is mostly useful for staffvms, where you might want to add additional repositories that aren't applicable to any other OCF infra. If you add them manually, puppet will delete them every time it runs. This provides a way to specify them in hieradata.

This is probably not the best approach, but I don't know enough about puppet to come up with something better. Reviews greatly appreciated.

### DIFF
--- a/hieradata/nodes/failure.yaml
+++ b/hieradata/nodes/failure.yaml
@@ -1,1 +1,15 @@
+classes:
+  - ocf::packages::docker
+  - ocf::packages::custom::apt
+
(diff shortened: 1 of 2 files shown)


jvperrin:
(also not sure how that would work with our current first stage setup, it likely wouldn't, but I'd say it's worth a try)

cg505:
damn that totally works. at least these was an interesting exercise in how puppet works.

PR #516, opened 2019-02-01
backups: switch from lftp to curl for box backup

dkess:
lftp was causing problems for us, and curl is what's recommended by box.com's official documentation. See https://community.box.com/t5/Managing-Content-Troubleshooting/I-m-Having-Trouble-Using-FTP-with-Box/ta-p/323.

I did a mini test of this and can confirm the files begin to upload. Since the next backup will be tomorrow, I canceled that test in favor of a proper "prod" test. So hopefully this can get merged before then.

### DIFF
--- a/modules/ocf_backups/files/upload-to-box
+++ b/modules/ocf_backups/files/upload-to-box
@@ -6,6 +6,8 @@ import subprocess
 import sys
 import time
 from io import BytesIO
+from itertools import chain
+from tempfile import NamedTemporaryFile
 
 import pycurl
 import pymysql
@@ -20,9 +22,8 @@ BOX_EMAILS = [
     'jvperrin@ocf.berkeley.edu',
     'kpeng3.4@berkeley.edu',
 ]
-BOX_FTP = 'ftps://ftp.box.com:990'
+BOX_FTP = 'ftp://ftp.box.com'
 FOLDER = 'ocf-backup-' + time.strftime('%Y-%m-%d')
-MAX_CONCURR_UPLOADS = 10
 
 # To get the OAuth keys for an account:
 #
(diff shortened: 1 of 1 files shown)


modules/ocf_backups/files/upload-to-box
```diff
@@ -107,30 +109,43 @@ def get_access_token(creds):
...
+    netrc_creds = 'machine ftp.box.com login {email} password {passwd}'.format(
         email=creds['email'],
         passwd=creds['password'],
-        concurr=MAX_CONCURR_UPLOADS,
-        archive=ARCHIVE,
-        folder=FOLDER,
     )
+    netrc_file = NamedTemporaryFile(mode='w')
+    netrc_file.write(netrc_creds)
+    netrc_file.flush()
```
jvperrin: This should get automatically deleted when the program ends, but I think putting a comment about it getting automatically deleted on cleanup would be good since I'm not sure how many people know how `NamedTemporaryFile` works. Also, making it into a context manager so it's more explicit when it's used and when it's closed would probably be good, and then it can just be put around the one `curl` that needs it:

```
# This temporary file will get deleted when it is closed (when this context ends),
# keeping the credentials from sticking around in a more permanent file
with NamedTemporaryFile(mode='w') as netrc_file:
    netrc_file.write(netrc_creds)
    netrc_file.flush()

    # curl stuff
```

modules/ocf_backups/files/upload-to-box
```diff
@@ -107,30 +109,43 @@ def get_access_token(creds):
...
+            '-1',
+            '--disable-epsv',
+            '--ftp-skip-pasv-ip',
+            '--ftp-ssl',
+            '--netrc-file', netrc_file.name,
+            '--ftp-create-dirs',
+            '--globoff',
+        ] + file_args,
         stdout=stdout_pipe,
     ).check_returncode()
```
redplanks: Nit: you can do `check=True` in the `subprocess.run` args to have the same effect.

modules/ocf_backups/files/upload-to-box
```diff
@@ -107,30 +109,43 @@ def get_access_token(creds):
...
-        archive=ARCHIVE,
-        folder=FOLDER,
     )
+    netrc_file = NamedTemporaryFile(mode='w')
+    netrc_file.write(netrc_creds)
+    netrc_file.flush()
+
+    # Specify each file and its destination on the command line
+    # -T file1 ftp://.../destfolder/ -T file2 ftp://.../destfolder/ ...
+    file_args = list(chain.from_iterable(
```
ethanhs: Nit: if you change the code below to
```
[
    'curl',
    '-1',
    '--disable-epsv',
    '--ftp-skip-pasv-ip',
    '--ftp-ssl',
    '--netrc-file', netrc_file.name,
    '--ftp-create-dirs',
    '--globoff',
    *file_args,
],
```

You don't need to wrap this in `list()`

modules/ocf_backups/files/upload-to-box
dkess: Perhaps explicitly passing `delete=True` here would be a bit better than writing a comment.

dkess:
Thanks everyone for the feedback! Just responded to everything.

modules/ocf_backups/files/upload-to-box
```diff
@@ -107,32 +108,47 @@ def get_access_token(creds):
...
+                '--disable-epsv',
+                '--ftp-skip-pasv-ip',
+                '--ftp-ssl',
+                '--netrc-file', netrc_file.name,
+                '--ftp-create-dirs',
+                '--globoff',
+                *file_args,
+            ],
+            check=True,
+            stdout=stdout_pipe,
```
cg505: curl sends progress output to stderr

PR #518, opened 2019-02-04
Add asciinema

ethanhs:


jvperrin:
Do we need this on all hosts, or should it just be on login servers? My vote would be login servers (`extrapackages` instead of `packages`), but idk. Thoughts?

cg505:
agree

ethanhs:
Moved to extrapackages in  b1f4ed2.

PR #519, opened 2019-02-05
Add select Debian settings for firefox-esr

ethanhs:
Based on the diff the changes Debian made: https://i.fluffy.cc/xqGM9D9bpZ8DRZqttzhqHTtt6Xpx0Xbq.html

Feel free to dispute any of the changes I made.

### DIFF
--- a/modules/ocf/templates/firefox/prefs.js.erb
+++ b/modules/ocf/templates/firefox/prefs.js.erb
@@ -1,3 +1,25 @@
+// This is the Debian specific preferences file for Firefox ESR
+// You can make any change in here, it is the purpose of this file.
+// You can, with this file and all files present in the
+// /etc/firefox-esr directory, override any preference you can see in
+// about:config.
+//
+// Note that pref("name", value, locked) is allowed in these
+// preferences files if you don't want users to be able to override
+// some preferences.
+
+pref("extensions.update.enabled", true);
+// Use LANG environment variable to choose locale
+pref("intl.locale.requested", "");
+
+// Default to classic view for about:newtab
+pref("browser.newtabpage.enhanced", false, sticky);
+
+// Default to no suggestions in the urlbar. This still brings a panel asking
+// the user whether they want to opt-in on first use.
+pref("browser.urlbar.suggest.searches", false);
+
+// OCF modifications made after here
 lockPref("browser.startup.homepage", "<%= @browser_homepage %>");
 pref("browser.cache.disk.capacity", 0);
 pref("browser.download.useDownloadDir", false);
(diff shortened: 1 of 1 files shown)


modules/ocf/templates/firefox/prefs.js.erb
```diff
@@ -1,3 +1,25 @@
...
+
+// Default to classic view for about:newtab
+pref("browser.newtabpage.enhanced", false, sticky);
+
+// Default to no suggestions in the urlbar. This still brings a panel asking
+// the user whether they want to opt-in on first use.
+pref("browser.urlbar.suggest.searches", false);
+
+// OCF modifications made after here
 lockPref("browser.startup.homepage", "<%= @browser_homepage %>");
```
kpengboy: `lockPref` is going away in a future Firefox version too, we ought to change this too

modules/ocf/templates/firefox/prefs.js.erb
```diff
@@ -1,3 +1,25 @@
...
+pref("extensions.update.enabled", true);
+// Use LANG environment variable to choose locale
+pref("intl.locale.requested", "");
+
+// Default to classic view for about:newtab
+pref("browser.newtabpage.enhanced", false, sticky);
+
+// Default to no suggestions in the urlbar. This still brings a panel asking
+// the user whether they want to opt-in on first use.
+pref("browser.urlbar.suggest.searches", false);
```
cg505: Our concept of "first use" is not the same as most people using Debian. We might not want this.

modules/ocf/templates/firefox/prefs.js.erb
```diff
@@ -1,3 +1,25 @@
+// This is the Debian specific preferences file for Firefox ESR
+// You can make any change in here, it is the purpose of this file.
+// You can, with this file and all files present in the
+// /etc/firefox-esr directory, override any preference you can see in
+// about:config.
+//
+// Note that pref("name", value, locked) is allowed in these
+// preferences files if you don't want users to be able to override
+// some preferences.
```
keur: Even though Debian suggested these changes, I don't think there's any benefit in differentiating between these and "OCF prefrences." Since everything is hand-picked, they are all "OCF changes"

modules/ocf/templates/firefox/prefs.js.erb
```diff
@@ -1,3 +1,25 @@
...
+
+pref("extensions.update.enabled", true);
```
keur: All extensions live in `~/.config`, which gets cleared on logout. While this doesn't hurt us, it's kinda irrelevant since extensions are per-session and hence are always up-to-date. This setting is targeted at personal computers.

modules/ocf/templates/firefox/prefs.js.erb
keur: This could end up being annoying since this setting is also targeted at personal computers.

PR #520, opened 2019-02-06
mesos: pin to v1.7.0-2.0.3

abizer:
the latest from upstream requires glibc 2.27, whereas stretch provides 2.24.

### DIFF
--- a/modules/ocf_mesos/manifests/package.pp
+++ b/modules/ocf_mesos/manifests/package.pp
@@ -3,11 +3,20 @@ class ocf_mesos::package {
     stage => first,
   }
 
+  # at this time, v1.7.1 requires glibc >= 2.27 but stretch provides 2.24
+  $mesos_version = '1.7.0-2.0.3'
+
   # We need Java 8 to be the default java.
   include ocf::packages::java
   ocf::repackage { 'mesos':
     # mesos recommends zookeeper
-    recommends => false,
+    recommends => false;
+  } ->
+  apt::pin { 'mesos':
+    ensure   => present,
+    packages => ['mesos'],
+    priority => 1001,
+    version  => $mesos_version;
   }
 
   # TODO: if we decide that we *won't* have servers which are both masters and


modules/ocf_mesos/manifests/package.pp
```diff
@@ -3,11 +3,20 @@
...
   }
 
+  # at this time, v1.7.1 requires glibc >= 2.27 but stretch provides 2.24
+  $mesos_version = '1.7.0-2.0.3'
+
   # We need Java 8 to be the default java.
   include ocf::packages::java
   ocf::repackage { 'mesos':
     # mesos recommends zookeeper
-    recommends => false,
```
jvperrin: I think you can still leave the trailing comma here?

modules/ocf_mesos/manifests/package.pp
```diff
@@ -3,11 +3,20 @@
     stage => first,
   }
 
+  # at this time, v1.7.1 requires glibc >= 2.27 but stretch provides 2.24
+  $mesos_version = '1.7.0-2.0.3'
```
jvperrin: I'd probably just put this into the one place it's used and put it with a `TODO`, since we'd still like to unpin at some point in the future and shouldn't need this as a long-term option.

kpengboy: I thought we were migrating to Kubernetes

abizer:
we are, but we still need marathon to work until it's ready, and with the
update last night, marathon was 502'ing

modules/ocf_mesos/manifests/package.pp
abizer: this is a holdover from something else I did, I replaced it with the semicolon

abizer:
going to merge this to try to fix the current deploy issues on the agents

modules/ocf_mesos/manifests/package.pp
jvperrin: Sure, but that doesn't really seem relevant to the version pinning

PR #522, opened 2019-02-09
Set node_exporter textfile collector directory option in Hiera.

redplanks:
Since `prometheus::node_exporter` is [included as a resource delcaration twice](https://sourcegraph.ocf.berkeley.edu/search?q=node_exporter), which isn't allowed by Puppet, it is failing on fallingrocks and *-whiteout.

I opted to move the textfile collector directory option to Hiera, but I could also have moved node_exporter's `version` option to `common.yaml`. Opinions?

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -87,3 +87,6 @@ kubernetes::master_nodes:
     - coup
     - autocrat
     - deadlock
+
+# Prometheus config
+prometheus::node_exporter::version: 0.17.0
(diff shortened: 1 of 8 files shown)


redplanks:
I'm a little concerned that if the version is specified in hieradata, we may be less likely to remember to bump it; but maybe that's just me.

dkess:
Agree with @jvperrin that the comment isn't necessary. Thanks for doing this, hieradata makes more sense for this kind of thing.

I don't think remembering to update applications is dependent on where in Puppet the version is listed. If anything, it would make sense to move more of our version numbers to hieradata, since that makes it easier to centralize this information.

If we're worried about forgetting to update stuff (which is a valid concern), we should think of new systems in place to ensure we get reminders... I have some loose ideas in my head about how we could do this, though that discussion belongs somewhere else.

PR #524, opened 2019-02-12
Enable http firewall rule on kubernetes masters

dkess:
This should finally end our woes related to Kubernetes firewall issues. This is tested (all masters are currently on my environment) and it appears to work.


dkess:
No, these are public URLs for kubernetes services

kpengboy:
...So is this like Marathon-lb's public endpoint?

dkess:
Yes

PR #525, opened 2019-02-12
add redeclipse to desktop packages

cg505:
I've wanted to do this for a while, since Red Eclipse is in the Debian repos, but unfortunately there was a key repeat bug that made it pretty much unplayable. I finally got the the bottom of it. It turns out there was a bug in SDL2 2.0.5 that caused this. The fix is https://hg.libsdl.org/SDL/rev/b48d8a98e261. I rebuilt libsdl2-2.0-0 with this patch applied and pushed it to the internal OCF apt repo as version 2.0.5+dfsg1-2. This means redeclipse will be buggy until someone runs apt-dater. Alternatively, if you notice the bug when playing, just run `sudo apt install libsdl2-2.0-0` to fix.

### DIFF
--- a/modules/ocf_desktop/manifests/packages.pp
+++ b/modules/ocf_desktop/manifests/packages.pp
@@ -29,7 +29,8 @@ class ocf_desktop::packages {
     # FUSE
     ['fuse', 'exfat-fuse']:;
     # games
-    ['armagetronad', 'gl-117', 'gnome-games', 'minecraft-launcher', 'wesnoth', 'wesnoth-music']:;
+    ['armagetronad', 'gl-117', 'gnome-games', 'minecraft-launcher', 'redeclipse',
+      'wesnoth', 'wesnoth-music']:;
     # graphics/plotting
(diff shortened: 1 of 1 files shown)


abizer:
for backporting packages, we usually append `ocf<n>` to the version number:
https://www.ocf.berkeley.edu/docs/staff/procedures/backporting-packages/

jvperrin:
Specifically adding `~ocf<n>` to the package version is done so that when a newer upstream version comes along (or one from backports), even if it is the same version, it is preferred over our backported version. For patches though, applying `ocf<n>` means that it'll be preferred over any upstream versions, so I agree with @abizer that adding `ocf<n>` to the package would be a good idea so your changes aren't overwritten if a new version is released upstream without the fix.

cg505:
The fix is minor so honestly if a new version comes along without the fix, we should probably install that, notice it's missing the patch, and then patch it. I will make the change.

cg505:
Also thanks for the link lol, I thought we probably had a guide for this but I couldn't find the docs so I just used some random one on the Debian site.

cg505:
Backported properly. Merging.

PR #527, opened 2019-02-15
tell xfce to use a compositor

cg505:
this fixes window tearing


dkess:
Have you tested this with 3D games? Dual monitors?

cg505:
This has been enabled in my config since September and I have never had a problem.

PR #529, opened 2019-02-19
kubernetes: add the cert signing script for user auth

dkess:
This adds a script called `certsign`, which can be used to authenticate OCF staffers to Kubernetes. "Clients" can generate a keypair and send it to this script to get it signed with their username. Then the generated certificate can be used to authenticate with Kubernetes.

The script is protected by `sudo` and `SUDO_USER` (similar to `makemysql`), so users can only get certs signed with their own username.

This is tested and working, but it essentially does nothing since we don't yet grant any permissions to OCF users.

### DIFF
--- a/modules/ocf/manifests/ssl/lets_encrypt/dns_common.pp
+++ b/modules/ocf/manifests/ssl/lets_encrypt/dns_common.pp
@@ -8,7 +8,9 @@ class ocf::ssl::lets_encrypt::dns_common {
 
   # we use this package in the le_cert_info fact, to gather information about
   # whatever certs are currently present
-  package { 'python3-cryptography':; }
+  ocf::repackage { 'python3-cryptography':
+    backport_on => ['stretch'],
+  }
 
   $letsencrypt_ddns_key = assert_type(Stdlib::Base64, lookup('letsencrypt::ddns::key'))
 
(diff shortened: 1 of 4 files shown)


dkess:
This is basically the same as https://github.com/ocf/krb-certsign/pull/1, except instead of being a "web app", this is accessible over SSH, using sudo for authentication.

modules/ocf/manifests/ssl/lets_encrypt/dns_common.pp
```diff
@@ -8,7 +8,9 @@
 
   # we use this package in the le_cert_info fact, to gather information about
   # whatever certs are currently present
-  package { 'python3-cryptography':; }
+  ocf::repackage { 'python3-cryptography':
+    backport_on => ['stretch'],
+  }
```
cg505: Take a gander at https://github.com/ocf/puppet/pull/449#discussion_r247018609 if you haven't. If we are directly using this for things besides letsencrypt, we should probably just move it to modules/ocf/manifests/packages.pp or something.

dkess: I don't think this is as necessary for this repo, since it's not running as a fact. It only gets run by users, so we don't have to worry about it running during the bootstrap process.

dkess: However since it's depended upon in more than one place, moving it to `packages.pp` is the right move, I'll do that

PR #530, opened 2019-02-20
add archlinuxcn to mirror server

yvbbrjdr:
Don't merge yet. Need to add health check and put credentials into puppet master

### DIFF
--- a/modules/ocf_mirrors/files/project/archlinuxcn/sync-archive
+++ b/modules/ocf_mirrors/files/project/archlinuxcn/sync-archive
@@ -1,0 +1,5 @@
+#!/bin/sh -eu
+
+/usr/local/bin/rsync-no-vanished -rtlH --delete-after --delay-updates \
+	--safe-links --password-file=/opt/mirrors/project/archlinuxcn/sync_password \
+	ucb-ocf@sync.repo.archlinuxcn.org::repo /opt/mirrors/ftp/archlinuxcn
--- a/modules/ocf_mirrors/manifests/projects/archlinuxcn.pp
+++ b/modules/ocf_mirrors/manifests/projects/archlinuxcn.pp
@@ -1,0 +1,34 @@
+class ocf_mirrors::projects::archlinuxcn {
+  file {
+    '/opt/mirrors/project/archlinuxcn':
+      ensure  => directory,
+      source  => 'puppet:///modules/ocf_mirrors/project/archlinuxcn/',
+      owner   => mirrors,
+      group   => mirrors,
+      mode    => '0755',
+      recurse => true;
+  }
+
+  file {
+    '/opt/mirrors/project/archlinuxcn/sync_password':
+        content   => lookup('mirrors::archlinuxcn_sync_password'),
+        show_diff => false,
+        owner     => mirrors,
+        group     => mirrors,
+        mode      => '0400';
+  }
+
+  ocf_mirrors::monitoring { 'archlinuxcn':
+    type          => 'unix_timestamp',
+    upstream_host => 'mirrors.tuna.tsinghua.edu.cn',
+    upstream_path => 'archlinuxcn',
+    ts_path       => 'lastupdate',
+  }
+
+  ocf_mirrors::timer {
+    'archlinuxcn':
+      exec_start => '/opt/mirrors/project/archlinuxcn/sync-archive',
+      hour       => '0/4',
+      minute     => '00';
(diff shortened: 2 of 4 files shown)


modules/ocf_mirrors/files/project/archlinuxcn/sync-archive
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh -eu
+
+/usr/local/bin/rsync-no-vanished -rtlH --delete-after --delay-updates --safe-links \
+	ucb-ocf@sync.repo.archlinuxcn.org::repo /opt/mirrors/ftp/archlinuxcn
```
abizer: where do the credentials for this come from

yvbbrjdr: They sent me the password via email. It's an environment variable that is not suitable to be stored publicly on GitHub.

modules/ocf_mirrors/manifests/projects/archlinuxcn.pp
```diff
@@ -0,0 +1,17 @@
...
+  file {
+    '/opt/mirrors/project/archlinuxcn':
+      ensure  => directory,
+      source  => 'puppet:///modules/ocf_mirrors/project/archlinuxcn/',
+      owner   => mirrors,
+      group   => mirrors,
+      mode    => '0755',
+      recurse => true;
+  }
+
```
yvbbrjdr: @abizer They said they'll add a lastsync timestamp in a few days for us to monitor updates.

modules/ocf_mirrors/files/project/archlinuxcn/sync-archive
dkess: This should go in the Puppet private share, I can explain to you how to do this at a later point

yvbbrjdr:
@abizer @dkess Health monitoring added.

modules/ocf_mirrors/manifests/projects/archlinuxcn.pp
```diff
@@ -0,0 +1,33 @@
...
+  ocf_mirrors::monitoring { 'archlinuxcn':
+    type          => 'unix_timestamp',
+    upstream_host => 'mirrors.tuna.tsinghua.edu.cn',
+    ts_path       => 'lastupdate',
+  }
+
+  ocf_mirrors::timer {
+    'archlinuxcn':
+      exec_start => '/opt/mirrors/project/archlinuxcn/sync-archive',
+      hour       => '00,04,08,12,16,20',
```
dkess: I'd recommend doing '0/4' on this line which will have the same effect

modules/ocf_mirrors/manifests/projects/archlinuxcn.pp
```diff
@@ -0,0 +1,33 @@
...
+        content   => lookup('mirrors::archlinuxcn_sync_password'),
+        show_diff => false,
+        owner     => mirrors,
+        group     => mirrors,
+        mode      => '0400';
+  }
+
+  ocf_mirrors::monitoring { 'archlinuxcn':
+    type          => 'unix_timestamp',
+    upstream_host => 'mirrors.tuna.tsinghua.edu.cn',
```
dkess: I don't think this will work-- you probably have to set the `upstream_path` here to `archlinuxcn` (similar to [here](https://github.com/ocf/puppet/blob/master/modules/ocf_mirrors/manifests/projects/manjaro.pp))

dkess:
Left a few comments. Ideally we'd test this before merging, but since fallingrocks is on abizer's environment that might be difficult. If we can't take it off that environment by the time you make the fixes, we can just merge this without testing.

modules/ocf_mirrors/manifests/projects/archlinuxcn.pp
yvbbrjdr: Fixed.

modules/ocf_mirrors/manifests/projects/archlinuxcn.pp
yvbbrjdr: Fixed.

yvbbrjdr:
@dkess It's actually not that urgent. We can test and merge it at any time as we see fit, even after we migrate fallingrocks.

yvbbrjdr:
Since the mirror server has been migrated, please merge it.

PR #534, opened 2019-02-24
Add mastodon .well-known redirect

fydai:


abizer:
you may want to fix this git history

jvperrin:
I'm a little confused, why does mastodon need access to `.well-known/host-meta` on death when it's on a separate domain?

dkess:
It uses this to verify that we own the ocf.berkeley.edu domain

PR #535, opened 2019-02-27
Allows ocfstaff to list iptables, with packet counts

mcint:


dkess:
Did a staffer request this? I suppose it's fine, hopefully this doesn't leak anything important...

modules/ocf/templates/sudoers.erb
```diff
@@ -27,7 +27,7 @@ ocfdeploy ALL=NOPASSWD: /usr/local/sbin/puppet-trigger
 #includedir /etc/sudoers.d
 
 # Allow ocfstaff to run a few commands without password for easier scripting.
-%ocfstaff ALL=(ALL) NOPASSWD: /usr/bin/virsh list, /usr/bin/virsh list --all
+%ocfstaff ALL=(ALL) NOPASSWD: /usr/bin/virsh list, /usr/bin/virsh list --all, /sbin/iptables -L[ -vx]*
```
kpengboy: Let me evaluate this, I don't think it does what it seems...

fydai:
closed because it's not that necessary and it could be a security issue

PR #536, opened 2019-02-28
add kanboard kubernetes nfs pvs

cg505:
This also slightly restructures how we define PVs (put them all in a directory).


cg505:
I agree also. I'll open an issue for that.

PR #539, opened 2019-03-04
Add kubectl default config

dkess:
This adds a default, working, `KUBECONFIG` file that uses the cert generation script. Now staffers can use `kubectl` on supernova!

### DIFF
--- a/modules/ocf/files/shell/config.fish
+++ b/modules/ocf/files/shell/config.fish
@@ -14,6 +14,7 @@ set -x PATH /opt/share/utils/bin \
     /usr/lib \
     /sbin \
     /usr/games
+set -x KUBECONFIG /etc/kubectl.conf
 
 if test -e /opt/puppetlabs/bin
     set -x PATH $PATH /opt/puppetlabs/bin
--- a/modules/ocf/files/shell/config.fish
+++ b/modules/ocf/files/shell/config.fish
@@ -14,6 +14,7 @@ set -x PATH /opt/share/utils/bin \
     /usr/lib \
     /sbin \
     /usr/games
+
 set -x KUBECONFIG /etc/kubectl.conf
 
 if test -e /opt/puppetlabs/bin
--- a/modules/ocf/manifests/packages/kubectl.pp
+++ b/modules/ocf/manifests/packages/kubectl.pp
@@ -1,0 +1,12 @@
+class ocf::packages::kubectl {
+  include ocf::packages::kubernetes
+
+  package { 'kubectl':; }
+
+  $cluster_cert_base64 = base64('encode', lookup('kubernetes::kubernetes_ca_crt'))
+  file {
+    '/etc/kubectl.conf':
+      content => template('ocf/kubectl/kubectl.conf.erb'),
+      mode    => '0644',
+  }
+}
--- a/modules/ocf/manifests/packages/kubectl.pp
+++ b/modules/ocf/manifests/packages/kubectl.pp
@@ -4,6 +4,7 @@ class ocf::packages::kubectl {
   package { 'kubectl':; }
 
   $cluster_cert_base64 = base64('encode', lookup('kubernetes::kubernetes_ca_crt'))
+
   file {
     '/etc/kubectl.conf':
       content => template('ocf/kubectl/kubectl.conf.erb'),
--- a/modules/ocf/templates/kubectl/kubectl.conf.erb
+++ b/modules/ocf/templates/kubectl/kubectl.conf.erb
@@ -1,5 +1,5 @@
 users:
-- name: ocfstaffer
+- name: kerberos-staff-login
   user:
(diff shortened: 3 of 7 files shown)


modules/ocf/files/shell/config.fish
```diff
@@ -14,6 +14,7 @@ set -x PATH /opt/share/utils/bin \
     /usr/lib \
     /sbin \
     /usr/games
+set -x KUBECONFIG /etc/kubectl.conf
```
keur: nit: newline

cg505:
Think you just need to rebase to fix the tests.

modules/ocf/templates/kubectl/kubectl.conf.erb
```diff
@@ -0,0 +1,20 @@
...
+    certificate-authority-data: <%= @cluster_cert_base64.gsub(/\n/, "") %>
+    server: https://kubernetes:6443
+
+contexts:
+- name: prod
+  context:
+    cluster: prod
+    user: ocfstaffer
+
+current-context: prod
```
cg505: Not sure how this will work if staffers want to add other contexts in their local kubeconfig file and switch to them. We might be better off just telling people "run `kubectl config use-context prod` to access prod".

modules/ocf/templates/kubectl/kubectl.conf.erb
```diff
@@ -0,0 +1,20 @@
+users:
+- name: ocfstaffer
```
keur: is the "-er" necessary?

modules/ocf/templates/kubectl/kubectl.conf.erb
```diff
@@ -0,0 +1,20 @@
...
+  user:
+    exec:
+      command: "/opt/share/utils/sbin/gen-kubernetes-cert"
```
cg505: I think it's preferable to not specify the full path, in case the script moves or something. This would also let staffers patch the script and put the patched version in their `PATH` to test changes.

modules/ocf/templates/kubectl/kubectl.conf.erb
```diff
@@ -0,0 +1,20 @@
...
+      apiVersion: "client.authentication.k8s.io/v1beta1"
+
+clusters:
+- name: prod
+  cluster:
+    certificate-authority-data: <%= @cluster_cert_base64.gsub(/\n/, "") %>
```
keur: ruby has a `delete` function you can use for this. imo more readable than gsub.

modules/ocf/manifests/packages/kubectl.pp
```diff
@@ -0,0 +1,12 @@
+class ocf::packages::kubectl {
+  include ocf::packages::kubernetes
+
+  package { 'kubectl':; }
+
+  $cluster_cert_base64 = base64('encode', lookup('kubernetes::kubernetes_ca_crt'))
```
keur: nit: newline

modules/ocf/templates/kubectl/kubectl.conf.erb
keur: staffers can also just change the `KUBECONFIG` variable to test changes. the likelihood of this path changing is pretty low

modules/ocf/templates/kubectl/kubectl.conf.erb
keur: I'd rather not have to opt-in to use default prod settings. Why can't staffers point to their own customized kubeconfig

cg505:
Do we want this on tsunami? Should we consider putting it in `ocf_admin` instead?

modules/ocf/templates/kubectl/kubectl.conf.erb
dkess: What other contexts would we want to use? I couldn't think of any, so I figured a system-wide config was OK.

That being said, if this comes up in the future (for example, if we want a separate context for cluster admin privileges), we can [merge configuration files](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/) between a system-wide and user-specific config. In that case, the user config would override the `current-context` given here.

modules/ocf/templates/kubectl/kubectl.conf.erb
dkess: Actually I'll change it to kerberos-login since that better describes what it's doing anyways

modules/ocf/templates/kubectl/kubectl.conf.erb
dkess: coolio

dkess:
I want this to be in `extrapackages` so it can also work on desktops, I think it's OK for it to be on tsunami, it won't work rn anyways but it'll come in handy if we ever want to provide Kubernetes as a service to our users.

PR #540, opened 2019-03-05
kubernetes: add haproxy redirects for ocf.io and local domain shortcuts

dkess:
This gives us the same behavior we've expected from Marathon services, where we can type something like "labmap.ocf.io" and have it redirect to "labmap.ocf.berkeley.edu". It also works for local shortcuts, so on the desktops you can go to "labmap/" in the browser and it'll properly redirect.

There are some caveats here that make it not quite up to parity with how it's handled in Marathon:

* This adds a redirect for `https://*.dev-kubernetes.ocf.berkeley.edu`, which I should probably get rid of, not sure how though. Alternatively, we could figure out how to properly handle this case.
* For services with even more shortcuts (for example, `sg` should redirect to `sourcegraph` and `phpmyadmin` should redirect to `pma`), this won't have the expected behavior. Again, I'm not sure what the best way would be to handle this case.

### DIFF
--- a/modules/ocf_kubernetes/manifests/master/loadbalancer.pp
+++ b/modules/ocf_kubernetes/manifests/master/loadbalancer.pp
@@ -34,17 +34,20 @@ class ocf_kubernetes::master::loadbalancer {
   service { 'keepalived': }
 
   $vip = 'lb-kubernetes'
+
+  class { 'ocf_kubernetes::master::loadbalancer::ssl':
+    vip => $vip,
+  }
+
+  # Redirect from "cname.ocf.io" to "cname.ocf.berkeley.edu" for each cname.
   $cnames = ldap_attr($vip, 'dnsCname', true)
   $alias_redirects = flatten($cnames.map |$domain| {
+    # When accessing domains from within the OCF subnet, going to the hostname
+    # (e.g. typing "labmap/" into the web browser) should also redirect.
     prefix(['.ocf.io', ''], $domain).map |$fqdn| {
       {'redirect' => "prefix https://${domain}.ocf.berkeley.edu code 301 if { hdr(host) -i ${fqdn} }"}
     }
   })
-
-
-  class { 'ocf_kubernetes::master::loadbalancer::ssl':
-    vip => $vip,
-  } ->
 
   haproxy::frontend { 'kubernetes-frontend':
     mode    => 'http',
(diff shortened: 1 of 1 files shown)


modules/ocf_kubernetes/manifests/master/loadbalancer.pp
```diff
@@ -33,8 +33,17 @@
...
 
+  $vip = 'lb-kubernetes'
+  $cnames = ldap_attr($vip, 'dnsCname', true)
+  $alias_redirects = flatten($cnames.map |$domain| {
+    prefix(['.ocf.io', ''], $domain).map |$fqdn| {
+      {'redirect' => "prefix https://${domain}.ocf.berkeley.edu code 301 if { hdr(host) -i ${fqdn} }"}
+    }
+  })
+
+
```
cg505: nit: extra newline

modules/ocf_kubernetes/manifests/master/loadbalancer.pp
```diff
@@ -43,10 +52,11 @@
       '0.0.0.0:80'  => [],
       '0.0.0.0:443' => ['ssl', 'crt', $haproxy_ssl],
     },
-    options => {
-      'default_backend' => 'kubernetes-backend',
-      'redirect'        => 'scheme https code 301 if !{ ssl_fc }',
-    };
+    options => [
+      {'default_backend' => 'kubernetes-backend'},
+    ] + $alias_redirects + [
```
keur: can we clean up this formatting? not readable at all

modules/ocf_kubernetes/manifests/master/loadbalancer.pp
```diff
@@ -33,8 +33,17 @@
   } ~>
   service { 'keepalived': }
 
+  $vip = 'lb-kubernetes'
+  $cnames = ldap_attr($vip, 'dnsCname', true)
+  $alias_redirects = flatten($cnames.map |$domain| {
+    prefix(['.ocf.io', ''], $domain).map |$fqdn| {
```
keur: i would add a comment about why the empty prefix is there, as it is not immediately obvious. A comment in general about what this is doing would be good.

keur: declare `vip` here and move the following block to be directly above the `haproxy::frontend` where it is actually used

dkess: I wanted to move this to be closer, but I was worried that it would break the arrow relationshp between `ocf_kubernetes::master::loadbalancer::ssl` and the `haproxy::frontend`. Not sure how this can be done within Puppet syntax.

keur: It is done with `require`

fydai:
Marathon uses a 307 redirect, should kubernetes use a 307 too? Otherwise looks good!

dkess:
Thanks everyone for the feedback, I've fixed the style to address these concerns.

@fydai where are you seeing 307 redirects? When test in Firefox it looks like 301s to me.

@cg505 yeah, once this is merged we shouldn't need to handle this within applications.

PR #541, opened 2019-03-05
Migrate jaws from marathon slave to kubernetes worker

dkess:
This is tested and working in prod! Mostly just making a PR for this in case there's anything I missed in this transition. I had to manually stop `mesos-slave` on jaws.


dkess:
The point of this is to facilitate the RT migration. RT is pretty resource-hungry, so we need to migrate workers before that, not after. Also, the mesos slaves seem pretty underutilized now (see https://mesos.ocf.berkeley.edu/#/agents) so I don't think this is a huge deal.

PR #542, opened 2019-03-06
Add kanboard shortcut shorturls

dkess:


keur:
I would recommend using the new api using `/s/` ;)

PR #546, opened 2019-03-09
Allow internal traffic from staffvms

fydai:
I want to run kubectl on my staffvm, and this allows me to do that.


fydai:
The following command seems to be a workaround, closing.
alias kubectl='ssh supernova KUBECONFIG=/etc/kubectl.conf kinit --keytab=keytab_file fydai/admin@OCF.BERKELEY.EDU kubectl'

PR #549, opened 2019-03-11
Only apply 3 monitor fix to eruption

MrMinos:
Edit: We have another three monitor setup at the lab that does not have middle monitor in a vertical setup. Would be nice to change the script to only apply on eruption which currently has vertical middle monitor setup.

### DIFF
--- a/modules/ocf_desktop/files/xsession/fix-displays
+++ b/modules/ocf_desktop/files/xsession/fix-displays
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 # Tile multiple monitors side-by-side
 import re
+import socket
 import subprocess
 import time
 
@@ -9,15 +10,11 @@ def get_monitors():
     lines = subprocess.check_output(['xrandr']).decode('utf8').splitlines()
     return [line.split()[0] for line in lines if ' connected' in line]
 
-def get_hostname():
-    lines = subprocess.check_output(['hostname']).decode('utf8').splitlines()
-    hostname = lines[0] if lines else None
-    return hostname
 
 if __name__ == '__main__':
     time.sleep(1)
     monitors = get_monitors()
-    hostname = get_hostname()
+    hostname = socket.gethostname()
 
     # tile monitors side-by-side
     prev = None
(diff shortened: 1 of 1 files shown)


dkess:
Why is this necessary?

abizer:
```python
import socket
hostname = socket.gethostname()
```

MrMinos:
Alright tested on hurricane, which currently is on my environment, that this works.

modules/ocf_desktop/files/xsession/fix-displays
```diff
@@ -22,8 +24,8 @@ if __name__ == '__main__':
         prev = monitor
 
     # special case: 3 monitors
```
kpengboy: Don't forget to update comment

MrMinos: Updated, thanks kpengboy!

MrMinos:
Added in the commit `ed44001` to ensure the monitor ordering at OS level to follow the physical port ordering at the graphics card.

This commit may break the current double monitor setup if they were not setup by physical port ordering at the graphics card. We will have to ensure the port ordering before rolling out to master branch.

```
    # For example, with the below grahpics card, excluing DVI port
    # Left most monitor should go to the left most port (DP 1) and so on
    #   
    # +------------------------------------------------+
    # |   +----------------+    | +------------------+ |
    # |   |    D V I       |    | |  |  Exhaust   |  | |
    # |   +----------------+    | +------------------+ |
    # +------------------------------------------------+
    # |   +-------+ +--------+  +-------+   +-------+  |
    # |   +  DP 1 | |  HDMI  |  +  DP 2 |   +  DP 3 |  |
    # |    \------+ \-------/    \------+    \------+  |
    # +------------------------------------------------+
```

hieradata/type/desktop.yaml
```diff
@@ -3,6 +3,7 @@ classes:
 
 ocf::auth::glogin: [[ocf, LOCAL]]
 ocf::auth::gsudo: [ocfstaff]
+ocf::auth::usudo: [minos]
```
cg505: plz don't forget to remove this

MrMinos:
Ah you caught me, I was trying to sneak into having root xD

minos-cs184:
Today, after hours; we can merge this into master and if some of the hosts are not correctly oriented, I can go over those with my family members and show how fix-display works.

PR #559, opened 2019-03-25
do not backport nviida drivers

cg505:
Since the versions in stretch are newer than the versions in stretch-backports, this was causing problems on up-to-date systems.


ethanhs:
>the versions in stretch are newer than the versions in stretch-backports

wat

redplanks:
It looks like [stretch-backports just upgraded](https://packages.debian.org/stretch-backports/nvidia-driver); they were the same version last night when I checked.

cg505:
lmao
The question remains, should we backport?

redplanks:
The backported version is newer right now, so I'm in favor of keeping backports

cg505:
sgtm

PR #561, opened 2019-03-26
Mirroring openbsd and freebsd

MCer4294967296:
The title pretty much says everything.

Currently the config is to sync openbsd from `usa.openbsd.org` which is located in Colorado (not very sure if that's a good idea,) and freebsd from MIT mainly because I cannot find a rsync host in the United State on the official list (`www.freebsd.org/doc/handbook/mirrors-ftp.html`) and MIT happens to host freebsd.

(Also, have we considered a template for sync-archive as well?)

### DIFF
--- a/modules/ocf_mirrors/manifests/projects/freebsd.pp
+++ b/modules/ocf_mirrors/manifests/projects/freebsd.pp
@@ -11,7 +11,7 @@ class ocf_mirrors::projects::freebsd {
 
   ocf_mirrors::monitoring { 'freebsd':
     type          => 'unix_timestamp',
-    upstream_host => 'mirror.math.princeton.edu',
+    upstream_host => 'ftp.freebsd.org',
     upstream_path => '/pub/FreeBSD',
     ts_path       => 'TIMESTAMP';
   }
--- a/modules/ocf_mirrors/manifests/projects/freebsd.pp
+++ b/modules/ocf_mirrors/manifests/projects/freebsd.pp
@@ -11,7 +11,8 @@ class ocf_mirrors::projects::freebsd {
 
   ocf_mirrors::monitoring { 'freebsd':
     type          => 'unix_timestamp',
-    upstream_host => 'mirrors.mit.edu',
+    upstream_host => 'mirror.math.princeton.edu',
+    upstream_path => '/pub/FreeBSD',
     ts_path       => 'TIMESTAMP';
   }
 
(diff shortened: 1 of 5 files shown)


MCer4294967296:
True, lemme seek another repo

MCer4294967296:
and it seems like.. MIT's timestamp is 2013-11-04 despite its last modified time being 2019-Mar-26..? `mirror.mit.edu/FreeBSD/TIMESTAMP`

MCer4294967296:
Now we are syncing from rit and monitoring from Princeton :3

dkess:
Also, I assume you'll need to contact the maintainers of these distros to ensure we're listed as official mirrors?

MCer4294967296:
Yes, fallingrocks is currently already hosting those two (under my puppet environment).

And I don't really understand how they handle mirrors like us: on their mirror list, none of MIT, Princeton, or RIT is present. So...

modules/ocf_mirrors/manifests/projects/freebsd.pp
```diff
@@ -0,0 +1,25 @@
...
+      source  => 'puppet:///modules/ocf_mirrors/project/freebsd',
+      owner   => mirrors,
+      group   => mirrors,
+      mode    => '0755',
+      recurse => true;
+  }
+
+  ocf_mirrors::monitoring { 'freebsd':
+    type          => 'unix_timestamp',
+    upstream_host => 'mirror.math.princeton.edu',
```
jvperrin: I think we should really be using the official upstream for this (`http://ftp.freebsd.org/pub/FreeBSD/TIMESTAMP`), instead of another mirror that could be out of date.

modules/ocf_mirrors/manifests/projects/openbsd.pp
```diff
@@ -0,0 +1,25 @@
...
+      source  => 'puppet:///modules/ocf_mirrors/project/openbsd',
+      owner   => mirrors,
+      group   => mirrors,
+      mode    => '0755',
+      recurse => true;
+  }
+
+  ocf_mirrors::monitoring { 'openbsd':
+    type          => 'unix_timestamp',
+    upstream_host => 'mirrors.mit.edu',
```
jvperrin: Likewise here, I think this should be `https://ftp.openbsd.org/pub/OpenBSD/timestamp` instead so we're checking the ultimate source of packages, not another mirror that could be outdated.

PR #566, opened 2019-04-01
Revert "Remove chronic from lets-encrypt-update temporarily, run dail…

dkess:
…y (#491)"

This reverts commit 172763aeee65977be20eff4fc70a2101084314a7.

This hasn't been giving us problems so I assume we don't need it anymore, unless I'm missing something?


dkess:
Note that if this does become a problem again, instead of sending it to rootspam, we should put it on a systemd timer instead (like mirrors), that way we can still see the logs.

PR #567, opened 2019-04-02
Pin kubernetes packages in apt.

redplanks:


dkess:
Oh also you should have a comment that explains exactly what's going on here

redplanks:
Within the template, or in this PR?

dkess:
in the puppet source just say "this holds the Kubernetes version because we want to upgrade it manually"

modules/ocf/templates/apt/kubernetes-pin.pref.erb
```diff
@@ -0,0 +1,9 @@
+# This file is managed by Puppet. DO NOT EDIT.
+# Pins each package in $kube_packages to the hiera-specified version,
+# so it won't get upgraded automatically by apt.
+<% @kube_packages.each do |pkg| -%>
+Package: <%= pkg %>
+Pin: version <%= scope().call_function('hiera', ['kubernetes::kubernetes_package_version']) %>
```
dkess: Instead of grabbing from hiera like this, I think you should do it like in https://github.com/ocf/puppet/blob/master/modules/ocf_kubernetes/manifests/master.pp#L38 and https://github.com/ocf/puppet/blob/master/modules/ocf_kubernetes/templates/static-tokens.csv.erb

PR #569, opened 2019-04-05
kubernetes lb: add extra alias redirects

dkess:
The Puppet code for this is kind of ugly, but it works. This lets us have aliases for websites, like `sg.ocf.berkeley.edu` redirecting to `sourcegraph.ocf.berkeley.edu`.

### DIFF
--- a/modules/ocf_kubernetes/manifests/master/loadbalancer.pp
+++ b/modules/ocf_kubernetes/manifests/master/loadbalancer.pp
@@ -1,6 +1,12 @@
 class ocf_kubernetes::master::loadbalancer {
   include ocf::firewall::allow_web
 
+  $extra_aliases = {
+    'pma'      => ['phpmyadmin'],
+    'metabase' => ['mb'],
+    'sourcegraph' => ['sg'],
+  }
+
   $kubernetes_worker_nodes = lookup('kubernetes::worker_nodes')
 
   # At any given time, only one kubernetes master will hold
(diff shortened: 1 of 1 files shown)


modules/ocf_kubernetes/manifests/master/loadbalancer.pp
```diff
@@ -114,16 +120,32 @@
...
+    }
+
+    # Append .ocf.berkeley.edu, .ocf.io, and empty string for each of the extra aliases
+    $extra_fqdns = flatten($extra_names.map |$name| {
+      prefix(['.ocf.berkeley.edu', '.ocf.io', ''], $name)
+    })
+
+    $redir_fqdns = $main_fqdns + $extra_fqdns
+
+    notify {$redir_fqdns:}
```
jvperrin: Remove this since it's just for debugging

modules/ocf_kubernetes/manifests/master/loadbalancer.pp
```diff
@@ -114,16 +120,32 @@
...
+
+    # Get the list of aliases for this name, or an empty list if there are none
+    $extra_names = flatten([$extra_aliases[$domain]]).filter |$val| {
+      $val =~ NotUndef
+    }
+
+    # Append .ocf.berkeley.edu, .ocf.io, and empty string for each of the extra aliases
+    $extra_fqdns = flatten($extra_names.map |$name| {
+      prefix(['.ocf.berkeley.edu', '.ocf.io', ''], $name)
+    })
```
jvperrin: Why not just hardcode these (and the `$extra_aliases` above) in an array or something instead? As this is, it's really hard to figure out what is going on, and hardcoding just a few domains would make it much more explicit and clear at the small cost of a bit of repetition.

keur:
@jvperrin I agree. But what about using something like PuppetDB instead of hardcoding the values into Puppet?

dkess:
Gonna close this and instead make another PR where all names are hardcoded.

PR #572, opened 2019-04-06
fix-displays daisy chaining support

dkess:
This lets us daisy-chain displays together, which is nice because it makes it easier for staffers to connect lab monitors to their personal machines. We need to do a strange thing where we turn of the monitors (`xset dpms force off`) and turn them back on for this to work, otherwise chained monitors won't turn on.

This PR also includes a commit to disable the triple-monitor special case on eruption, which isn't needed anymore.

### DIFF
--- a/modules/ocf_desktop/files/xsession/fix-displays
+++ b/modules/ocf_desktop/files/xsession/fix-displays
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 # Tile multiple monitors side-by-side
-import re
 import socket
 import subprocess
 import time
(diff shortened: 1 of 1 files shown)


modules/ocf_desktop/files/xsession/fix-displays
```diff
@@ -30,8 +30,19 @@ if __name__ == '__main__':
     # |    \------+ \-------/    \------+    \------+  |
     # +------------------------------------------------+
 
+    is_daisy_chaining = any(monitor.endswith('.8') for monitor in monitors)
+    if is_daisy_chaining:
+        sorted_monitors = list(sorted(monitors, reverse=True))
+    else:
+        sorted_monitors = list(sorted(monitors))
```
cg505: why tho

dkess: The first monitor in the daisy chain is always the leftmost one, but it also will sort itself as the last one. So we need to reverse it.

cg505: could you add a comment?

MrMinos:
The code is tested on hurricane(triple monitor) and cyclone (double monitor). Lgtm.

dkess:
Before merging this, we need to fix another issue where, it the display goes to sleep (for example when idling), when it wakes back up, only the first monitor will turn back on. We need to somehow manually trigger the other screens to turn back on as well.

MrMinos:
`/usr/lib/pm-utils/sleep.d` scripts in this folder gets run during sleep/wakeup cycle

[0]:https://askubuntu.com/questions/226278/run-script-on-wakeup

dkess:
Closing in favor of #588

PR #575, opened 2019-04-08
mail: restrict staff@ and hpc-users@ to staff

cg505:
this might be possible to simplify, i don't know much about postfix

### DIFF
--- a/modules/ocf_mail/files/site_ocf/aliases
+++ b/modules/ocf_mail/files/site_ocf/aliases
@@ -32,53 +32,53 @@ site-manager: sm
 owner-sm: officers
 
 staff: staff-current
-wheel: officers
-owner-staff: wheel
+wheel: staff-current
+owner-staff: root
 
 # request tracker
 bod: "|/usr/bin/rt-mailgate --queue 'bod' --action correspond --url https://rt.ocf.berkeley.edu"
 bod-comment: "|/usr/bin/rt-mailgate --queue 'bod' --action comment --url https://rt.ocf.berkeley.edu"
-owner-bod: wheel
+owner-bod: root
 
 bureaucracy: "|/usr/bin/rt-mailgate --queue 'bureaucracy' --action correspond --url https://rt.ocf.berkeley.edu"
 bureaucracy-comment: "|/usr/bin/rt-mailgate --queue 'bureaucracy' --action comment --url https://rt.ocf.berkeley.edu"
-owner-bureaucracy: wheel
+owner-bureaucracy: root
 
 devnull: "|/usr/bin/rt-mailgate --queue 'devnull' --action correspond --url https://rt.ocf.berkeley.edu"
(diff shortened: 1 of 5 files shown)


modules/ocf_mail/files/site_ocf/aliases
```diff
@@ -32,7 +32,7 @@ site-manager: sm
 owner-sm: officers
 
 staff: staff-current
-wheel: staff-current
+wheel: officers
```
dkess: Why did you change this?

cg505: It doesn't make sense that owner-*@ all goes to staff@ in my opinion, especially as staff grows. Also, it would mean that all of those addresses would need to be specified in `protected-destinations`.

cg505: http://www.postfix.org/aliases.5.html ctrl-f owner-name

we might want to change this to root actually

kpengboy: Traditionally, we have used `wheel` as a mailing list for technical discussion among staff. Unlike root@, it was not polluted with alerts and noisy messages, and unlike officers@, it was specifically for technical discussion.

We deliberately put all staff (as opposed to, say, root staff only) on wheel@ so that none of them would be left out of technical discussion. (This was also a time when staff was a lot smaller.)

Who do you think should be on this list today? Or is this list obsolete?

cg505: I think it's fair to say that technical discussion general does not happen on wheel@ at this point. I didn't even know this list existed. Since staff@ is so large at this point, I don't think it makes sense to send delivery failures etc to all staff. For this I think root@ or sm@ is more reasonable.

dkess: In that case, you should change the `owner-*` lists to point somewhere else (like root), since we do mention [in the docs](https://www.ocf.berkeley.edu/docs/staff/mailing-lists/) that `wheel@` is a technical discussion list for staff. This should be in a separate PR since it's out-of-scope for this fix.

cg505: Okay. The separate PR should be merged before this one, since I haven't protected wheel@ in this PR.

cg505: Alternatively: we keep wheel and protect it as well, and change to owner-* lists to go to root. what do you think?

dkess: That sgtm

abizer:
I think hpc-users was supposed to be an announce-only mailing list, rather than a discussion one

cg505:
That was my impression as well. We don't want to spam that list. If desired, we could let people on the list send as well.

cg505:
Going to merge this unless anyone opposes.

PR #577, opened 2019-04-10
Add /sdocs shorturl

squashcat:
Please approve my new hacking skills


squashcat:
> if adding this it would be ideal to add an optional capture group at the end, so that urls like "ocf.io/sdocs/procedures/virtual-hosting/" and such might work, rather than strictly terminating it.

it is done @abizer

PR #578, opened 2019-04-12
Switch ubuntu mirror upstream from MIT to Enzu.

redplanks:
rsync with MIT's mirror kept failing, so I switched to this one temporarily, and it works fine. It's listed as a 20GBPS mirror too [here](https://launchpad.net/ubuntu/+archivemirrors).


ethanhs:
I hear Google's mirrors are nice and speedy.



<sup><sup>jk</sup></sup>

PR #583, opened 2019-04-15
Change flags in sync script of Parrot

MCer4294967296:
Parrot have been failing for a while, the reason was something like "cannot open directory `.~tmp~`". After adapting some flags from other projects, the problem seems solved. This also suggest we can consider using a fixed set of flags for all sync scripts (except those `--exclude`).

By the way I have no idea what Parrot is about.

### DIFF
--- a/modules/ocf_mirrors/files/project/parrot/sync
+++ b/modules/ocf_mirrors/files/project/parrot/sync
@@ -1,3 +1,3 @@
 #!/bin/sh -eu
-/usr/local/bin/rsync-no-vanished -rltpHS --delete-excluded \
+/usr/local/bin/rsync-no-vanished -rltpHS --delete-after --delay-updates\
 	rsync://archive.parrotsec.org/parrot/ /opt/mirrors/ftp/parrot


dkess:
Can you link where you found this fix from?

MCer4294967296:
modules/ocf_mirrors/files/project/archlinux/sync-archive
(I have no idea if this link will work)

MCer4294967296:
fallingrocks is currently on production. I'm willing to investigate what actually is happening.

MCer4294967296:
I think I know what happened - I would say it's an error of our upstream.
So the error message says `rsync: opendir "/iso/testing/.~tmp~" (in parrot) failed: Permission denied (13)`. Instead of me thinking it's failing to open the local dir, it's actually a dir on the upstream that rsync is trying to access, and that dir has permission 700.
As for why the patch eliminates the error, it's because of the `--delay-updates` flag. rsync man page says this flag will create as default a dir called `.~tmp~`, and the creation of this dir as the temp folder will equivalently add an `--exclude-dir` flag, at which point this dir on the upstream will no longer be traversed and thus no errors any more.

Also, that iso/testing/ directory on the upstream contains nothing other than the strange temp folder.

kpengboy:
Is this still an issue?

MCer4294967296:
Since it's not merged, I guess the systemd service is still failing. But merging this would theoretically solve the failure.

So is this still an issue? idk xD

PR #586, opened 2019-04-20
add rsync logs to mirrors stats collection

abizer:
This adds support for rsync logs to the mirrors stats collection script.

I apologize for writing this absolutely garbage code in 2017


modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
...
+                print('Error on line {}:\\  {}\\'.format(n, line))
+                traceback.print_exc()
+
+            if line_date.date() != log_date:
+                continue
+
+            # extract dist name from request url
+            # '/debian/pool/main/h/hwdata/...' -> 'debian'
+            dist = stats[6]
+            dist = dist.split('/')[1] if '/' in dist else dist
```
keur: Since you already pull in path, you should create a `Path` object with `dist` and use `parts` on it. Right now this is kinda messy.

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
...
+            except ValueError:
+                print('Error on line {}:\\  {}\\'.format(n, line))
+                traceback.print_exc()
+
+            if line_date.date() != log_date:
+                continue
+
+            # extract dist name from request url
+            # '/debian/pool/main/h/hwdata/...' -> 'debian'
+            dist = stats[6]
```
keur: I would put an entire line of the logfile in the comment so it's apparent what is actually being parsed.

modules/ocf_mirrors/files/collect-mirrors-stats
keur: When will it be the case that there is no `/` in the URI? Also, do we want to ignore the request if it's not a `GET`?

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
...
+            dist = dist.split('/')[1] if '/' in dist else dist
+
+            # record if we returned http 2xx/3xx
+            if stats[8][0] in ('2', '3') and dist in projects:
```
keur: Hmm... This is pretty ugly. What about at the top of this method you give each of the columns that you use in stats a more descriptive variable name.

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
...
+
+    This function is project-agnostic.
+    """
+
+    tx = 0
+    rx = 0
+    with filename.open('r') as f:
+        for line in f.readlines():
+            # log format is <datetime> <pid> sent <tx> bytes received <rx> bytes total size
+            if line.find('sent') > 0:
```
keur: hopefully we never mirror a project called sent

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
...
+                        help='rsync log files(s) to process')
+    parser.add_argument('date', nargs='?', default=date.today() - timedelta(1),
+                        help='date for use in filtering log entries')
+
+    args = parser.parse_args()
+
+    # first, figure out what projects we're getting stats for
+    sources = []
+    for project in MIRRORS_DATA_PATH.iterdir():
+        if project.is_dir() and not project.name.startswith('.'):
```
keur: err... `.` and `..` aren't going to be in the output of iterdir

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
...
+        apache_log_files = args.apache
+    else:
+        apache_log_files = []
+        for log_name in APACHE_LOG_FILES:
+            # logrotate rotates the log at 6am, but this script
+            # runs at midnight. So to capture 24h of data, we need
+            # to parse the rotated log as well. (the .1 file)
+            log = APACHE_LOG_PATH / log_name
+            apache_log_files.append(log)
+            apache_log_files.append(Path(str(log) + '.1'))
```
keur: `Path` objects have a `with_suffix` method you should use here.

modules/ocf_mirrors/files/rsyncd.conf
```diff
@@ -12,113 +12,142 @@ dont compress = *.gz *.tgz *.zip *.z *.rpm *.deb *.iso *.bz2 *.tbz
 [apache]
 	comment = apache project mirror
 	path = /opt/mirrors/ftp/apache
+	log file = /var/log/rsync/apache.log
```
keur: this file is template-able

modules/ocf_mirrors/files/collect-mirrors-stats
abizer: the only requests in the logs are HEAD and GET, and HEAD is just a file listing to I didn't think it was valuable to include, but we may as well

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
...
+APACHE_LOG_PATH = Path('/var/log/apache2')
+APACHE_LOG_FILES = [
+    Path('mirrors.ocf.berkeley.edu_access.log'),
+    Path('mirrors.ocf.berkeley.edu-ssl_access_ssl.log')
+]
+RSYNC_LOG_PATH = Path('/var/log/rsync')
+
+
+def log_to_mysql(projects, dt=None, quiet=False):
+    log_date = dt or date.today()
```
ethanhs: Nit: maybe just `dt=date.today()`?

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/python3
+import argparse
+import os
+import traceback
+from datetime import date
+from datetime import datetime
+from datetime import timedelta
+from pathlib import Path
```
ethanhs: +1 for using pathlib :)

modules/ocf_mirrors/files/collect-mirrors-stats
abizer: mm yeah holdover from the previous script

modules/ocf_mirrors/files/collect-mirrors-stats
abizer: that is very nice

modules/ocf_mirrors/files/rsyncd.conf
abizer: likely, maybe we could do some `concat` stuff here

modules/ocf_mirrors/files/collect-mirrors-stats
ethanhs: I would recommend writing an regex to match the log format. See e.g. https://stackoverflow.com/a/12544587. You could even do named captures (`(?P<date>...)`) so that you can reference them like `match['date']`. Though maybe that is overly complicated.

cg505: I'd prefer something like
```python
status_code = int(stats[STATUS_COL])
if status_code >= 200 and status_code < 400 and dist in projects:
    ...
```
This way, if the thing we think is the status code is in fact not the status code, the int conversion will fail instead of just thinking that it was a 400 or something.

Also, why do we increment "other" when we don't get 2xx or 3xx? Seems like we should just ignore the traffic completely.

modules/ocf_mirrors/files/collect-mirrors-stats
cg505: Like @ethanhs said above, I think a regex would also be cleaner here.

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
...
+            apache_log_files.append(log)
+            apache_log_files.append(Path(str(log) + '.1'))
+
+    # and the same for rsync
+    if args.rsync:
+        rsync_log_files = args.rsync
+    else:
+        rsync_log_files = []
+        for log_file in RSYNC_LOG_PATH.iterdir():
+            # log files are named <project>.log
```
cg505: Why do we have this comment? Did you mean to add an `if log_file.ends_with('.log')`?

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
...
+        projects[project] = {'up': 0, 'down': 0}
+
+    # pull the data itself from apache
+    for log_filename in apache_log_files:
+        projects = process_apache_log(projects, log_filename, args.date)
+
+    # ...and rsync
+    for log_filename in rsync_log_files:
+        tx, rx = process_rsync_log(log_filename)
+        projects[log_filename.stem]['up'] += rx
```
cg505: Put the `# log files are named <project>.log` comment here imo

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
...
+            rsync_log_files.append(log_file)
+
+    # set up data structures
+    projects = {}
+    for project in sources:
+        projects[project] = {'up': 0, 'down': 0}
+
+    # pull the data itself from apache
+    for log_filename in apache_log_files:
+        projects = process_apache_log(projects, log_filename, args.date)
```
cg505: Construct `projects` in `process_apache_log` (instead of above) and then just pass `sources` as a parameter to this function. That way we don't do this awkward thing where we pass in `projects` as a parameter, it's mutated, and then we assign it back to itself. (The assignment is currently a no-op.)

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
...
+    # ...and rsync
+    for log_filename in rsync_log_files:
+        tx, rx = process_rsync_log(log_filename)
+        projects[log_filename.stem]['up'] += rx
+        projects[log_filename.stem]['down'] += tx
+
+    # print out some stats!
+    if not args.quiet:
+        for project in projects:
+            print('{:20} {:10} {:10}'.format(
```
cg505: Maybe we should dynamically calculate the field widths? Probably not worth it though.

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/python3
```
dkess: Can you add a comment at the top explaining what this script is supposed to do?

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -0,0 +1,168 @@
...
+            # log format is <datetime> <pid> sent <tx> bytes received <rx> bytes total size
+            if line.find('sent') > 0:
+                stats = line.split()
+                tx += int(stats[4])
+                rx += int(stats[7])
+
+    return tx, rx
+
+
+if __name__ == '__main__':
```
dkess: Use a main function

modules/ocf_mirrors/files/collect-mirrors-stats
dkess: Or just make it a required param, since it's always provided as an argument in this script?

modules/ocf_mirrors/files/collect-mirrors-stats
abizer: other contains all kinds of 'interesting' data, like directory traversal attempts, people trying to break the server with strange unicode garbage, it's actually a decent amount of data, usually a couple hundred MB a day or something like that.

modules/ocf_mirrors/files/collect-mirrors-stats
abizer: I think that breaks the desired abstraction. we shouldn't construct `projects` from inside the function, because either we'll be: (a) overloading the function from 'just' processing a log file to having to know about how to construct the projects dictionary or (b) we won't know what a valid project is or isn't given just the log file itself. iteratively adding to the dictionary isn't ideal, but it's reminiscent of an iterative ETL approach and allows us to avoid having to do some kind of dictionary merging outside of the function.

modules/ocf_mirrors/files/collect-mirrors-stats
abizer: this runs in quiet mode in prod, the print display is just for debugging

modules/ocf_mirrors/files/collect-mirrors-stats
abizer: no, it's just a comment so the reader knows what the contents of the list look like

modules/ocf_mirrors/files/collect-mirrors-stats
abizer: the 'nice' approach leads to a lot of breakage. the `line[8][0]` approach has the benefit of being so simple that when the log line in question isn't valid, it simply dumps it into `other`, whereas trying to parse it as as int in many cases leads to a `ValueError` when there is some unicode garbage in the log line.

modules/ocf_mirrors/files/collect-mirrors-stats
abizer: so Pathlib is nice and clean, but adds a lot of overhead according to the profiler. 43 seconds of the 2 minute runtime of the script is inside of Pathlib, now. Not that it matters for a script that runs once a day.

PR #588, opened 2019-04-24
Daisy chaining support for multi-monitor setup

MrMinos:
Continuing from puppet/PR #572, this PR will allow for a reliable daisy chained, multi-monitor setup.

As dkess mentioned in the reference PR
>This PR also includes a commit to disable the triple-monitor special case on eruption, which isn't needed anymore.

### DIFF
--- a/modules/ocf_desktop/files/suspend/fix-daisy
+++ b/modules/ocf_desktop/files/suspend/fix-daisy
@@ -1,7 +1,4 @@
 #!/bin/bash
 
-case "$1" in
-    resume)
-        xset dpms force off
-        xset dpms force on
-esac
+xset dpms force off
+xset dpms force on
--- a/modules/ocf_desktop/files/suspend/fix-daisy
+++ b/modules/ocf_desktop/files/suspend/fix-daisy
@@ -1,4 +1,10 @@
 #!/bin/bash
 
+# when using daisy chaining on nvidia graphics card on debian,
+# primary monitor is on but the rest of the daisy chained
+# monitors are asleep, and does not wake up.
+#
+# below code forces the monitors to wake up
+
 xset dpms force off
 xset dpms force on
--- a/modules/ocf_desktop/files/suspend/fix-daisy
+++ b/modules/ocf_desktop/files/suspend/fix-daisy
@@ -1,4 +1,7 @@
 #!/bin/bash
 
-xset dpms force off
-xset dpms force on
+case "$1" in
+    resume)
+        xset dpms force off
+        xset dpms force on
+esac
--- a/modules/ocf_desktop/files/suspend/fix-daisy
+++ b/modules/ocf_desktop/files/suspend/fix-daisy
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+xset dpms force off
+xset dpms force on
--- a/modules/ocf_desktop/files/suspend/fix-daisy
+++ b/modules/ocf_desktop/files/suspend/fix-daisy
@@ -7,4 +7,3 @@
 # below code forces the monitors to wake up
 
 xset dpms force off
-xset dpms force on
--- a/modules/ocf_desktop/files/suspend/fix-daisy
+++ b/modules/ocf_desktop/files/suspend/fix-daisy
@@ -7,3 +7,5 @@
 # below code forces the monitors to wake up
 
 xset dpms force off
+xset dpms force on
+xset dpms 0 0 0
--- a/modules/ocf_desktop/files/xsession/fix-displays
+++ b/modules/ocf_desktop/files/xsession/fix-displays
@@ -11,6 +11,9 @@ def get_monitors():
 
 
 def is_daisy_chaining(monitors):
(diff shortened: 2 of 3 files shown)


modules/ocf_desktop/manifests/suspend.pp
```diff
@@ -10,8 +10,19 @@
     '/etc/acpi/events/powerbtn-acpi-support':
       source  => 'puppet:///modules/ocf_desktop/suspend/powerbtn-acpi-support',
       require => Package['acpi-support-base'];
+
+    # place files to handle daisy chain sleep issue
+    '/usr/lib/pm-utils/sleep.d/999fix-daisy':
+      mode   => '0755',
+      source => 'puppet:///modules/ocf_desktop/suspend/fix-daisy';
+
+    # place files to handle daisy chain sleep issue
```
dkess: What is the daisy chain sleep issue? Explain it in a comment

modules/ocf_desktop/files/xsession/fix-displays
```diff
@@ -11,6 +10,20 @@ def get_monitors():
     return [line.split()[0] for line in lines if ' connected' in line]
 
 
+def is_daisy_chaining(monitors):
```
dkess: Explain what "daisy chaining" is in the docstring

modules/ocf_desktop/files/xsession/fix-displays
```diff
@@ -31,26 +44,20 @@ if __name__ == '__main__':
...
-        subprocess.check_call(['xrandr', '--output', monitors[1],
-                               '--rotate', 'left', '--pos', '{}x{}'.format(x_pos, middle_offset)])
-
-        x_pos = str(int(x_pos) + 1200)
-        subprocess.check_call(['xrandr', '--output', monitors[2],
-                               '--pos', '{}x{}'.format(x_pos, right_offset)])
+    if daisy_chained:
+        # Disable power saving, this is necessary for some reason
+        subprocess.check_call(['xset', 'dpms', 'force', 'off'])
+        subprocess.check_call(['xset', 'dpms', 'force', 'on'])
```
dkess: Perhaps we should call out to the new `fix-daisy` script here.

modules/ocf_desktop/manifests/suspend.pp
MrMinos: Done in the commit below.

modules/ocf_desktop/files/xsession/fix-displays
MrMinos: Yeah, that works!

modules/ocf_desktop/files/xsession/fix-displays
MrMinos: Yeet, included it.

MrMinos:
Tested the latest commit on `drought`, 'madcow', briefly on 'hurricane'

Let me know if there's any improvement that can be made.

cg505:
Definitely would be nice if we could do this without disable dpms, but if it works then shipit.
Please squash when you merge.

MrMinos:
> Definitely would be nice if we could do this without disable dpms, but if it works then shipit.
> Please squash when you merge.

We cannot do this without `xset dpms` unfortunately, because we still don't know why exactly the daisy-chained downstream monitors do not wake up without the command.

MrMinos:
One note, cyclone's monitors does not work with DisplayPort MST, for unknown reasons. Do not try.

modules/ocf_desktop/files/suspend/fix-daisy
```diff
@@ -0,0 +1,11 @@
...
+
+# when using daisy chaining on nvidia graphics card on debian,
+# primary monitor is on but the rest of the daisy chained
+# monitors are asleep, and does not wake up.
+#
+# below code forces the monitors to wake up
+
+xset dpms force off
+xset dpms force on
+xset dpms 0 0 0
```
dkess: I think this line is disabling all monitors from going to sleep. We shouldn't be doing this. I thought forcing off and then back on was sufficient to fix the problem?

MrMinos: Following up from in-person conversation:
- There was a reason the line `xset dpms 0 0 0` was necessary - mainly there was a case where the above two lines fixes it, but as soon as the displays go to sleep the same issue happnes again because the `fix-daisy` script does not get applied.
- Only apply this `fix-daisy` script to multiple monitor setup - since the vast majority of the desktops are single monitor setup, we won't be wasting as much energy than to disable the sleep everywhere.

cg505:
[this change broke my workflow](https://xkcd.com/1172/): now the lab gets hot at night because all the desktops are still awake

MrMinos:
> [this change broke my workflow](https://xkcd.com/1172/):

THAT'S HORRIFYING.

It would be good to see a comparison data on this. We only started measuring temp and CO2 after this was merged.

As mentioned above, we can apply `fix-daisy` only to daisy chained monitors. I will try to squeeze in a patch either tonight or sometime soon mid-week at the latest.

PR #593, opened 2019-04-28
kubernetes: add webui

dkess:
This commit adds files that enable the Kubernetes web dashboard, including:

* static bearer tokens that give ocfroot and ocfstaff privileges
* nginx configs to take PAM auth and pass these tokens to the dashboard
* Kubernetes resource definitions for installing the dashboard into the cluster.

We need to pass an "allow insecure login" flag to the dashboard container since it doesn't understand that it's actually being served over HTTPS. Unfortunately, this means we have to copy their whole service definition into version control (`webui.yaml`) to change that one value.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,6 +31,8 @@ repos:
     -   id: check-xml
         exclude: ^modules/ocf_desktop/files/pam/.*\.xml$
     -   id: check-yaml
+        args:
+        -   --allow-multiple-documents
     -   id: debug-statements
     -   id: double-quote-string-fixer
     -   id: detect-private-key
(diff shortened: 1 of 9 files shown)


Baisang:
Is this extensible to other services? E.g. if I want another kubernetes service to have a similar nginx/pam based authentication. If so, is there a way to make it extensible without significant puppet changes (allowing for faster deploys)?

cg505:
"hi we use X-Did-It-Come-From-HTTPS-Originally: true, can we add this"

dkess:
This kind of setup is possible for other services. We usually prefer to do it within the container itself (like what RT does). The Kubernetes web UI is a special case because it needs to run in the `kube-system` namespace and I'd prefer to not mess with their deployment strategy too much. Also, the web UI isn't actually trusting the username HTTP header-- instead, we pass a bearer token. So this setup is a bit unusual and unlikely to be replicated anyways.

I do agree it's a problem that we do OCF login pages in so many different ways. With the amount of login services we now have, it's definitely worth looking into an SSO solution like ory/hydra.

PR #594, opened 2019-04-30
only disable sleep on double monitors and up

MrMinos:
As referenced in #588, we don't need single monitor desktops to be kept awake to prevent DisplayPort MST display issue.

More refined solution would be to apply the `fix-daisy` script to only the daisy chained monitors.

### DIFF
--- a/modules/ocf_desktop/files/suspend/fix-daisy
+++ b/modules/ocf_desktop/files/suspend/fix-daisy
@@ -6,6 +6,11 @@
 #
 # below code forces the monitors to wake up
 
-xset dpms force off
-xset dpms force on
-xset dpms 0 0 0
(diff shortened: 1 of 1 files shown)


MrMinos:
Tested on `hurricane` and `surge`.

MrMinos:
> Maybe @cg505 can finally do work now

Indeed.

> thanks for doing this! most ideally, the script shouldn't keep desktops from sleeping at all, but that would definitely be more involved (if doable rn)

Yes, until we know the reason why nvidia DisplayPort MST (daisy chaining) does not turn on all the monitors, likely we will have to rely on a workaround. Currently this seems like the best option.

cg505:
ty @MrMinos

PR #596, opened 2019-05-01
Use /srv/prometheus as default textfile dir

dkess:
This consolidates us to a single folder, present on every host, that can be used for textfile metrics. Now, adding custom metrics to a host is as easy as writing to `/srv/prometheus`!

The first two commits in this PR move the existing textfile hosts (`mirrors` and `printhost`) to the new folder. The final commit solidifies it as the default.

### DIFF
--- a/modules/ocf_mirrors/manifests/init.pp
+++ b/modules/ocf_mirrors/manifests/init.pp
@@ -193,10 +193,6 @@ class ocf_mirrors {
   }
 
   package { ['python3-prometheus-client']: }
-  # Temporary-- to be moved to all hosts
-  file { '/srv/prometheus':
-    ensure =>directory;
-  } ->
   File<|title == '/srv/prometheus'|> {
     owner => 'mirrors',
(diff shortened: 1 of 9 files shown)


modules/ocf_mirrors/manifests/init.pp
```diff
@@ -193,10 +193,7 @@
   }
 
   package { ['python3-prometheus-client']: }
-  file {
-    '/opt/mirrors/prometheus':
-      ensure => directory,
-      owner  => 'mirrors',
-      group  => 'mirrors',
+  File<|title == '/srv/prometheus'|> {
+    owner =>  'mirrors',
```
cg505: what about group?

dkess:
Thanks for comments. I've added the `group` attribute. This was tested on printhost and mirrors yesterday. I'll merge this later today.

cg505:
"I'll merge this later today." *merges 20 minutes later*

PR #601, opened 2019-05-04
install python3-venv everywhere

abizer:
re: https://github.com/ocf/ocfweb/pull/460, https://github.com/ocf/ocfweb/pull/459


dkess:
tbh I always wondered why we didn't have this to begin with

ethanhs:
@dkess probably because we were using virtualenv everywhere, since that was used by many people for a long time.

PR #603, opened 2019-05-06
Add textfile exporter for Puppet environment

dkess:
This adds an `ocf_env` environment, which always has an associated label that says which environment the host is on. This could make alerts a bit more verbose, since they can now show if the machine is on someone's environment. We could also expand this later to include other info, like type (desktop, server, etc).

### DIFF
--- a/modules/ocf/templates/environment.prom.erb
+++ b/modules/ocf/templates/environment.prom.erb
@@ -1,0 +1,8 @@
+<%
+def prometheus_escape(text)
+    return text.gsub(/([\\"\n])/, {"\n"=>"\\n", "\\"=>"\\\\", "\""=>"\\\""})
+end
+-%>
+# HELP ocf_env The Puppet environment this node is on
+# TYPE ocf_env gauge
+ocf_env{environment="<%= prometheus_escape(@environment) %>"} 1.0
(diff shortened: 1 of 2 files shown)


modules/ocf/templates/environment.prom.erb
```diff
@@ -0,0 +1,8 @@
+<%
+def prometheus_escape(text)
+    return text.gsub(/([\\"\n])/, {"\n"=>"\\n", "\\"=>"\\\\", "\""=>"\\\""})
```
kpengboy: The last kv can be written as `'"' => '\"'`, no? Or am I making things up

kpengboy: The last kv can be written as `'"' => '\n'`, no? Or am I making things up

dkess: It probably could, but I chose to stick with double quotes everywhere for consistency.

dkess: It probably could, but I chose to stick with double quotes everywhere for consistency.

dkess:
I've been holding off on merging this since @redplanks suggested looking into https://github.com/voxpupuli/puppet-prometheus_reporter, which would be a lot more useful. Unfortunately I'm not sure how to install it.

PR #611, opened 2019-05-08
Revert "Temp stop managing desktop drivers since nvidia driver packag…

redplanks:
…es are auto-uninstalling right now."

This reverts commit 78bcdeb1a2e0cf9f909e7c9c244e797f7f1c2bf0.

### DIFF
--- a/modules/ocf_desktop/manifests/init.pp
+++ b/modules/ocf_desktop/manifests/init.pp
@@ -11,6 +11,7 @@ class ocf_desktop {
 
   include ocf_desktop::crondeny
   include ocf_desktop::defaults
+  include ocf_desktop::drivers
   include ocf_desktop::firewall_output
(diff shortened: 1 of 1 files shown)


redplanks:
Tested on cyanide. Everything installs properly now.

dkess:
We need to reboot each desktop after the driver gets updated. So instead of pushing this to master right away, we should individually put the affected machines on your environment, let it upgrade, reboot. Then once they are all upgraded, we can merge this and put them all on production.

jvperrin:
Or push to master after the lab closes and reboot all then? We can `ssh-list` out the reboot too since you can run a shutdown [without entering a password](https://github.com/ocf/puppet/blob/master/modules/ocf/templates/sudoers.erb#L33) for this kind of occasion.

PR #613, opened 2019-05-08
Move puppetlabs-kubernetes to working patched version

fydai:
Upstream PR is open, should be merged in the next release. Meanwhile this allows the kubernetes masters to be set back to production.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -17,7 +17,7 @@ mod 'puppetlabs-concat',               '4.2.1'
 mod 'puppetlabs-firewall',             '1.12.0'
 mod 'puppetlabs-hocon',                '1.0.0'
 mod 'puppetlabs-inifile',              '2.2.2' # Dependency of puppetlabs-puppetdb
-mod 'puppetlabs-kubernetes',           '3.1.0'
+mod 'puppetlabs-kubernetes',           '4.0.1'
 mod 'puppetlabs-haproxy',              '2.2.0'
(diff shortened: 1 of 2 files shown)


Puppetfile
```diff
@@ -18,6 +18,7 @@ mod 'puppetlabs-firewall',             '1.12.0'
 mod 'puppetlabs-hocon',                '1.0.0'
 mod 'puppetlabs-inifile',              '2.2.2' # Dependency of puppetlabs-puppetdb
 mod 'puppetlabs-kubernetes',           '3.1.0'
+mod 'puppetlabs-kubernetes', :git => 'https://github.com/fydai/puppetlabs-kubernetes'
```
dkess: Should the line above this be removed?

fydai:
yes

Puppetfile
Baisang: You made your commit off of master of puppetlabs-kubernetes in your fork, so by doing this you would be upgrading the version we are using to their master branch at the same time.

You may want to consider rebasing your changes off of version 3.1.0 in your fork, tagging it as something else, and using that as the version.

Baisang:
That being said, if it still works 100%, feel free to resolve the issue I brought up.

fydai:
As my pull request has been merged, I have updated this pull request to use the newest version of puppetlabs-kubernetes, 4.0.1, which includes my changes.

PR #615, opened 2019-05-10
[HPC] Add timeout to nvidia_gpu_info, and add persistence daemon

redplanks:
In case NVIDIA crashes completely, `nvidia-smi` will hang, so we don't want puppet to hang forever or fail.

Adding the `nvidia-persistenced` package should load the `/dev/nvidia` files on boot; right now, they aren't loaded until someone runs a command that references them, and they can't be loaded from within a Singularity container.

### DIFF
--- a/modules/ocf_hpc/facts.d/nvidia_gpu_info
+++ b/modules/ocf_hpc/facts.d/nvidia_gpu_info
@@ -14,7 +14,12 @@ def main():
         print('nvidia_gpu_info={}')
         return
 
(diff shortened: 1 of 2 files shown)


kpengboy:
Somewhat unrelated, but perhaps we should have some monitoring for the case when the nvidia driver is unresponsive? Or automatic remediation?

redplanks:
We could do monitoring if a job is in the "completing" state for more than a few minutes, which is how we've found out nvidia has stopped responding in the past.

The only remediation is rebooting corruption as far as I can tell right now, and that's still a manual process unfortunately.

PR #616, opened 2019-05-12
add option to use new response policy zone

cg505:
This requires ocf/dns#44 to be merged first.


cg505:
it seems this is no longer necessary

PR #617, opened 2019-05-13
gen-desktop-leases: handle missing desktop-leases.conf

cg505:
This has been a bug all along, but we never noticed until I spun up dev-pestilence which didn't have a desktop-leases.conf (because this script was supposed to generate it).


dkess:
What usually generates this file? Are you sure that this is the best way to handle it? Perhaps we should ensure that this file is always present instead, so that way this script doesn't fail silently?

cg505:
The file is generated by this script. The only reason we read the file is to avoid writing it if nothing changes.

PR #618, opened 2019-05-13
Simple puppet-trigger motd modification (still kinda WIP)

Baisang:
Something like this would be helpful to have visibility as to why a host was switched to a specific environment.

We include a `--reason` flag in the script (and make it mandatory if we're switching to a non-production env), and write to `/etc/motd` after puppet has ran.

I'm not too sure how to make this work well if `-t` flag isn't included, since I think the puppet run is in the background, so it's likely it would overwrite the file after we finish editing it.

Another issue is after a host is already switched to an environment, if someone just uses `-t` to retrigger puppet without `-e` flag, it would also just wipe out the `/etc/motd` file

Any suggestions for these two problems would be appreciated, this is still very much WIP

### DIFF
--- a/modules/ocf/files/puppet-trigger
+++ b/modules/ocf/files/puppet-trigger
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 """Trigger puppet agent runs."""
 import argparse
+import datetime
 import os
 import re
 import socket
@@ -49,6 +50,20 @@ def switch_to_environment(env):
     )
 
 
+def update_motd(env, reason):
+    dt = datetime.datetime.now()
+    user = os.environ.get('SUDO_USER', 'unknown')
+    if env is None:
+        env = ''
+    with open('/etc/motd', 'a') as f:
+        f.write('[{dt}] Switched to current environment {env} by user {user} for reason: {reason}\n'.format(
+            dt=dt,
+            env=env,
+            user=user,
+            reason=reason,
+        ))
+
+
 def validate_environment(environment):
     if not re.match(r'[a-zA-Z_\-0-9]+$', environment):
         raise ValueError('environment has weird characters')
@@ -75,6 +90,10 @@ def main(argv=None):
         '--no-daemonize', action='store_true',
         help='run in foreground'
     )
+    parser.add_argument(
+        '-r', '--reason',
+        help='reason for switching environments; will update the motd'
+    )
 
     args = parser.parse_args(argv)
 
@@ -83,6 +102,9 @@ def main(argv=None):
         return 1
 
     if args.environment:
+        if args.environment != 'production' and not args.reason:
+            log('You must specify a reason to switch to a non-production environment')
+            return 1
         switch_to_environment(args.environment)
 
     if args.test:
@@ -96,6 +118,9 @@ def main(argv=None):
         flags.append('--no-daemonize')
 
     trigger_run(flags)
+
+    if args.reason:
+        update_motd(args.environment, args.reason)
 
 
 if __name__ == '__main__':


Baisang:
for Increased Visibility we can later add functionality to email to a common mailing list as well, that way all environment switches can be logged and tracked

modules/ocf/files/puppet-trigger
```diff
@@ -49,6 +50,20 @@ def switch_to_environment(env):
     )
 
 
+def update_motd(env, reason):
+    dt = datetime.datetime.now()
+    user = os.environ.get('SUDO_USER', 'unknown')
+    if env is None:
+        env = ''
+    with open('/etc/motd', 'a') as f:
+        f.write('[{dt}] Switched to current environment {env} by user {user} for reason: {reason}\n'.format(
```
ethanhs: Will this reset when the environment is put on production? Also I don't know, but I was taught updating `/etc/update-motd.d` was better. Though perhaps not in this instance?

modules/ocf/files/puppet-trigger
```diff
@@ -83,6 +102,9 @@ def main(argv=None):
         return 1
 
     if args.environment:
+        if args.environment != 'production' and not args.reason:
+            log('You must specify a reason to switch to a non-production environment')
```
ethanhs: Maybe reword this to point to the `--reason` flag? Something like `You must specify a reason with the --reason flag to switch off of the production environment`.

modules/ocf/files/puppet-trigger
jvperrin: I believe this will reset on the next puppet run actually, regardless of environment: https://github.com/ocf/puppet/blob/8ec63f78a25c28dfb4f00ba60279dc71bfc1c58e/modules/ocf/manifests/motd.pp#L14

What about writing the reason to a separate file and only displaying it in the motd if it's not on the production env?

abizer:
maybe have a file like /etc/puppet-reason and have it `cat`ed by all the default .shellrcs?

Baisang:
After discussion with @jvperrin at the OCF satellite lab in SF today, I propose the following solution:

1. Have puppet require a file resource, `/etc/puppet-reason` to exist
2. Add a small script to puppet that is distributed to all hosts, `motd_reason` or something, that appends `/etc/puppet-reason` to `/etc/motd` (and does nothing if environment is production)
3. In puppet, have `motd_reason` exec after the `/etc/motd` file is created, ensuring we always can append the reason
4. The puppet-trigger script would just overwrite `/etc/puppet-reason` whenever a reason is provided

This way
1. Switching env to production will always clear the reason 
2. Switching env to non-production with a reason, and then re-triggering puppet without specifying a reason, will keep the same reason
3. We ensure puppet doesn't need to overwrite `/etc/motd` file all the time on production env puppet runs

cg505:
@Baisang It seems like this will create a lot of puppetspam because /etc/motd will be "changed" each puppet run when a node isn't on production. It seems like it might be cleaner to somehow have puppet include this file in the motd using the templating system, so we don't write the file and then immediately overwrite it.

Baisang:
@cg505 @dkess says you shouldn't read the puppet mailing list

Baisang:
Joking aside, I think there's a separate problem of why are we rerunning puppet so often when people have switched puppet from production to their own environment? I feel like that's out of scope for this change

redplanks:
Puppet runs happen every half hour cause of cron regardless of env though.

Baisang:
@redplanks yeah, I'm saying we should change that, but doesn't have to be as part of this change

kpengboy:
1. Agreed that it would be best to log reasons for puppet environment changes somewhere.
2. How would your proposed system interact with direct changes to a host's puppet environment through `ldapvi`?

Baisang:
@kpengboy in that case, on the next puppet run, it would:

1. if env was changed to production, would not print anything in motd
2. if env was not production, would print whatever is in the motd (could be an inaccurate reason, i suppose). 

So, not a great interaction. But is changing env using ldapvi common? I don't think that should be encouraged practice at all.

kpengboy:
> But is changing env using ldapvi common?

To my generation of root staff, this is actually not that strange of an idea. Before ~2016, it was the only way to switch the puppet environment.

Ideally, we'd be able to associate a reason with a particular puppet change. Then a direct puppet environment change in LDAP would have no corresponding reason and would not display any. But this really is a wishlist feature anyway.

ethanhs:
I think it would be better to have *something* working. Since ldapvi is (hopefully) going to become more rare in usage over time, I think we should fix things for the common case, and if we do want to support modifications via ldapvi, perhaps document somewhere that one should update the motd?

abizer:
we can fix both of these problems, I think. `puppet-trigger` writes to /etc/puppet-reason and when puppet runs, it cats that file into the motd. If puppet is on a non-production environment, and that file is empty, the run panics, which works regardless of whether the change was made via puppet-trigger or ldapvi. if the run is on production, puppet deletes that file.

this would however be a major hindrance to the useful but uncommon "move all hosts to my env to purge something" model we occasionally employ.

redplanks:
We can always do "make the purge commit to master and revert" instead, even if that is a bit messier history.

cg505:
After conversation with @Baisang, we figured out a way to do this without having the puppet run as "Changed" each time.

Ideally, we could just include `/etc/motd-reason` into the template for `/etc/motd`, but it doesn't seem like there's a good way to include a file from the host into a puppet template. Instead, we came up with the following approach.

1. `puppet-trigger` writes a change to `/etc/motd-reason`.
2. Puppet will write the motd template into `/etc/motd-puppet` instead of `/etc/motd`.
3. Puppet will use an Exec to concatenate these files, notified on both files.

Baisang:
@vaibhavjayaraman also said he was interested in working on this, so I'll let him have it

PR #622, opened 2019-05-13
move texlive to desktops only

abizer:


kpengboy:
What's so wrong about running TeX on tsunami though?

abizer:
it's still on tsunami, just installed manually, and has been removed from the other machines because it needlessly takes up 2.5GB of disk space

kpengboy:
If we intend to keep this on tsunami, perhaps record that explicitly in Puppet?

PR #624, opened 2019-05-16
add /munin redirect

MrMinos:
There's `/mon` shorturl but `/munin` would be nice to have.


MrMinos:
Tested on `death`

MrMinos:
thanks @jvperrin @abizer

PR #627, opened 2019-05-17
Remove mesos slave from desktops

yvbbrjdr:
Signed-off-by: Jerry Zhou <jerryzhou@ocf.berkeley.edu>


abizer:
we need purge the resources of this class too

PR #628, opened 2019-05-18
Make ocf::utils depend on the git package

fydai:
I just had a puppet run fail because ocf::utils tried to execute before git was installed.


redplanks:
If that's possible, maybe we should make installing git in the puppet first stage instead

fydai:
The puppet run only failed because networking was disabled in the BIOS, so there is no need for this pull request.

PR #629, opened 2019-05-21
update-groups: don't add pubstaff

abizer:
I don't remember why we continue to add them, but since the pubstaff account isn't used anymore, I think it should be safe to remove


kpengboy:
FYI I think back in the day it was so we could compare pubstaff's login stats to individual staffers' login stats.

PR #630, opened 2019-05-23
Add nfs exports for create on kubernetes

d00ble:
Allows for create to be run on kubernetes without having to pin it to dataloss.

### DIFF
--- a/modules/ocf_filehost/manifests/init.pp
+++ b/modules/ocf_filehost/manifests/init.pp
@@ -12,28 +12,7 @@ class ocf_filehost(
       # We don't root_squash admin, ssh, or apphost because they need to access
       # /services/crontabs/$server/ as root.
       options => ['rw', 'fsid=0', 'no_subtree_check', 'no_root_squash'],
-      hosts   => ['admin', 'www', 'dev-www', 'ssh', 'dev-ssh', 'apphost', 'dev-apphost'];
-
-  '/opt/homes/services/discourse':
-      options => ['rw', 'no_subtree_check'],
-      hosts   => lookup('kubernetes::worker_nodes');
-
-  '/opt/homes/services/mastodon':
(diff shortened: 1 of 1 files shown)


fydai:
I would add the workers export onto the /opt/homes export too.

d00ble:
> I would add the workers export onto the /opt/homes export too.

I dont quite get what you mean fydai

d00ble:
I suppose if I'm adding the worker nodes to the /opt/homes export then I can remove all the other exports because they're redundant?

cg505:
Does create actually need/use `/opt/homes/services`? I can't find any reference to this in the create codebase (besides the k8s/marathon definitions).

fydai:
I guess we can just add the worker nodes to /opt/homes and remove everything else. That doesn't seem that great from a security perspective (kubernetes workers can now access all of NFS) but that seems to be the best solution as of now.

d00ble:
> Does create actually need/use `/opt/homes/services`? I can't find any reference to this in the create codebase (besides the k8s/marathon definitions).

/opt/homes/services/http/users holds all the webdirs for users. I suppose we could just export just that directory, but the mastodon service mounted the whole directory so I just copied that.

cg505:
>/opt/homes/services/http/users holds all the webdirs for users.
Right, of course
>the mastodon service mounted the whole directory
Not sure what you mean. Mastodon mounts `/opt/homes/services/mastodon`. It seems like the equivalent mount for create would be `/opt/homes/services/http` or `/opt/homes/services/http/users`

d00ble:
@cg505 Sorry typo, that should say mesos not mastodon.

cg505:
ah

cg505:
it's fine I guess then

d00ble:
I'll merge if there are no more suggestions

PR #631, opened 2019-05-24
Run nvidia-modprobe on reboot on HPC compute nodes.

redplanks:
Mitigates [this bug](https://github.com/sylabs/singularity/issues/2203).

I have no idea why this works; the `nvidia-uvm` kernel module is already loaded on boot, and just `nvidia-modprobe -u` doesn't fix it.


kpengboy:
> I have no idea why this works
mystery.fuqu.jp

PR #635, opened 2019-05-29
Minimal configuration for ubuntu on desktops

fydai:


fydai:
Whoops wrong repository

PR #638, opened 2019-05-30
Add XMPP server to dev-flood

dkess:
I want to install an XMPP server so we can get Biboumi running, which will allow users to access IRC from XMPP clients (could be good for mobile). It's a Slack alternative that's easy for us to set up and maintain.

We can also potentially provide XMPP as a service to our users.

Next PR will be adding this to actual flood. Then I'll work on Biboumi.

### DIFF
--- a/modules/ocf_xmpp/manifests/init.pp
+++ b/modules/ocf_xmpp/manifests/init.pp
@@ -1,0 +1,57 @@
+class ocf_xmpp {
+  # Needed to connect to MySQL
+  package { ['lua-dbi-mysql']: }
+
+  user { 'prosody':
+    groups  => 'ssl-cert',
+    require => [Package['prosody'], Package['ssl-cert']],
+  }
+
+  ocf::firewall::firewall46 {
+    '101 allow xmpp':
+      opts   => {
+        chain  => 'PUPPET-INPUT',
+        proto  => 'tcp',
+        dport  => [5222, 5269],
+        action => 'accept',
+      }
+  };
+
+  $vhost_name = $::host_env ? {
+    'dev'  => 'dev-xmpp.ocf.berkeley.edu',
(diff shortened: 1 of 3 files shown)


dkess:
this is tested and working btw

dkess:
According to jvperrin in IRC, we can use DNS to request a cert for `ocf.berkeley.edu`.

dkess:
That will be handled in a later PR, note that this one won't actually install Prosody on `flood` since it only modifies the hiera classes for `dev-flood`.

modules/ocf_xmpp/manifests/init.pp
```diff
@@ -0,0 +1,57 @@
+class ocf_xmpp {
+  # Needed to connect to MySQL
+  package { ['lua-dbi-mysql']: }
+
+  user { 'prosody':
```
fydai: use ocfprosody to not conflict with usernames?

dkess: The `prosody` name has been reserved for a while, and this is the default name used by the package anyways.

modules/ocf_xmpp/manifests/init.pp
```diff
@@ -0,0 +1,57 @@
...
+        chain  => 'PUPPET-INPUT',
+        proto  => 'tcp',
+        dport  => [5222, 5269],
+        action => 'accept',
+      }
+  };
+
+  $vhost_name = $::host_env ? {
+    'dev'  => 'dev-xmpp.ocf.berkeley.edu',
+    'prod' => 'ocf.berkeley.edu',
```
kpengboy: I think this should be `xmpp.ocf.berkeley.edu`?

dkess: Yeah, this is correct

PR #639, opened 2019-05-30
kubernetes: solidify ocfroot-binding and ocfstaff-view-binding

dkess:
These ClusterRoleBindings were applied but never actually codified anywhere. This puts them in Puppet, so they'll get re-applied if we ever need to reprovision the cluster.


fydai:
I tried and it doesn't seem like you can through the staff web UI

PR #640, opened 2019-05-30
Kubernetes secrets

fydai:
This series of pull requests move secrets configuration away from kubernetes masters and into Jenkins, as it applies the other kubernetes configurations.
https://github.com/ocf/mastodon/pull/4
https://github.com/ocf/utils/pull/109
https://github.com/ocf/metabase/pull/7


modules/ocf_jenkins/manifests/init.pp
```diff
@@ -17,6 +17,14 @@
     stage => first,
   }
 
+  file {
```
abizer: logically shouldn't this be `/etc/ocf-kubernetes/secrets` since it's reasonable to expect there may be other k8s-related things we may want on machines of this class in the future

modules/ocf_puppet/manifests/puppetserver.pp
```diff
@@ -39,12 +39,12 @@
...
   }
 
-  # Allow kubernetes masters to access kubernetes secrets
+  # Allow Jenkins to access kubernetes secrets
   puppet_authorization::rule { 'kubernetes-secrets':
     match_request_path   => '^/puppet/v3/file_(content|metadata)s?/kubernetes-secrets$',
     match_request_type   => 'regex',
     match_request_method => ['get'],
-    allow                => suffix(lookup('kubernetes::master_nodes'), '.ocf.berkeley.edu'),
+    allow                => 'reaper.ocf.berkeley.edu',
```
abizer: use the service CNAME

modules/ocf_kubernetes/manifests/master.pp
```diff
@@ -62,21 +62,6 @@
       mode   => '0755';
   }
 
-  # These are needed because puppetlabs-kubernetes sets the permissions to 600
-  # but the certsign script, running under kubernetes-ca, needs to access them
-
-  File['/etc/kubernetes/pki'] {
```
abizer: why is this being removed?

modules/ocf_jenkins/manifests/init.pp
fydai: whoops that's a typo thanks

modules/ocf_kubernetes/manifests/master.pp
```diff
@@ -62,21 +62,6 @@
...
-  }
-
-  File['/etc/kubernetes/pki/ca.crt'] {
-    owner  => 'kubernetes-ca',
-  }
-
-  File['/etc/kubernetes/pki/ca.key'] {
-    owner  => 'kubernetes-ca',
-  }
-
```
dkess: Why did you remove these? Either way, it belongs in a separate commit.

modules/ocf_kubernetes/manifests/master.pp
fydai: Somehow both me and dkess wrote a fix to the kubernetes-ca permissions problem. His solution is further down and is much better than mine.

modules/ocf_puppet/manifests/puppetserver.pp
jvperrin: Does this work with `jenkins.ocf.berkeley.edu` instead, or does it have to be the fqdn of the host? It would be better if this was the service name to be more generic rather than the specific hostname of the jenkins host.

fydai: puppet doesn't like that

jvperrin: lol abizer just mentioned this too, guess I'm slow

modules/ocf_jenkins/manifests/init.pp
```diff
@@ -17,6 +17,14 @@
...
+    '/etc/ocf-kubernetes-secrets':
+      ensure => 'directory',
+      source => 'puppet:///kubernetes-secrets',
+      purge  => true,
+      mode   => '0700';
+  }
```
dkess: You might need to make this owned by the `jenkins-deploy` user.

modules/ocf_puppet/manifests/puppetserver.pp
fydai: It needs to be the fqdn, I tested jenkins.ocf.berkeley.edu first

modules/ocf_kubernetes/manifests/master.pp
dkess: This might still be needed, I think we added this when `puppetlabs-kubernetes` got upgraded and changed how file permissions work. Either way it should happen in a separate commit.

fydai: They are still needed, I just tested on deadlock.

PR #641, opened 2019-05-31
add cooperc key and box email to backups

cg505:
I feel like people who have access to backups should have to sign something. Also, we should clean out the keys directory.


dkess:
Are you sure you exported your key correctly? I'm not able to view the data using `gpg` or `pgpdump`.

cg505:
maybe gpg on my laptop is too new uhhh

cg505:
merged instead of rebase/squashing to preserve GPG signature on the commit

abizer:
will the real cooperc please sign these commits
i repeat, will the real cooperc please sign these commits
we're going to have a problem here

PR #642, opened 2019-06-01
kubernetes: stop nginx from forwarding all vhosts

dkess:
Previously, the front-end nginx would forward any request (with any host header) to the Kubernetes ingress controller. This commit gets rid of that catch-all server block, and only forwards vhosts with the http_vhost declaration (previously http_redirect). This is more in-line with how we handled Marathon vhosts.

This can be used to create "hidden" Kubernetes vhosts that aren't accessible publicly, for example snmp_exporter. We can still access these via the `downstream-proxy` which is firewalled off.


cg505:
This defeats the purpose of ingress-nginx and breaks `*.dev-kubernetes.ocf.berkeley.edu`. This makes it much more difficult to test new things on kubernetes, since you have to put all the masters on your env (which is a risky operation). Using the internal downstream proxy is difficult when you are not physically in the lab.

keur:
@cg505 `nginx` takes wildcard names and we should add an entry for `*.dev-kubernetes.ocf.berkeley.edu`.

Also, the commit didn't explain everything. If we want to run an internal HTTP service and make it accessible only within the OCF, we can't have `nginx` forwarding every host or the loadbalancer could be leveraged to bypass firewall rules on Kubernetes hosts.

cg505:
Perhaps we should run a separate instance of ingress-nginx to handle hosts that should only be accessible within the network, and then declaratively enumerate only those exceptions in puppet.

keur:
cc: @dkessler

dkess:
It's fine for us to manually declare servers here (for the purpose of bring explicit). We can use a wildcard server for the dev hostnames.

PR #644, opened 2019-06-01
Add xmpp and certs to prod flood

dkess:
The way the certs are handled here is a little wonky, but I'm not sure what a better way would be. Otherwise, this works!

### DIFF
--- a/modules/ocf_irc/manifests/init.pp
+++ b/modules/ocf_irc/manifests/init.pp
@@ -1,9 +1,20 @@
 class ocf_irc {
-  include ocf::ssl::default
   include ocf_irc::ircd
   include ocf_irc::services
   include ocf_irc::webirc
   include ocf_irc::znc
+
+  # The prod server also needs a cert for ocf.berkeley.edu, since we use XMPP
+  # domain delegation. See https://prosody.im/doc/certificates#which_domain
+  if $::host_env == 'prod' {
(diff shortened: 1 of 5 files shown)


modules/ocf_irc/manifests/init.pp
```diff
@@ -1,10 +1,21 @@
...
+  # domain delegation. See https://prosody.im/doc/certificates#which_domain
+  if $::host_env == 'prod' {
+    ocf::ssl::bundle { $::fqdn:
+      domains =>  ocf::get_host_fqdns() + ocf::get_host_fqdns('ocf.io') + ['ocf.berkeley.edu'],
+    }
+  } else {
+    ocf::ssl::bundle { $::fqdn:
+      domains =>  ocf::get_host_fqdns() + ocf::get_host_fqdns('ocf.io'),
+    }
+  }
```
kpengboy: I am really not a fan of putting XMPP-specific logic in the `ocf_irc` module. Have we considered making the `ocf_xmpp` module get its own SSL certs for its specific logic?

dkess: What about renaming the IRC module to `ocf_chat` and merging these? They all need to be on the same host anyways (I think)

dkess:
Going to close this PR and instead work on revamping how this is organized in Puppet.

PR #651, opened 2019-06-02
ocfweb: marathon => Kubernetes, take 2

fydai:


fydai:
It wasn't working because the masters were on my env, not on production.

PR #654, opened 2019-06-04
Move Prosody to ocf_irc

dkess:
This PR moves Prosody to the IRC module, since it needs to share its SSL cert. It also removes the external `mayflower-prosody` module since we don't really need to use it.

We also use a more recent version of Prosody from `stretch-backports`, which would have been difficult to do with the external Puppet module.


modules/ocf_irc/templates/prosody.cfg.lua.erb
```diff
@@ -0,0 +1,80 @@
...
+
+-- Store hashed passwords in database instead of plaintext
+authentication = "internal_hashed"
+
+storage = "sql"
+
+sql = {
+    driver = "MySQL",
+    database = "prosody",
+    username = "prosody",
```
abizer: this should probably be called ocfprosody

modules/ocf_irc/manifests/xmpp.pp
```diff
@@ -0,0 +1,60 @@
+class ocf_irc::xmpp {
+  # lua-dbi-mysql is needed to connect to MySQL
+  #package { ['lua-dbi-mysql']: }
```
kpengboy: Remove this dead code and move the comment where it is currently used

modules/ocf_irc/manifests/xmpp.pp
```diff
@@ -0,0 +1,60 @@
...
+        chain  => 'PUPPET-INPUT',
+        proto  => 'tcp',
+        dport  => [5222, 5269],
+        action => 'accept',
+      }
+  }
+
+  $vhost_name = $::host_env ? {
+    'dev'  => 'dev-xmpp.ocf.berkeley.edu',
+    'prod' => 'ocf.berkeley.edu',
```
kpengboy: Perhaps document that we are using a special SRV record so this is correct as is?

modules/ocf_irc/templates/prosody.cfg.lua.erb
dkess: The prosody username was already reserved before this. Unfortunately it might be too much effort to migrate it at this point

PR #656, opened 2019-06-05
Install Biboumi

dkess:
Biboumi is an XMPP-IRC bridge. This installs it on `flood`. I also included a separate commit with some minor config changes for Prosody.

### DIFF
--- a/modules/ocf_irc/manifests/biboumi.pp
+++ b/modules/ocf_irc/manifests/biboumi.pp
@@ -1,0 +1,42 @@
+class ocf_irc::biboumi {
+  ocf::repackage { 'biboumi':
+    backport_on => ['stretch'],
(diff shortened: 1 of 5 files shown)


modules/ocf_irc/manifests/biboumi.pp
```diff
@@ -0,0 +1,42 @@
...
+
+  file {
+    '/etc/biboumi/biboumi.cfg':
+      content   => template('ocf_irc/biboumi.cfg.erb'),
+      mode      => '0600',
+      show_diff => false,
+      require   => Package['biboumi'],
+      notify    => Service['biboumi'],
+      owner     => '_biboumi';
+
```
kpengboy: Nit: delete blank line

PR #657, opened 2019-06-06
Install moinmoin on death and tsunami

dkess:
MoinMoin is a web wiki software that we could offer to our users. This installs the base package in /opt, which will streamline the installation process-- users won't need to download this, since it'll already be installed.

The plan is to have an `easymoinmoin` script which installs the wiki in a user's webspace. Downloading and extracting this takes some time (lots of small files on NFS), so a pre-download will make this a lot faster!

### DIFF
--- a/modules/ocf/manifests/moinmoin.pp
+++ b/modules/ocf/manifests/moinmoin.pp
@@ -1,0 +1,18 @@
+class ocf::moinmoin {
+  # Version number is paired with its sha256 hash
+  # See https://moinmo.in/MoinMoinDownload
+  $versions = {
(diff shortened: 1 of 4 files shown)


modules/ocf/manifests/moinmoin.pp
```diff
@@ -0,0 +1,18 @@
+class ocf::moinmoin {
```
fydai: Can we put this as ocf::packages::moinmoin instead?

dkess: I'm not actually sure what the convention is here, since we aren't actually installing a package. I'll ask around.

dkess:
Gonna close this for now since I'm not convinced MoinMoin will be simple for us to manage. However this is still a good starting point if anyone wants to look into it in the future.

PR #660, opened 2019-06-07
Ceph

fydai:


hieradata/common.yaml
```diff
@@ -95,3 +95,10 @@ kubernetes::master_nodes:
 # Prometheus config
 prometheus::node_exporter::version: 0.17.0
 prometheus::node_exporter::extra_options: '--collector.textfile.directory /srv/prometheus'
+
+# Ceph configuration
+ceph::mons:
+  - pandemic
+  - scurvy
+
+ceph::fsid: a910ccd1-1116-4873-b3f9-c2cb7eed4e56
```
Baisang: Where is this magic # from?

modules/ocf_ceph/manifests/osd.pp
```diff
@@ -0,0 +1,33 @@
+# Ceph object store daemon configuration
+# Set ocf_ceph::osd::disks to be the disks allocated to ceph
+# If you have issues just dd if=/dev/zero the disk and try again
```
Baisang: What issues did you see here? Worth mentioning

modules/ocf_ceph/templates/ceph.conf.erb
```diff
@@ -0,0 +1,26 @@
...
+auth_supported = cephx
+
+# Write 3 replicas if possible
+osd_pool_default_size = 3
+
+# Allow pool deletion for debugging purposes
+mon_allow_pool_delete = true
+
+# Interal ceph stuff I don't understand, bump if ceph ever complains
+osd_pool_default_pg_num = 512
```
Baisang: Would put a link to here: https://stackoverflow.com/questions/40771273/ceph-too-many-pgs-per-osd which tries to explain it

modules/ocf_ceph/manifests/osd.pp
jvperrin: Agreed, and zeroing disks is perhaps a dangerous solution to suggest here without enough context

modules/ocf_ceph/templates/ceph.conf.erb
```diff
@@ -0,0 +1,26 @@
+[global]
+# ID of the ceph cluster
+fsid = a910ccd1-1116-4873-b3f9-c2cb7eed4e56
```
abizer: use the hiera variable?

hieradata/nodes/pandemic.yaml
```diff
@@ -2,6 +2,18 @@ classes:
     - ocf_kvm
     - ocf_mesos::slave
     - ocf_mesos::secrets
+    - ocf_ceph::mon
+    - ocf_ceph::osd
+
+ocf_ceph::osd::disks:
+    - /dev/sda
```
abizer: I kind of wish there was a better way to specify this

modules/ocf_ceph/manifests/admin.pp
```diff
@@ -0,0 +1,16 @@
+# Ceph admin key class, allows ceph to be used from the command line
+
+class ocf_ceph::admin {
+  include ocf_ceph
```
abizer: maybe `require` things instead?

modules/ocf_ceph/manifests/osd.pp
```diff
@@ -0,0 +1,33 @@
...
+
+  $bootstrap_osd_key = lookup('ceph::bootstrap_osd_key')
+
+  file {
+    default:
+      owner => 'ceph',
+      group => 'ceph';
+    '/var/lib/ceph/bootstrap-osd':
+      ensure => directory;
+    '/var/lib/ceph/bootstrap-osd/ceph.keyring':
```
abizer: this formatting is a bit weird

modules/ocf_ceph/manifests/osd.pp
```diff
@@ -0,0 +1,33 @@
...
+
+  $disks.each |$disk| {
+    exec { "prepare-${disk}":
+      command => "ceph-volume lvm prepare --data ${disk}",
+      unless  => "ceph-volume lvm list ${disk}",
+      notify  => Exec['activate-disks'],
+    }
+  }
+
+  exec { 'activate-disks':
```
abizer: is it possible to use a notifying arrow on the loop?

hieradata/common.yaml
fydai: UUID of the ceph cluster

Baisang: Yeah, I get that. What's important to document for magic #s is:
- Why is this number needed?
- How do we make new numbers and make sure they don't conflict with anything?
- When would we never need to change this?

modules/ocf_ceph/manifests/osd.pp
fydai: unsure how it would be improved

modules/ocf_ceph/manifests/osd.pp
fydai: I tried and it gave me a syntax error

hieradata/nodes/pandemic.yaml
fydai: You could probably do something like /dev/sda - /dev/sdh if somebody wanted to write a parser for that, I don't think it's worth the extra complexity though

modules/ocf_ceph/templates/ceph.conf.erb
fydai: done

modules/ocf_ceph/manifests/admin.pp
fydai: done

modules/ocf_ceph/manifests/mon.pp
```diff
@@ -0,0 +1,62 @@
+# Ceph monitor + manager + metadata server configuration
+# To bootstap, run puppet once, then run
+# sudo -u ceph ceph-mon --mkfs -i $hostname --keyring /var/lib/ceph/mon-bootstrap-keyring
```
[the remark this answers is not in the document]
fydai: Fixed.

fydai:
I have made all the requested changes, please re-review.

fydai:
I am unsure whether firewall rules are needed. If they are, I will add them in a future PR.

fydai:
Updated PR merged, https://github.com/ocf/puppet/pull/648

modules/ocf_ceph/manifests/mon.pp
```diff
@@ -26,11 +23,15 @@
       group  => 'ceph',
   }
 
+  exec { "bootstrap-mgr-${::hostname}":
+    command => "sudo -u ceph ceph-mon --mkfs -i ${::hostname} --keyring /var/lib/ceph/mon-bootstrap-keyring",
```
kpengboy: Would suggest setting `user => 'ceph'` instead

kpengboy:
As for firewall rules: will any servers in the DMZ (e.g. tsunami, death, desktops) be making connections to Ceph servers? On what ports? If so, you should add firewall rules covering those cases.

modules/ocf_ceph/manifests/init.pp
```diff
@@ -0,0 +1,26 @@
...
+
+class ocf_ceph {
+  require ocf::packages::ceph
+
+  $ceph_fsid = lookup('ceph::fsid')
+
+  $ceph_mons_list = lookup('ceph::mons')
+  $ceph_mons = $ceph_mons_list.join(',')
+  $ceph_mon_ips = $ceph_mons_list.map
+    |$node| { ldap_attr($node, 'ipHostNumber') }.join(',')
```
keur: nit: weird formatting. we usually would put `|$node|` on the lame line as the map function

modules/ocf_ceph/manifests/init.pp
```diff
@@ -0,0 +1,26 @@
...
+
+  $ip_addr = $::ipaddress
+
+  file { '/etc/ceph/ceph.conf':
+    content => template('ocf_ceph/ceph.conf.erb'),
+    require => Package['ceph'],
+  }
+
+  file { '/var/lib/ceph':
```
keur: this seems better suited for the `mon.pp` file since nothing here actually uses this.

modules/ocf_ceph/manifests/mon.pp
```diff
@@ -0,0 +1,62 @@
...
+  exec { "export-mgr-key-${::hostname}":
+    command => "ceph auth export mgr.${::hostname} -o ${mgr_keyring_file}",
+    creates => $mgr_keyring_file,
+  }
+
+  exec { "make-mds-key-${::hostname}":
+    command => "ceph auth add mds.${::hostname} ${mds_perms}",
+    unless  => "ceph auth get mds.${::hostname}",
+  } ->
+  exec { "export-mds-key-${::hostname}":
```
keur: chaining 5 execs together is kinda odd and indicates poor design. would recommend putting this in a script

modules/ocf_ceph/manifests/mon.pp
```diff
@@ -0,0 +1,62 @@
...
+
+  file {
+    '/var/lib/ceph/mon-bootstrap-keyring':
+      owner     => 'ceph',
+      group     => 'ceph',
+      show_diff => false,
+      content   => template('ocf_ceph/mon-keyring.erb'),
+  }
+
+  service { ["ceph-mon@${::hostname}", "ceph-mgr@${::hostname}", "ceph-mds@${::hostname}"]:
```
keur: put these on separate lines

modules/ocf_ceph/manifests/mon.pp
```diff
@@ -0,0 +1,62 @@
...
+class ocf_ceph::mon {
+  require ocf_ceph
+  require ocf_ceph::admin
+
+  $mon_key = lookup('ceph::mon_key')
+
+  $mgr_perms = "mon 'allow profile mgr' osd 'allow *' mds 'allow *'"
+  $mgr_keyring_file = "/var/lib/ceph/mgr/ceph-${::hostname}/keyring"
+
+  $mds_perms = "mon 'allow profile mds' osd 'allow *' mds 'allow *'"
```
keur: why is this a variable if it's being used once in the exec. this is really odd.

modules/ocf_ceph/manifests/mon.pp
```diff
@@ -0,0 +1,62 @@
+# Ceph monitor + manager + metadata server configuration
+
+class ocf_ceph::mon {
+  require ocf_ceph
+  require ocf_ceph::admin
+
+  $mon_key = lookup('ceph::mon_key')
+
+  $mgr_perms = "mon 'allow profile mgr' osd 'allow *' mds 'allow *'"
```
keur: ibid

modules/ocf_ceph/manifests/mon.pp
```diff
@@ -0,0 +1,62 @@
...
+  } ->
+  exec { "export-mds-key-${::hostname}":
+    command => "ceph auth export mds.${::hostname} -o ${mds_keyring_file}",
+    creates => $mds_keyring_file,
+  }
+
+  file {
+    '/var/lib/ceph/mon-bootstrap-keyring':
+      owner     => 'ceph',
+      group     => 'ceph',
```
keur: I'm seeing a lot of owner/group => ceph. might just want to use resource defaults for this module

modules/ocf_ceph/manifests/init.pp
jvperrin: Agreed, that's how ruby does it normally (which puppet is based off)

jvperrin: Personally I'd go with this:

```puppet
$ceph_mon_ips = $ceph_mons_list.map |$node| {
  ldap_attr($node, 'ipHostNumber')
}.join(',')
```

modules/ocf_ceph/manifests/init.pp
```diff
@@ -0,0 +1,26 @@
...
+  $ceph_mons_list = lookup('ceph::mons')
+  $ceph_mons = $ceph_mons_list.join(',')
+  $ceph_mon_ips = $ceph_mons_list.map
+    |$node| { ldap_attr($node, 'ipHostNumber') }.join(',')
+
+  $ip_addr = $::ipaddress
+
+  file { '/etc/ceph/ceph.conf':
+    content => template('ocf_ceph/ceph.conf.erb'),
+    require => Package['ceph'],
```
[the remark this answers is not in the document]
jvperrin: I'd actually say to keep this as-is, since it makes it clear what this actually depends on (it needs the package to be installed first), and that dependency should survive even if this resource is moved to another file, etc.

I'd agree it could technically be removed, but I think keeping it is better to be more explicit about dependencies and not rely on the implicit behavior of puppet to run in manifest order. Thoughts?

modules/ocf_ceph/manifests/mon.pp
jvperrin: As a note, I wouldn't actually use resource defaults, since they can leak between manifests, but using a variable for this and then passing that in to the `*` parameter of these resources should reduce the duplication a bit.

modules/ocf_ceph/manifests/osd.pp
jvperrin: I think spaces between the different files/directories certainly helps to show they are distinct

modules/ocf_ceph/manifests/init.pp
fydai: This is needed by osd.pp.

modules/ocf_ceph/manifests/mon.pp
fydai: This is two sets of 2 chained execs. Each chained exec has one which creates the key and another which exports the key. I think this is reasonable.

modules/ocf_ceph/manifests/mon.pp
fydai: Otherwise the exec gets too long

modules/ocf_ceph/manifests/mon.pp
fydai: ibid

modules/ocf/manifests/packages/ceph/apt.pp
```diff
@@ -0,0 +1,15 @@
...
+    id      => '08B73419AC32B4E966C1A330E84AC2C0460F3994',
+    source  => 'https://download.ceph.com/keys/release.asc',
+    require => Package['apt-transport-https'],
+  }
+
+  apt::source { 'ceph':
+    location => 'https://download.ceph.com/debian-luminous',
+    release  => $::lsbdistcodename,
+    repos    => 'main',
+    require  => [Apt::Key['ceph'], Package['apt-transport-https']],
```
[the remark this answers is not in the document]
fydai: This is copy pasted from the docker apt.pp, I think that they should be kept the same for consistency purposes.

modules/ocf_ceph/manifests/mon.pp
keur: It is not good

fydai:
it is time for me to continue working on this

fydai:
New PR here (https://github.com/ocf/puppet/pull/854)

PR #661, opened 2019-06-09
Repuppet slapd.conf

fydai:
I tested this on dev-firestorm and it works. We should keep as much configuration as possible in puppet, instead of in the incomprehendable mess that is /etc/ldap/slapd.d.

### DIFF
--- a/modules/ocf_ldap/manifests/init.pp
+++ b/modules/ocf_ldap/manifests/init.pp
@@ -21,6 +21,10 @@ class ocf_ldap {
   }
 
   file {
+    '/etc/ldap/slapd.conf':
+      content => template('ocf_/ldap/slapd.conf.erb'),
+      require => Package['slapd'];
+
     '/etc/ldap/schema/ocf.schema':
       source  => 'puppet:///modules/ocf_ldap/ocf.schema',
       require => Package['slapd'];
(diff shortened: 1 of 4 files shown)


fydai:
I tested this by making a dump of the database on firestorm, and reloading the database into dev-firestorm. There was no in-database config in dev-firestorm when I reload. I tested not adding the necessary schemas (like the cosine schema), and slapd always yelled at me on import I didn't have all the necessary schemas. This leads me to believe that the inetorgperson schema is unnecessary, as slapd was okay with me not adding it. memberOf won't work out unfortunately, but I found a workaround for checking whether an account is sorried by just checking their loginShell. Another test I can try is directly copy pasting the database over from firestorm to dev-firestorm. As far as I can tell, nobody is using the monitoring. I suspect that monitoring will be significantly easier to add, given that we're back to the file configuration.

cg505:
>it's a direct revert from 2 years in the past

@jvperrin do you mean that these files were in puppet before and got removed, or something else?

jvperrin:
Yeah, I mean that this is re-adding the files as they were 2 years ago, but from what I can see, not much was changed and what was changed is not relied on at least.

modules/ocf_ldap/manifests/init.pp
```diff
@@ -6,10 +6,10 @@
   package { ['slapd', 'libarchive-zip-perl']:; }
   service { 'slapd':
     subscribe => [
-      File[
-        '/etc/ldap/krb5.keytab',
+      File['/etc/ldap/schema/ocf.schema',
```
abizer: fix indentation

PR #665, opened 2019-06-13
modules/ocf: Add hidepid to all OCF systems

keur:
Remount procfs with the hidepid flag set to 2. This prevents non-root
users from seeing processes that aren't their own using either command
line tools or by going into /proc/<pid> directly.

### DIFF
--- a/modules/ocf/facts.d/polkit-priv-drop
+++ b/modules/ocf/facts.d/polkit-priv-drop
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -euo pipefail
+
+FACT="polkit_priv_drop"
+pkg=$(dpkg-query -s policykit-1 2>/dev/null)
+[[ $? -ne 0 ]] && echo "$FACT=false"
+version=$(echo "$pkg" | grep Version | cut -d' ' -f2)
+if dpkg --compare-versions "$version" 'ge' '0.115'; then
+    echo "$FACT=true"
+else
+    echo "$FACT=false"
+fi
--- a/modules/ocf/manifests/hidepid.pp
+++ b/modules/ocf/manifests/hidepid.pp
@@ -1,4 +1,22 @@
 class ocf::hidepid {
+
+  $group_cap_ptrace = 'capptrace'
+
+  group { $group_cap_ptrace:
+    ensure => present,
+    name   => $group_cap_ptrace,
+    system => true,
+    before => Mount['/proc'];
+  } ->
+
+  ocf::systemd::override { 'hidepid':
+    unit    => 'systemd-logind.service',
+    content => "[Service]\nSupplementaryGroups=${group_cap_ptrace}\n",
+  }
+
+  # NOTE: When policykit is eventually upgraded to version 0.115 we will need
+  # to make a similar change to that unit file.
+
   mount { '/proc':
     # Remounts proc with hidepid=2. This prevents users from seeing
     # processes that aren't their own using command line tools and
@@ -7,6 +25,6 @@ class ocf::hidepid {
     remounts => true,
     device   => '/proc',
     fstype   => 'procfs',
-    options  => 'remount,rw,hidepid=2',
+    options  => "rw,hidepid=2,gid=${group_cap_ptrace}";
   }
 }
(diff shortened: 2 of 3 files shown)


keur:
@jvperrin I've tested this on my staff vm. It applies cleanly to a running system, and `ensure => mounted` adds an entry to `/etc/fstab`.

kpengboy:
As pointed out on IRC: may need to address https://wiki.debian.org/Hardening#Mounting_.2Fproc_with_hidepid

modules/ocf/manifests/hidepid.pp
```diff
@@ -0,0 +1,12 @@
+class ocf::hidepid {
+  mount { '/proc':
+    # Remounts proc with hidepid=2. This prevents users from seeing
+    # processes that aren't their own using command line tools and
+    # by manually going into /proc/<pid>.
+    ensure   => mounted,
+    remounts => true,
+    device   => '/proc',
+    fstype   => 'procfs',
+    options  => 'remount,rw,hidepid=2',
```
kpengboy: I think the `remount` is supplied by the above `remounts => true` already

keur:
This needs to be fixed for buster. They started using capabilities for lots of packages instead of setuid so this won't work.

keur:
@kpengboy I made the changes to work with `systemd-logind`. The version of `policykit-1` on buster is 0.105-25, so it is not the version that drops privs.

modules/ocf/manifests/hidepid.pp
```diff
@@ -0,0 +1,36 @@
...
+
+  mount { '/proc':
+    # Remounts proc with hidepid=2. This prevents users from seeing
+    # processes that aren't their own using command line tools and
+    # by manually going into /proc/<pid>.
+    ensure   => mounted,
+    remounts => true,
+    device   => '/proc',
+    fstype   => 'procfs',
+    options  => "rw,hidepid=2,gid=${group_cap_ptrace}";
```
kpengboy: Needs dependency on `$group_cap_ptrace`

modules/ocf/manifests/hidepid.pp
```diff
@@ -0,0 +1,36 @@
...
+
+  ocf::systemd::override { 'hidepid-logind':
+    unit    => 'systemd-logind.service',
+    content => "[Service]\nSupplementaryGroups=${group_cap_ptrace}\n",
+  }
+
+  if str2bool($::polkit_installed) {
+    # NOTE: Futureproof for policykit for version 0.115. Since polkit < 0.115
+    # runs as root this change incurs no security cost.
+    ocf::systemd::override { 'hidepid-polkit':
```
kpengboy: I feel like having this unconditionally might be better, unless there is really a reason we must check for existence of PolKit

keur: This will crash on death for example, because `polkit` is not installed there. Some other package is pulling it in that is not present on all of our systems.

keur: I have a hunch that `polkit` is only on systems with `cups`. After poking around with some rdepends on supernova, it looks like `colord` is the only hard dependency on `polkit`, and `cups` on `colord`.

kpengboy: Also: according to that Debian wiki page, since polkitd itself does the privilege drop, adding a supplementary group doesn't help. Adding the `polkitd` user to `$group_cap_ptrace` might though.

keur: Hmm... That would work. What about adding `AmbientCapabilities=CAP_SYS_PTRACE` to the unit?

modules/ocf/manifests/hidepid.pp
```diff
@@ -0,0 +1,42 @@
...
+  $group_cap_ptrace = 'capptrace'
+
+  group { $group_cap_ptrace:
+    ensure => present,
+    name   => $group_cap_ptrace,
+    system => true,
+    before => Mount['/proc'];
+  }
+
+  # systemd-logind and policykit need cap_sys_ptrace with hidepid=2.
```
kpengboy:
```suggestion
  # systemd-logind and policykit need to access info about all processes.
```

modules/ocf/facts.d/polkit-priv-drop
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
kpengboy: Suggestion: perhaps a check for the existence of the polkit user might be simpler? (I don't have strong feelings about this fwiw)

PR #667, opened 2019-06-14
Prune images older than a month

fydai:
Biohazard is running out of space. We don't use images older than a week anyways.


dkess:
Kubernetes caches images anyways, so we probably don't have to worry about these cleanups causing problems.

PR #669, opened 2019-06-14
Fix postgres cert renewal issue

fydai:
I tested this on lethe and puppet is finally working again.

### DIFF
--- a/modules/ocf_postgres/manifests/init.pp
+++ b/modules/ocf_postgres/manifests/init.pp
@@ -32,6 +32,20 @@ class ocf_postgres {
       };
   }
 
+  # Postgres expects the letsencrypt certs to be owned by root
+  # The renewal script expects the letsencrypt certs to be owned by ocfletsencrypt
+  # We change the owner to ocfletsencrypt before the renewal script runs
+  # Then we change the owner back to root after it runs
+  exec { 'chown-letsencrypt':
+    command => 'chown -R ocfletsencrypt:ssl-cert /var/lib/lets-encrypt/certs',
+    before  => Class['Ocf::Ssl::Default'],
+  }
+
+  exec { 'chown-root':
+    command => 'chown -R root:ssl-cert /var/lib/lets-encrypt/certs',
+    require => Class['Ocf::Ssl::Default'],
+  }
+
   Class['Ocf::Ssl::Default'] ~> Class['Postgresql::Server']
 
   file {


modules/ocf_postgres/manifests/init.pp
```diff
@@ -32,6 +32,22 @@
       };
   }
 
+  # Postgres expects the letsencrypt certs to be owned by root
+  # The renewal scrypt expects the letsencrypt certs to be owned by ocfletsencrypt
```
jvperrin: nit: `scrypt` -> `script`

modules/ocf_postgres/manifests/init.pp
```diff
@@ -32,6 +32,22 @@
...
+  # We change the owner to ocflets-encrypt before the renewal script runs
```
jvperrin: nit `ocflets-encrypt` -> `ocfletsencrypt`

modules/ocf_postgres/manifests/init.pp
```diff
@@ -32,6 +32,22 @@
...
+    before  => Class['Ocf::Ssl::Default'],
+  }
+
+  exec {'chown-root':
+    command => 'chown -R root:ssl-cert /var/lib/lets-encrypt/certs',
+    require => Class['Ocf::Ssl::Default'],
+  }
+
+  # Change owner of /var/lib/lets-encrypt-certs to ocfletsencrypt
+  # postgres expects Class
```
jvperrin: I think both of these comments are too far down? At least they don't appear to pertain to the line below here.

modules/ocf_postgres/manifests/init.pp
```diff
@@ -32,6 +32,22 @@
...
+  # Postgres expects the letsencrypt certs to be owned by root
+  # The renewal scrypt expects the letsencrypt certs to be owned by ocfletsencrypt
+  # We change the owner to ocflets-encrypt before the renewal script runs
+  # Then we change it back after the renewal script runs
+  exec { 'chown-letsencrypt':
+    command => 'chown -R ocfletsencrypt:ssl-cert /var/lib/lets-encrypt/certs',
+    before  => Class['Ocf::Ssl::Default'],
+  }
+
+  exec {'chown-root':
```
jvperrin: nit: space between `{` and `'`

fydai:
Having the owner as ocfletsencrypt causes postgres ssl to not work, as postgres expects the owner to be root. The only workaround I've found online is to copy over the letsencrypt certs somewhere else to be owned by root/postgres. (https://serverfault.com/questions/807024/use-letsencrypt-certificates-for-postgresql). Jamming in the execs was the best solution I came up with to allow 1) keep the certs owned by root when postgres needs to use it and 2) when renewing, have the certs owned by ocfletsencrypt

abizer:
I rejected this approach when I was working on it earlier because it's a one off hack. The correct way to address this, IMO, is to amend the logic in `ocf::ssl` to set permissions correctly based on what the `require`ing class requests as the ultimate owners of the certs. Hence the addition of the `owner` and `group` parameters to the class back when I added this module.

fydai:
I don't see a particularly elegant way of solving this issue in `ocf::ssl` either. How could it ensure that the final owner could be set as `root:ssl_cert`, but requiring dehydrated run as `ocfletsencrypt:ssl_cert` without just moving the `exec`s into `ocf::ssl`?. I think this is fine as a solution because it addresses the issue of postgres not getting its cert updated, and it is reasonably clear that it is a workaround for an issue for an interaction between postgres and dehydrated.

PR #673, opened 2019-06-17
temporarily keep desktops awake during summer atdp hours

cg505:
i'll open a revert PR to be merged once ATDP finished


fydai:
8am please, they're already here today

PR #674, opened 2019-06-17
move pandemic to k8s and remove jaws

cg505:


fydai:
Remove kubernetes::worker from jaws in hiera

PR #675, opened 2019-06-17
ocf/hidepid: remount proc with gid=ocfstaff

keur:
Users in ocfroot can already see all processes with sudo.


dkess:
I think we should also let `ocfstaff` view processes.

Have you tested this?

keur:
I tested this and it works.

kpengboy:
This will not work because it will override the previous `gid=capptrace` clause. AFAICT the only way to work around that is to add all staff to `capptrace`. (Can we also think of a better name for that group? Perhaps `viewallprocs`?)

PR #676, opened 2019-06-18
xmpp: remove register module

dkess:
mod_register handles account creation and password changes. We don't need either, since both of these are handled by the makexmpp script.

We could wait until https://github.com/ocf/utils/pull/112 is merged to merge this, but since we're not getting account creation requests anyways I don't think it matters so much.


keur:
>dkess-patch-1

PR #680, opened 2019-06-21
Remove marathon

fydai:


modules/ocf_mesos/files/master/dns/mesos-dns.service
```diff
@@ -1,12 +0,0 @@
-[Unit]
-Description=Mesos DNS
```
jvperrin: You should also be able to remove this section from our DNS config, since we don't need to forward on DNS queries to this service any more: https://sourcegraph.ocf.berkeley.edu/github.com/ocf/dns@2dae37acd4ae27ec790417b2c0b81fc220587e10/-/blob/etc/named.conf.local#L9-12

modules/ocf_ns/templates/named.conf.options.erb
```diff
@@ -39,11 +39,7 @@ options {
 
   dnssec-enable yes;
   dnssec-lookaside auto;
-
-  // Don't validate DNSSEC yet, since mesos-dns doesn't support it
-  // dnssec-must-be-secure doesn't work either unfortunately :(
-  // Neither does using disable-algorithms on only the mesos zone
-  dnssec-validation no;
+  dnssec-validation yes;
```
jvperrin: I'm a bit concerned about including this change in with the rest of these changes, having it in a separate commit would be better overall in case we need to revert just this part. Anyway, not a big deal since we can just change this part of this file to revert if needed.

fydai: I tested this on pestilence for about 15 minutes, and everything seemed to work. Are there still issues with dnssec (https://github.com/ocf/dns/issues/25)?

modules/ocf_mesos/files/master/dns/mesos-dns.service
fydai: I plan to go through all the other repos stripping out code later today. Thanks for the information though!

PR #681, opened 2019-06-22
kubernetes: add PersistentVolume for Discourse

dkess:
yeah it's happening


dkess:
fydai approved this from IRC

PR #682, opened 2019-06-23
Initial configuration to allow setting up OTP secrets

cg505:
This is the first PR that must be merged for OTP sudo. It sets up the puppetserver/desktops to be ready to create and store secrets.

### DIFF
--- a/modules/ocf/manifests/auth.pp
+++ b/modules/ocf/manifests/auth.pp
@@ -184,6 +184,13 @@ class ocf::auth($glogin = [], $ulogin = [[]], $gsudo = [], $usudo = [], $nopassw
   # sudo user/group access controls
   package { 'sudo': }
   file {
+    '/etc/otp':
+      mode      => '0400',
+      source    => 'puppet:///otp-secrets/',
+      recurse   => true,
+      purge     => true,
+      force     => true,
+      show_diff => false;
     '/etc/pam.d/sudo':
       source  => 'puppet:///modules/ocf/auth/pam/sudo',
       require => Package['sudo'];
--- a/modules/ocf_desktop/manifests/packages.pp
+++ b/modules/ocf_desktop/manifests/packages.pp
@@ -41,7 +41,7 @@ class ocf_desktop::packages {
     # notifications
     ['libnotify-bin', 'notification-daemon']:;
     # security tools
-    ['scdaemon']:;
+    ['scdaemon', 'yubikey-manager']:;
     # utilities
     ['wakeonlan']:;
     # Xorg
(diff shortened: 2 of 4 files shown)


modules/ocf_puppet/files/write-otp-config
```diff
@@ -0,0 +1,24 @@
...
+
+if [ "$#" -ne 2 ]; then
+    echo "Usage: write-otp-config <username> <yubikey public ID>"
+    exit 1
+fi
+
+username="$1"
+yubikeyID="$2"
+
+if grep -q "$username" /opt/puppet/shares/private/otp/yubikeys; then
```
fydai: grep -q "^$username"

dkess: actually should be `grep -q '^$username:`

modules/ocf_puppet/files/write-otp-config
```diff
@@ -0,0 +1,24 @@
...
+username="$1"
+yubikeyID="$2"
+
+if grep -q "^$username:" /opt/puppet/shares/private/otp/yubikeys; then
+    echo "$username already registered. Appending yubikey, not reconfiguring google authenticator."
+    sed -i "/^$username:/ s/$/:$yubikeyID/" /opt/puppet/shares/private/otp/yubikeys
+    exit 0
+fi
+
+echo "$username:$yubikeyID" >> /opt/puppet/shares/private/otp/yubikeys
```
fydai: does this add you twice? if so, shouldn't this be in a else block

cg505: see line 16

modules/ocf_puppet/files/write-otp-config
```diff
@@ -0,0 +1,24 @@
...
+
+set -euo pipefail
+
+if [ "$#" -ne 2 ]; then
+    echo "Usage: write-otp-config <username> <yubikey public ID>"
+    exit 1
+fi
+
+username="$1"
+yubikeyID="$2"
```
kpengboy: Nit: in this repository, it seems the established convention is to use lowercase_with_underscores

modules/ocf_puppet/files/write-otp-config
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
kpengboy: Inform me, how is this script intended to be run? Is this supposed to be run by the user or some other job?

cg505:
btw, with feedback from fydai, a lot of this script is pending rewrite. This is not in a mergable state right now

cg505:
I'm thinking that I'm going to move the script to utils and expect people to call it directly on lightning. We don't really need this amount of tooling, it's just all obscuring what's actually happening.

cg505:
okay, I removed the script, drastically simplified it, and moved it to ocf/utils#140

cg505:
would appreciate some new reviews since all the reviews are super old now (although the code hasn't really changed).

PR #683, opened 2019-06-23
[do not merge] require OTP for sudo

cg505:
This is built on top of #682 to change the PAM configuration for sudo to require a OTP as well as a root principal password. Ideally, this is a yubikey YubiOTP. Otherwise, it will fall back to a typical Google Authenticator style TOTP.

This should not be merged until after #682 and ocf/utils#140, and after all current rootstaff have used the code in those PRs to set their OTP secrets.


dkess:
Maybe re-submit this after #682 is merged? Looks like some of the changes from there are getting duplicated here.

modules/ocf/manifests/auth.pp
```diff
@@ -182,14 +182,25 @@
   }
 
   # sudo user/group access controls
-  package { 'sudo': }
+  package { ['sudo', 'libpam-google-authenticator', 'libpam-yubico']: }
```
jvperrin: I think this might cause duplicate resource declaration problems since it's declared here and also in ocf_puppet. I'd just remove it from ocf_puppet since it's already here (and add a comment about it).

cg505:
Yeah, the first two commits of this PR are also in #682 so that's why the diff is hard to parse.

cg505:
I set the target branch to otp-puppetserver so that the diff actually makes sense, but to be clear, in production it will work like: merge #682 and switch target back back to master, wait until everyone has OTP secrets regsitered, then merge this into master.

cg505:
y'all still want this or should I close it?

cg505:
Seems like there is not a lot of interest for this any more. #682 and ocf/utils#140 should be reverted.

PR #684, opened 2019-06-24
New printing setup to avoid PDF rasterization upon open

jxlin123:
**TL;DR: fydai suggested I work on a new printing setup to remove PDF rasterization upon opening a PDF (see Keur's pull request, https://github.com/ocf/puppet/pull/202) and instead rasterize locally right before sending the job off to whiteout. I've done this by setting up a local CUPS configuration for each machine, and added some improvements as well.**

Several months ago, fydai suggested changing the current printing setup to remove rasterization from the default PDF opener on the desktops. See [this Discourse thread](https://leprosy.ocf.berkeley.edu/forum/t/move-printing-rasterization-from-pdf-opener/222) for more information, as well as my response at the end. I have since added a few more enhancements, which hopefully will make printing PDFs less of a hassle for lab users.

In summary, I have
- Set up a local CUPS configuration that rasterizes PDFs via a Tea4CUPS filter, and then passes on the job to printhost (whiteout)
- Changed ImageMagick default memory limits to better support the rasterization process (my experiments have shown that the default memory limits sometimes result in the `convert` program failing, hence stopping any "rasterization")
- Removed [the old PDF rasterization setup](https://github.com/ocf/puppet/pull/202), which keur wrote to prevent printhost from being overloaded with unrasterized PDFs. With this new setup, there would be no need for rasterization upon opening a PDF on the desktops. **Important: note that one of my commits (specifically, 21b7a72) actually ensures that the old files are ABSENT in Puppet; if this pull request gets approved, I plan on following up with another PR to remove this code.**
- Re-enabled the built-in PDF viewers in Google Chrome and Firefox (since PDF printing should now all be rasterized)
- Removed the enforcer check for Google Docs on printhost. My experiments seem to suggest that Google Docs are printed as PDFs, so if my setup works, then there would be no need to disallow printing from Google Docs anymore 

Acknowledgements: major thanks to fydai for getting me started and for helping me along the way. My appreciation goes out to others who have helped me as well...


modules/ocf/manifests/packages/imagemagick.pp
```diff
@@ -0,0 +1,15 @@
...
+  package { ['imagemagick']:; }
+
+  # Allows rasterizing filter to handle more complex
+  # PDFs without running out of memory
+  augeas { 'Increase ImageMagick maximum memory':
+    incl    => '/etc/ImageMagick-6/policy.xml',
+    lens    => 'Xml.lns',
+    context => '/files/etc/ImageMagick-6/policy.xml',
+    changes => [
+      'set policymap/policy[1]/#attribute/value 1GiB'
```
jvperrin: Kudos for getting something using augeas working, it's definitely not easy :)

modules/ocf/files/packages/cups/raster-filter
```diff
@@ -0,0 +1,28 @@
...
+set -euo pipefail
+
+input=$(mktemp --suffix='.pdf')
+
+# Feed filter input into temporary file
+cat - > "$input"
+
+# If the file is not a PDF, then just pass it thorugh
+# (Note that all PDFs start with "%PDF")
+if [ "$(head -c 4 "$input")" != "%PDF" ]; then
```
dkess: I'd use `file -bi "$input"` here to get the mimetype, and then check if it's `application/pdf`.

modules/ocf/files/packages/cups/raster-filter
```diff
@@ -0,0 +1,28 @@
...
+    cat "$input"
+    exit
+fi
+
+# Rasterizing PDF...
+output=$(mktemp --suffix='.pdf')
+
+if ! convert -density 250 "$input" "$output" 2> /tmp/error; then
+    echo 'Convert failed.'
```
dkess: Have you ever seen a conversion fail? If this happens, will it get reported to us (root@ mailing list) and will we be able to see the error message?

modules/ocf/files/packages/cups/raster-filter
```diff
@@ -0,0 +1,28 @@
...
+# (Note that all PDFs start with "%PDF")
+if [ "$(head -c 4 "$input")" != "%PDF" ]; then
+    cat "$input"
+    exit
+fi
+
+# Rasterizing PDF...
+output=$(mktemp --suffix='.pdf')
+
+if ! convert -density 250 "$input" "$output" 2> /tmp/error; then
```
dkess: Instead of using a tempfile here, you can replace `"$output"` with `-` to output directly to stdout.

cg505: Is /tmp/error ever checked? What's the purpose of this file?

modules/ocf/manifests/packages/cups.pp
```diff
@@ -1,20 +1,33 @@
...
 
   # install cups
-  package { [ 'cups', 'cups-bsd' ]: }
+  package { [ 'cups', 'cups-bsd', 'cups-tea4cups' ]: }
 
   file {
     # set print server destination
     '/etc/cups/client.conf':
-      content => "ServerName printhost\nEncryption Always\n",
+      content => "ServerName localhost\nEncryption Always\n",
```
cg505: Does this change work on whiteout?

modules/ocf_desktop/manifests/defaults.pp
```diff
@@ -4,14 +4,12 @@
     source  => 'puppet:///modules/ocf_desktop/xsession/mimeapps.list';
   }
 
+  # temporary only, to get rid of files!
   file { '/usr/local/bin/pdf-open':
-      mode    => '0755',
-      source  => 'puppet:///modules/ocf_desktop/xsession/pdf-open',
-      require => Package['libimage-exiftool-perl'];
```
cg505: is this package still needed on desktops? If not, maybe we can remove it from `packages.pp`

modules/ocf/manifests/packages/cups.pp
jvperrin: So if I understand it, it _is_ already working on whiteout, this PR adds a local CUPS server that then forwards on to whiteout after doing local rasterization. So it's moving the rasterization step from the script for opening PDFs to a local CUPS server.

modules/ocf_desktop/manifests/defaults.pp
jvperrin: Should be possible to remove this (it was added along with the pdf-open script in https://github.com/ocf/puppet/commit/e34981f26c4d9da12b367e74654ffe1736e7ed6d#diff-aec440fa3888a0f9bae5001579017f62), I agree it would be a good idea to do so.

modules/ocf/manifests/packages/cups.pp
jxlin123: Yes, as @jvperrin mentioned, I'm adding a local CUPS server that forwards jobs to whiteout. If I recall correctly, whiteout has its own CUPS setup specified separately in Puppet, and does not use the Puppet class `ocf::packages::cups` at all. I recall fydai mentioning that the main machines affected by my changes are the desktops, tsunami, and supernova.

modules/ocf/files/packages/cups/raster-filter
jxlin123: Sounds good, will add another commit accordingly that checks for MIME types

modules/ocf/manifests/packages/cups.pp
ethanhs: Why should these changes affect tsunami/supernova?

cg505: ah, that makes sense. I misunderstood what's going on here

modules/ocf_desktop/manifests/defaults.pp
jxlin123: Will make a commit to do so, thanks!

modules/ocf/manifests/packages/cups.pp
jxlin123: Opening conversation again...

@ethanhs A quick Github search shows that `ocf::packages::cups` is used by `ocf_desktop`, `ocf_admin`, and `ocf_ssh`. I had previously considered limiting my changes to the desktops only, but I figured that I can also deploy my changes to supernova and tsunami, especially if this new pipeline is better... 

(Back then, I've briefly mentioned this to fydai, and he said that having my changes apply to `ocf::packages::cups` should be fine, as opposed to separately creating a CUPS configuration for desktops only)

However, if this isn't that good of an idea, then I can change my code accordingly and move y configuration under `ocf_desktop`...

jxlin123:
Ok, added some new commits... I still haven't addressed all the issues mentioned previously, so I'm going to continue working on them...

And thank you all for the comments!

modules/ocf/manifests/packages/cups.pp
ethanhs: Ah I didn't realize we are already running cups on tsunami/supernova. Thanks for the explanation, and your continued work on this!

modules/ocf/files/packages/cups/raster-filter
jxlin123: With previous versions of the setup, I have seen some conversions fail. (That's one of the reasons I've decided to up the ImageMagick memory limit.) So far, the script doesn't really do anything other than output stderr to `/tmp/error`, so I'm going to try to change that so an email gets sent out.

modules/ocf/files/packages/cups/raster-filter
jxlin123: Alas, `/tmp/error` isn't checked. I think I just wanted any error messages to get redirected somewhere else for debugging, and kind of left that there.

@dkess made a comment about having the script send out an email upon `convert` failure, which I'll look into implementing. Also, I'll remove the `output` temporary file and simply output to stdout.

modules/ocf/files/packages/cups/raster-filter
dkess: I'd recommend writing a script in `utils` that does something similar to https://github.com/ocf/ldapsync/blob/master/ldapsync/email_ldapsync_problem

Where it runs a command (or takes stdin) and sends a failure email to ocfroot if anything gets output.

modules/ocf/files/packages/cups/raster-filter
jxlin123: Actually, it seems like `convert` writes its result (to the specified output file) as it runs, so in the case that `convert` fails, the result will be a mangled PDF. Thus, perhaps outputting directly to stdout may not be a good idea, since we need to first check that `convert` successfully ran before sending off the result to printhost. In that case, I'll keep the temporary output file.

modules/ocf/files/packages/cups/raster-filter
jxlin123: Sounds good, I'm basing my script off the example you gave and parts of enforcer. 

The OCF utils repo mentions that util scripts should be for scripts "which might be executed by a person." Since the script I'm writing preferably should only be executed by my filter, I'm going to look into placing it under `/usr/local/bin`. (The enforcer script is also placed at the same location, but on whiteout.) Or is this not a good location?

modules/ocf/files/packages/cups/raster-filter
dkess: Hmm, in that case, sure, you can use temporary files.

modules/ocf/files/packages/cups/convert_failure
```diff
@@ -0,0 +1,48 @@
...
+        {stdin}
+
+
+        tea4cups environment variables:
+        {tea_vars}
+
+        ====
+        Hostname: {hostname}
+        """.format(
+    stdin=sys.stdin.read(),
```
dkess: The `sys.stdin.read()` call is blocking, which could cause problems in this script. I'd fill in this template within the main function.

modules/ocf/files/packages/cups/convert_failure
```diff
@@ -0,0 +1,48 @@
...
+    tea_vars='\n'.join('  {}: {}'.format(k, v)
+                       for k, v in os.environ.items() if k.startswith('TEA')),
```
dkess: In the test emails to `root@` I've seen, this section has been empty. Do you know why that is? If it's impossible to get these environment variables, we shouldn't use them, but they can be very useful so we should try to get them working.

modules/ocf/files/packages/cups/convert_failure
```diff
@@ -0,0 +1,48 @@
...
+    hostname=socket.getfqdn()
+)
+)
+
+if __name__ == '__main__':
```
dkess: I know it's a bit silly, but for the sake of consistency across the code, you should call a `main()` function here instead of putting the implementation in the if statement.

modules/ocf/files/packages/cups/convert_failure
jxlin123: Thanks for the suggestion! I will fill in my `ERROR_MSG` placeholders within main.

To clarify, when you say that the `sys.stdin.read()` call is blocking, can you explain what you mean?

dkess: It performs an IO operation that isn't guaranteed to return instantly-- if `stdin` isn't filled up, the rest of the program will never run.

It also falls into the realm of program logic, which you want to make sure stays in the main function (or is called from the main function). Constants can be declared at the top level, but this isn't a constant. See https://stackoverflow.com/a/1954796 for some more explanation on why we do this.

modules/ocf/files/packages/cups/convert_failure
jxlin123: Noted, will do

modules/ocf/files/packages/cups/convert_failure
jxlin123: I've noticed, and have been looking into this. For now, I haven't been able to get these environment variables, so I'm going to remove this section of the code, but I'm going to keep looking into this...

jxlin123:
Ok, I've addressed some of the issues brought up recently... thanks all, once again!

Regarding my use of another temporary file in `raster-filter`: running `convert` may take some time, so I wanted to ensure some sort of ordering during filter execution. The simplest way I could think of was to first output the stderr of `convert` to a tempfile, and then output the tempfile to `convert_failure` if it contains any data.

Also note that I've had to resolve a small merge conflict along the way...

modules/ocf/files/packages/cups/convert_failure
```diff
@@ -0,0 +1,50 @@
...
+""")
+
+
+def main():
+    if sys.stdin.isatty():
+        sys.stderr.write('script should not be called interactively. Aborting...')
+    else:
+        error_message = ERROR_MSG.format(stdin=sys.stdin.read(),
+                                         hostname=socket.getfqdn()
+                                         )
```
keur: Did you run `pre-commit`? Formatting is weird

modules/ocf/files/packages/cups/convert_failure
```diff
@@ -0,0 +1,50 @@
...
+Tea4CUPS filter.
+
+Jobs printed from one of the desktops, or from supernova or tsunami,
+first go through raster-filter, which calls convert to reduce the
+density of any PDFs, essentially rasterizing them. If convert happens
+to fail, raster-filter will pipe stderr from convert to this script,
+which takes in stdin and includes it in an email to root@ocf. Thus,
+this script should not be called casually by regular users!
+
+Based off the enforcer script for whiteout, and the
```
keur: It's nice that you are paying respects, but I don't think these two lines are necessary.

modules/ocf/files/packages/cups/printers.conf
```diff
@@ -0,0 +1,41 @@
+# Printer configuration file for CUPS v2.2.1
```
keur: How much of this custom? If it's only a couple lines we can use `augeas` on the config that cups pulls in.

modules/ocf/files/packages/cups/raster-filter
```diff
@@ -0,0 +1,34 @@
...
+if [ "${mimetype%%;*}" != "application/pdf" ]; then
+    cat "$input"
+    exit
+fi
+
+# Rasterizing PDF...
+output=$(mktemp --suffix='.pdf')
+error=$(mktemp)
+
+if ! convert -density 250 "$input" "$output" 2> "$error"; then
```
keur: You are using two distinct styles for your if statements in this script. Please stick to one.

jxlin123: Addressed... I've removed the `-e` option when setting Bash settings, and now I check the return status of `convert`... see commit https://github.com/ocf/puppet/pull/684/commits/6bfcfc5c0b0dbbf0123dcaa5752ce5afa2605534

modules/ocf/files/packages/cups/printers.conf
jxlin123: Turns out, most of the lines are custom... since upon installation, CUPS starts out with no printers... so I suppose I might as well just include the entire file

modules/ocf/files/packages/cups/convert_failure
jxlin123: Done, see commit https://github.com/ocf/puppet/pull/684/commits/f319632133609bca62de836b443b1f66c404b293

modules/ocf/files/packages/cups/convert_failure
jxlin123: Done, see commit https://github.com/ocf/puppet/pull/684/commits/f319632133609bca62de836b443b1f66c404b293

modules/ocf/files/packages/cups/convert_failure
jxlin123: Ok, the reason why these environment variables are not available is because Tea4CUPS only makes them available for prehooks and posthooks, not filters. As per the [Tea4CUPS website](http://www.pykota.com/software/tea4cups):

> To help your own commands, Tea4CUPS makes available as part of the environment several variables which can be used from the commands you use in pre and post hooks (but not filters)

Since you mentioned that these variables can be useful, I may restructure my `raster-filter` to be a Tea4CUPS prehook. Otherwise, I think what I have now should suffice...

cg505:
Manually rebased and merged to 774155238540d7cf5682d794635faf0f16f4e21f.

ethanhs:
Thank you so much for your continued work on this @jxlin123, I am sure I am among many people very excited and glad to see these changes.

PR #686, opened 2019-06-26
Add ocf_lb for front end kubernetes termination

fydai:
This is the start of creating an OCF-wide SSL-terminating reverse proxy. Currently all this does is replace the load balancer on the existing kubernetes masters, in a way that easily supports adding new kubernetes clusters (which is coming soon I promise). This also fixes the weird slowness that all of kubernetes has right now.

### DIFF
--- a/hieradata/nodes/monsoon.yaml
+++ b/hieradata/nodes/monsoon.yaml
@@ -1,0 +1,2 @@
+classes:
+    - ocf_lb
--- a/hieradata/nodes/pileup.yaml
+++ b/hieradata/nodes/pileup.yaml
@@ -1,0 +1,2 @@
+classes:
+    - ocf_lb
--- a/hieradata/nodes/whirlwind.yaml
+++ b/hieradata/nodes/whirlwind.yaml
@@ -1,0 +1,2 @@
+classes:
+    - ocf_lb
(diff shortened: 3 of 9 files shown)


keur:
Is there a reason we don't use the HAProxy Puppet library, like we did [here](https://github.com/ocf/puppet/pull/440/files)

dkess:
Why is `ocf_lb` on the old Marathon masters? This belongs in the Kubernetes module.

It doesn't necessarily have to be on the masters, this should work if you install this on each worker and have it proxy directly to the locally running ingress.

fydai:
The HAProxy puppet module, when declaring a balancermember, requires specifing a `listening_service` specifing a front end. I have no idea what this actually does, since this isn't apparent in the config, and you can just put a random string instead of the name of front end and it would generate the same config.

I didn't really want to specify an argument that is literally not used, so I decided to move away from puppetlabs-haproxy. The only thing we were doing with puppetlabs-haproxy anyways was creating a config file, which I believe is just better done in a template.

fydai:
The reason I created this module is to eliminate duplication of duties once more kubernetes clusters are set up. Right now, the ocf_kubernetes::master class has two roles:
1. Terminating SSL and load-balancing to the workers for traffic to enter the cluster
2. Being a kubernetes master
If more clusters are created with the current setup, job 1 would be duplicated among the masters of a new cluster. Thus there would be more moving parts, with the extra keepaliveds and the haproxies that might break, and are unnecessary.

dkess:
Why are you worried about job 1 getting duplicated though? We only have to write the Puppet code once, the specifics for each cluster can go in Hiera. That's not a lot of moving parts.

"Centralizing" the load balancing is worrying because you're creating a single point of failure. The whole point of a dev cluster is to be able to test out configuration changes without risking the prod cluster. If they share the same load balancer, we lose that benefit.

Furthermore, including a reverse proxy within a Puppet module isn't unusual. For every HTTP service we have on Puppet that I can think of, we manage the reverse proxy in the same module.

fydai:
Replaced with https://github.com/ocf/puppet/pull/691

PR #687, opened 2019-06-26
chrome/firefox: allow SPNEGO auth for keycloak

dkess:
This is necessary for automatic Keycloak logins from desktops.

This commit also removes RT's SPNEGO, but we stopped using SPNEGO for RT anyways so it won't have an effect.


cg505:
Not really a part of this PR, but we should also put the chrome policy file at `/etc/chromium/policies/managed/ocf_policy.json`: I think this will make it take effect for brave.

PR #688, opened 2019-06-26
backups: stop backing up ldap:/etc/ldap/slapd.d/

dkess:
With #661 this directory no longer exists. It's been causing rsnapshot sync operations to silently fail (and by extension causing the actual backup operation to neve run). We don't need to back this up anymore anyways since it's now managed within Puppet.


dkess:
I guess cron doesn't report if there's no output, even if it returns an error code?

Regular verbose output doesn't actually say that anything is wrong. Mega-verbose output mentions that there's an error, but doesn't say what it is. The actual error was in journalctl logs.

We probably don't need to change anything here, since we were still getting emails telling us backups were failing. Maybe we could add some text to that email advising us to check journalctl, though.

PR #689, opened 2019-06-29
brave: switch keyserver to keys.openpgp.org

dkess:
The PGP infra is under attack, this will protect us from the attack in case the Brave key gets hit.

See https://gist.github.com/rjhansen/67ab921ffb4084c865b3618d6955275f


dkess:
I searched for the fingerprint and it was in there, but if there's a URL we can download it from, that's fine too.

dkess:
I've opened #690 as an alternative to this PR, however I am not entirely convinced that's the best way to do things. I looked at the management UI for my own key on keys.openpgp.org and it looks like I don't have the ability to completely remove the key from the keyserver (I can only remove the associated email addresses).

If we can rely on the key always being on keys.openpgp.org, this PR is better, since the other PR leaves us vulnerable if Brave's S3 bucket ever gets compromised.

PR #690, opened 2019-06-30
brave: download PGP key from official location

dkess:
The PGP infra is under attack, this will protect us from the attack in case the Brave key gets hit.

See https://gist.github.com/rjhansen/67ab921ffb4084c865b3618d6955275f

### DIFF
--- a/modules/ocf/manifests/packages/brave/apt.pp
+++ b/modules/ocf/manifests/packages/brave/apt.pp
@@ -2,8 +2,8 @@
 class ocf::packages::brave::apt {
   apt::key { 'brave':
     ensure => refreshed,
-    id     => 'D8BAD4DE7EE17AF52A834B2D0BB75829C2D4E821',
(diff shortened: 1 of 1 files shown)


dkess:
I've expressed my concerns with this PR in https://github.com/ocf/puppet/pull/689#issuecomment-506999307, discussion should go there.

dkess:
On second thought, looks like it's possible to download the key and specify the fingerprint (which is what we do in other parts of the repo) so maybe we should do that? Manually downloading the file seems silly, this is what Puppet should be doing for us in the first place...

fydai:
I think we should just specify the brave url to download the key and the fingerprint (the id parameter), like we do in other places. It's consistent with the rest of our apt.pp and makes our code more independent of the pgp pool.

dkess:
I've amended this PR to be more in-line with how we handle PGP keys in other parts of the repo.

PR #691, opened 2019-07-01
Move kubernetes from nginx to haproxy

fydai:
The only reason we were on nginx in the first place was because of the kube webui.
Now that we have keycloak doing that, we can move back to haproxy.

We also add back the alias generation, so we don't need to duplicate a bunch of names.


modules/ocf_kubernetes/functions/gen_aliases.pp
```diff
@@ -0,0 +1,29 @@
...
+
+function ocf_kubernetes::gen_aliases(Array $services) >> Hash[String, Array[String]] {
+  $expanded_services = $services.map |$service| {
+    case $service {
+      String : {
+        ["${service}.ocf.berkeley.edu", [$service, "${service}.ocf.io"]]
+      }
+      Tuple[String, Array[String]] : {
+        $service_name = $service[0]
+  $other_aliases = $service[1].map |$alias_name| {
```
jvperrin: nit: indentation

modules/ocf_kubernetes/manifests/master/loadbalancer.pp
```diff
@@ -2,6 +2,38 @@
   include ocf::firewall::allow_web
 
   $kubernetes_worker_nodes = lookup('kubernetes::worker_nodes')
+  $kubernetes_workers = $kubernetes_worker_nodes.map |$worker|
+    { [$worker, ldap_attr($worker, 'ipHostNumber')] }
+
+  $kubernetes_services =
+  [
```
jvperrin: super nit: I'd just put this on the end of the previous line, the indentation makes it pretty clear what's going on anyways:

```
  $kubernetes_services = [
    'auth',
    [...]
  ]
```

modules/ocf_kubernetes/manifests/master/loadbalancer.pp
```diff
@@ -2,6 +2,38 @@
...
+    'inventory',
+    ['sourcegraph', ['sg']],
+    'static',
+    'templates',
+  ]
+
+  $kubernetes_aliases = ocf_kubernetes::gen_aliases($kubernetes_services)
+
+  $kubernetes_proxy =
+  [
```
jvperrin: (same here as the previous comment)

modules/ocf_kubernetes/manifests/master/loadbalancer.pp
```diff
@@ -2,6 +2,38 @@
   include ocf::firewall::allow_web
 
   $kubernetes_worker_nodes = lookup('kubernetes::worker_nodes')
+  $kubernetes_workers = $kubernetes_worker_nodes.map |$worker|
+    { [$worker, ldap_attr($worker, 'ipHostNumber')] }
```
jvperrin: nit: Ruby usually keeps the `|$var|` and the `{` or `do` on the same line, so I think making this like:

```
  $kubernetes_workers = $kubernetes_worker_nodes.map |$worker| {
    [$worker, ldap_attr($worker, 'ipHostNumber')]
  }
```

would make it clearer to read and more idiomatic puppet.

modules/ocf_kubernetes/functions/gen_aliases.pp
```diff
@@ -0,0 +1,29 @@
...
+          ["${alias_name}.ocf.berkeley.edu", "${alias_name}.ocf.io", $alias_name]
+  }
+        $aliases = [$service_name, "${service_name}.ocf.io"] + flatten($other_aliases)
+
+        ["${service_name}.ocf.berkeley.edu", $aliases]
+      }
+
+      default: {
+        fail('Expected String or Tuple[String, Array[String]]')
```
jvperrin: Might be good to provide the example at the start of the file here too? (or maybe just mention in this error message to see the example in this file for usage)

modules/ocf_kubernetes/functions/gen_aliases.pp
```diff
@@ -0,0 +1,29 @@
+# Expands ['mastodon', ('pma', ['phpmyadmin'])]
+# to { 'mastodon.ocf.berkeley.edu' => 'mastodon.ocf.io', 'mastodon']
+#      'pma.ocf.berkeley.edu       => ['pma.ocf.io', 'pma', 'phpmyadmin.ocf.berkeley.edu',
+#                                      'phpmyadmin.ocf.io', 'phymyadmin'] }
+
+function ocf_kubernetes::gen_aliases(Array $services) >> Hash[String, Array[String]] {
```
jvperrin: I'm not a huge fan of the amount of "magic" that I feel this function is doing, but big thanks for including types for it and I think it's still useful to have. I'd say to keep it, just a bit reluctantly :P

modules/ocf_kubernetes/templates/master/loadbalancer/haproxy.cfg.erb
```diff
@@ -0,0 +1,60 @@
+global
+  daemon
+  log localhost local0
+  # TODO: send to syslog
```
jvperrin: Do you mean the remote syslog server here (i.e. `democracy`)? If so, I think just locally logging is enough, it should get sent on to the remote syslog server by the local rsyslog instance.

modules/ocf_kubernetes/templates/master/loadbalancer/haproxy.cfg.erb
```diff
@@ -0,0 +1,60 @@
...
+  <%- end -%>
+
+  # send *.dev-kubernetes.ocf.berkeley.edu into the cluster
+  use_backend kubernetes if { hdr_end(host) -i dev-kubernetes.ocf.berkeley.edu }
+
+
+# Some services don't need SSL termination
+# Either because it has already been done (ocfweb, irc)
+# Or because it doesn't need it (snmp_exporter)
+# As all such traffic should be internal, it is placed on a firewall'd port
```
jvperrin: I'd make this clear that it's an externally firewalled port (using the firewall the university gave us), not a local one (with iptables), since we do use both.

Also, something to note here, but the thing proxying to this location must have done the canonicalization step already (for instance rewriting `irc.ocf.io` to `irc.ocf.berkeley.edu`), so I'm not sure if that's worth mentioning here or if it's worth including anything to redirect those cases that slip through proxies.

modules/ocf_kubernetes/templates/master/loadbalancer/haproxy.cfg.erb
```diff
@@ -0,0 +1,60 @@
...
+frontend lb-proxy
```
jvperrin: I'd name this something like `lb-internal-http` instead, since it's not necessarily just proxied services (`snmp_exporter` for instance)

(same for renaming the variable passed through to this template to `kubernetes_http_services` or something versus `kubernetes_https_services` for the other one maybe? Idk, just suggestions, naming things is hard :)

modules/ocf_kubernetes/functions/gen_aliases.pp
fydai: I found a solution that doesn't involve this file.

modules/ocf_kubernetes/manifests/master/loadbalancer.pp
fydai: Fixed

modules/ocf_kubernetes/templates/master/loadbalancer/haproxy.cfg.erb
fydai: Fixed.
There are just 4 of them, and all of them do the canonicalization step step already, so I don't think it's necessary to add that code.

modules/ocf_kubernetes/templates/master/loadbalancer/haproxy.cfg.erb
fydai: names changed

PR #692, opened 2019-07-02
mail: remove expansion_limit for vhosts

dkess:
A group was hosting a large mailing list with vhosted mail, and this limit was causing it to break (rt#8335). This commit allows for arbitrarily large vhost lists.

In the future, we should look into providing a proper mailing-list service for users who want this functionality.


dkess:
I guess I could make it 1000. I think the mailing list mentioned in the ticket was hitting ~400 members.

jvperrin:
Yeah, maybe 1000 is good, although I also don't want to just run into this issue again in the future, but if it's at 1000 maybe it'll be clearer a limit was hit?

dkess:
Changed to 1000

PR #693, opened 2019-07-04
ocf_ldap: Add ocf-ldap-overlay module

keur:
Load the overlay and enable the module.

The schema created in https://github.com/ocf/puppet/pull/668 will still
work as a virtual attribute.

The code for the overlay can be found at
https://github.com/ocf/ldap-overlay. It builds packages for Debian
Stretch and Debian Buster.


keur:
I wrote a script to purge the `ocfEmail` attribute and it is done for the database on `dev-firestorm`. `dev-firestorm` is on my puppet environment, and the `ocfEmail` attribute is the virtual attribute. I am running the script on prod ldap to remove the `ocfEmail` values to the database right now. Please no one merge until that is complete.

PR #694, opened 2019-07-04
kubernetes: add irclogs to loadbalancer

dkess:


dkess:
I'm still working on the repo, but my plan is to clone their GitHub repo into the docker image instead of forking (similar to what we do for Puppetboard).

cg505:
I believe we need to make non-trivial changes to the code that we could potentially upstream so that's why I was asking

PR #695, opened 2019-07-07
extrapackages: add python3-requests-oauthlib

dkess:
This package will be used for an upcoming utils script (easyfilters). It can be used to interact with Google APIs.


dkess:
I've always assumed utils deliberately depended only on things packaged within Debian... if I could pick a pip package I would have chosen Google's own auth lib (which is what they recommend in their documentation) but it's not packaged in Debian. fwiw this one still can get the job done just as well.

PR #697, opened 2019-07-08
ocf_syslog: Send kubernetes service logs to /var/log/remote/services

jvperrin:
Currently these are just piling up in `/var/log/remote/deployment-*` instead of being filtered per service (which also means ocfstaff cannot read them as-is since they are owned by `root:ocfroot` if they are not identified as per-service logs). This also adds information to the log about which kubernetes agent the log came from, which could be useful in debugging.

### DIFF
--- a/modules/ocf_syslog/files/ocf.conf
+++ b/modules/ocf_syslog/files/ocf.conf
@@ -73,7 +73,7 @@ ruleset(name="ocf-remote") {
     # through docker hosts quite quickly and we want to group logs by service
     # for docker, not by host.
     if (re_match($hostname, '^[a-f0-9]{12}$') and re_match($syslogtag, '^[a-zA-Z0-9:_-]+$')) or
-       re_match($hostname, '^.*deployment-[a-f0-9]+-[a-z0-9]+$') then {
+       re_match($hostname, '^.*-[a-f0-9]{8,9}-[a-z0-9]{5}$') then {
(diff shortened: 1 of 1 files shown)


jvperrin:
Hmmm, I just realized this doesn't cover the case with `create-deployment-*` or other deployments, so I've updated this to include those cases too. (and tested it on `democracy`, of course, to make sure logs like `services/create/create.log` are being written to)

jvperrin:
Also, after this is merged and a couple weeks go by, we should be able to clean up all the extra deployment logs in `/var/log/remote`, which would be nice.

cg505:
I don't think this is a good solution. There's no guarantee that the hostname will contain `deployment` in general. This is only because most of our kubernetes deployments have been named `deployment` or `<appname>-deployment`. Theoretically though, these could be named anything. I think a better solution might be catching the general case of <name>-<8-9 hex digits>-<5 alphanumeric chars>. The regex would then be `^.*-[0-9a-f]{8,9}-[0-9a-z]{5}`.

This still will only work for deployments I think, and not other kubernetes resources such as nodePorts, statefulSets, or daemonSets.

jvperrin:
Hmmm, fair enough, is there any guarantee the random generation will be in that pattern? I suppose it doesn't matter though and that's a more flexible solution that relying on `deployment` being in the name somewhere.

cg505:
I'm not totally sure, although it would appear so at least at this point.

PR #700, opened 2019-07-09
Automatically install Jenkins plugin updates

jvperrin:
This uses the Jenkins SSH interface to install plugin updates on a regular basis instead of having a human (often me) go and install updates to plugins. It should also email root@ when plugin updates are installed. Maybe in the future some information could be added to the email to show what version the dependencies are too, but I wasn't thinking of including that initially, just for simplicity.

I've set everything up on the Jenkins side, it only consisted of adding user permissions for `jenkins-deploy` and adding a SSH public key.

I also moved the comment about user setup to be next to the `user` resource definition, which I think feels a bit more natural. (it's also in [our docs](https://www.ocf.berkeley.edu/docs/staff/backend/jenkins/#h2_jenkins-security-model) anyway)

### DIFF
--- a/modules/ocf_jenkins/files/update-plugins
+++ b/modules/ocf_jenkins/files/update-plugins
@@ -1,0 +1,13 @@
+#!/bin/bash -eu
+
+ssh_cmd="ssh -p 2222 -o StrictHostKeyChecking=no -i /opt/jenkins/deploy/ssh_cli jenkins-deploy@localhost"
+
+# Taken from https://stackoverflow.com/a/25647793 and modified to use SSH
+# instead of the jenkins-cli jar
+updates=$($ssh_cmd list-plugins | grep -e ')$' | awk '{ print $1 }');
+if [ ! -z "${updates}" ]; then
+    echo "Updating Jenkins Plugins: ${updates}"
+    $ssh_cmd install-plugin ${updates}
+    echo "Restarting Jenkins..."
+    $ssh_cmd safe-restart
+fi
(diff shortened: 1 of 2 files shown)


dkess:
I'm not a huge fan of sending to root every week, since these emails don't necessarily indicate problems. Perhaps sending to rebuild-spam would be better?

jvperrin:
@cg505 Not often, but it does happen. I can't think of any time it's happened in the last couple years that it's actually affected us, but I might also be forgetting sometime it did happen.

@dkessler Well the same is done with off-site backups currently (emailed to root and done weekly) but rebuild-spam would be fine too. One benefit of using email is that it's easily searchable, although if we get an irclogger for that channel then that would work too. My thought was to try this first with email (since it's easy) and potentially switch to something else later if the need arises.

Agreed though, more email is something I'm not a fan of, but I also think this is pretty low noise since it's only done weekly. Maybe the cron stuff etc. that currently goes to root@ should all go to an IRC channel instead? Anyway, I think that's the topic for a different review/change personally.

dkess:
Is there a reason why we're using SSH instead of simply running this procedure in the `jenkins-deploy` cronjob?

jvperrin:
Yes, because it SSHs into Jenkins (see port 2222), not into `reaper`. There's also a `jenkins-cli.jar` that can be downloaded to expose the same functionality, but then credentials are needed other than a private key and I couldn't get it working as reliably as using the SSH cli interface.

PR #702, opened 2019-07-13
Don't run ldap-lint on dev-ldap (it checks prod anyway)

jvperrin:
Maybe we should move this to a different host or just have it check `localhost`? Anyway, we probably don't want this running on `dev-ldap` anyway since we don't really care if that is broken or not, we only care about the consistency of the prod ldap.

### DIFF
--- a/modules/ocf_ldap/manifests/init.pp
+++ b/modules/ocf_ldap/manifests/init.pp
@@ -112,11 +112,13 @@ class ocf_ldap {
     }
   }
 
-  cron { 'ldap-lint':
-    command => '/opt/share/utils/sbin/ldap-lint',
-    user    => root,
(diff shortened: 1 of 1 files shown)


dkess:
I actually think `localhost` would be better here, since if we are experimenting on `dev-ldap`, it would be good to know if we accidentally messed something up.

jvperrin:
Yeah, ok I agree that's a better approach here, I'll work on doing that instead. Maybe could tackle https://github.com/ocf/utils/issues/119 at the same time, idk

jvperrin:
Closing this in favor of #835, which does quite a bit more to `ldap-lint` to make it more useful on `dev-firestorm` instead.

PR #704, opened 2019-07-13
Don't send debconf emails on desktops

jvperrin:
These emails aren't really useful and happen whenever a new desktop is provisioned or video drivers are updated. For instance, this has emailed in the past for the `libcuda1`, `xserver-xorg-video-nvidia`, and `libgl1-nvidia-glx` packages and just warns about a mismatch in video drivers and a need to reboot the desktop, which we do after provisioning them anyway.

I still need to actually test this on a desktop, testing on a newly provisioned one would be the easiest way to figure out if setting this in `/root/.debconfrc` actually works or not, although the manpages say that it should work in that file.

### DIFF
--- a/modules/ocf_desktop/manifests/drivers.pp
+++ b/modules/ocf_desktop/manifests/drivers.pp
@@ -1,6 +1,14 @@
 class ocf_desktop::drivers {
   include ocf::apt::i386
 
+  # Don't send emails from Debconf on configuring desktops since they aren't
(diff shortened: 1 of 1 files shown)


cg505:
This PR is breaking updates on desktops. It seems that by providing `/root/.debconfrc`, the original file `/etc/debconf.conf` is ignored completely, so then debconf complains about missing configuration.

PR #706, opened 2019-07-13
Move desktop files out of Xsession to /etc/skel (fixes #389)

### LINKED ISSUE #389: move desktop files in Xsession
Currently the .desktop files in ~/Desktop are written from the Xsession file directly. It would make a lot more sense to just put them in /etc/skel/Desktop as actual files.

jvperrin:
This should be a decent amount cleaner and easier to maintain, and we don't need the full conditional power that comes with using a script to generate the desktop files since they are all pretty simple. I
simplified the remote file launcher since we assume remote files are working properly in other places anyway (and it's 7-year old code that's not particularly relevant now), and if they didn't, then something else is wrong and we shouldn't just hide the desktop icon.

I also moved the sourcing of `~/remote/.desktoprc` to the very end of the file so that it's loaded last and so anyone can override any settings previously set in the file.

### DIFF
--- a/modules/ocf_desktop/files/skel/Desktop/brave.desktop
+++ b/modules/ocf_desktop/files/skel/Desktop/brave.desktop
@@ -1,0 +1,7 @@
+[Desktop Entry]
+Name=Brave
+StartupNotify=true
+Terminal=false
+Exec=/usr/bin/brave-browser %U
+Type=Application
+Icon=brave-browser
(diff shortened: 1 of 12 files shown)


jvperrin:
Yeah that would be nice, I think it was done like this because otherwise `.config` for example would be a dotfile in this repo and might not get seen, but I'm not actually sure why it was originally organized this way, just a guess.

PR #707, opened 2019-07-22
Revert "Don't send debconf emails on desktops"

cg505:
Reverts ocf/puppet#704

This PR is breaking updates on desktops. It seems that by providing `/root/.debconfrc`, the original file `/etc/debconf.conf` is ignored completely, so then debconf complains about missing configuration.


jvperrin:
What happened?

jvperrin:
Oh I see, just saw the comment on #704, ok

cg505:
On `apt upgrade`, the `debian-security-support` hook fails, complaining `debconf: Config database not specified in config file` (also see update to OP).

PR #708, opened 2019-07-25
mail: Allow forwarding addresses to reply to protected lists

abizer:
Earlier, we only allowed user@ocf.berkeley.edu to reply to mail sent
to staff@ and other protected lists. This prevents staff members
without GApps accounts or mail clients configured to authenticate
againt our SMTP server from replying to messages they receive as a
result of being on staff@ocf.berkeley.edu. This change should allow
staff members to reply to emails from the same addresses that our
systems would send them emails at in the first place.

### DIFF
--- a/modules/ocf_mail/files/site_ocf/update-access
+++ b/modules/ocf_mail/files/site_ocf/update-access
@@ -1,11 +1,22 @@
-#!/bin/bash -eu
-# Update addresses that are allowed to send mail to protected addresses.
+#!/bin/bash
+# Update addresses that are allowed to send to protected addresses. Those
+# being OCF addresses and corresponding LDAP forwarding addresses of staff
+set -euo pipefail
 
-set -o pipefail
+TMP=$(mktemp /tmp/staff-access.XXXXX)
+ACCESS_FILE="/etc/postfix/ocf/staff-access"
 
-# staff
-getent group ocfstaff | cut -d: -f4 | \
-	sed 's/,\|$/@ocf.berkeley.edu OK\n/g' \
-	> /etc/postfix/ocf/staff-access
+# Load the keytab so we can pull the mail attr from LDAP
+kinit -t /etc/postfix/ocf/smtp-krb5.keytab smtp/anthrax.ocf.berkeley.edu
 
-/usr/sbin/postmap /etc/postfix/ocf/staff-access
+staff_members="$(getent group ocfstaff | cut -d: -f4 | tr ',' ' ')"
+
+for user in $staff_members; do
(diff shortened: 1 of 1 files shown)


modules/ocf_mail/files/site_ocf/update-access
```diff
@@ -1,11 +1,21 @@
...
-# staff
-getent group ocfstaff | cut -d: -f4 | \
-	sed 's/,\|$/@ocf.berkeley.edu OK\n/g' \
-	> /etc/postfix/ocf/staff-access
+# Load the keytab so we can pull the mail attr from LDAP
+kinit -t /etc/postfix/ocf/smtp-krb5.keytab smtp/anthrax.ocf.berkeley.edu
+
+USERNAMES="$(getent group ocfstaff | cut -d: -f4 | tr ',' '\n')"
+
+while read -r user; do
```
cg505: imo we should just do `for user in $USERNAMES`. This also lets us `tr` with spaces instead of newlines which I find slightly simpler.

modules/ocf_mail/files/site_ocf/update-access
```diff
@@ -1,11 +1,21 @@
...
+TMP=$(mktemp /tmp/staff-access.XXXXX)
 
-# staff
-getent group ocfstaff | cut -d: -f4 | \
-	sed 's/,\|$/@ocf.berkeley.edu OK\n/g' \
-	> /etc/postfix/ocf/staff-access
+# Load the keytab so we can pull the mail attr from LDAP
+kinit -t /etc/postfix/ocf/smtp-krb5.keytab smtp/anthrax.ocf.berkeley.edu
+
+USERNAMES="$(getent group ocfstaff | cut -d: -f4 | tr ',' '\n')"
```
cg505: Why CAPS on `USERNAMES`?

modules/ocf_mail/files/site_ocf/update-access
```diff
@@ -1,11 +1,21 @@
...
+
+while read -r user; do
+	echo "$user@ocf.berkeley.edu OK" >> "$TMP"
+	attr_mail="$(ldapsearch -LLL uid="$user" mail 2>/dev/null | grep mail | tr -d ' ' | cut -d: -f2)"
```
cg505: Why `2>/dev/null` here? I would think we want to see the failure.
Also, what happens if a staffer were to not have a `mail` attribute?
Finally we should grep for `^mail:` to avoid matching usernames with `mail` as a substring.

abizer: redirecting to stderr to capture the SASL output, but I realize the `-Q` option enables SASL quiet mode

PR #710, opened 2019-07-29
purge /root/.debconfrc from desktops

cg505:
See #704 and #707. This allows us to successfully use `apt` on the desktops again.


fydai:
This can be closed, there are no more .debconfrc on desktops

PR #712, opened 2019-07-30
admin: add user and creds for easyfilters script

dkess:
The easyfilters script will be in utils, but this change adds the necessary secrets file and sudo rules for running it.


cg505:
ref ocf/utils#123

ethanhs:
(see linked PR for resolution)

PR #713, opened 2019-07-30
Backup VM XML from hypervisors (#463)

### LINKED ISSUE #463: back up VM XML
we should start dumping the VM XML files and including them in backups if possible. we can have a scrip that logs in to the hypervisors as ocfbackups, parses the output of `virsh list --all` or some more cli-friendly thing, and then `"ssh -K ocfbackups@$hypervisor virsh dumpxml $domain" > $hypervisor-$domain.xml"` or something like that

jvperrin:
This ends up with a tree like this in `hal:/opt/backups/live/`: https://i.fluffy.cc/gKpD2XcwfrXfpt8b97WqhkkVpMXX75Vx.html

I've been running this for a couple weeks now (at least since July 20th anyway from backup dates), and it seems to be going smoothly.

Closes #463

### DIFF
--- a/modules/ocf_backups/files/rsnapshot.conf
+++ b/modules/ocf_backups/files/rsnapshot.conf
@@ -52,6 +52,12 @@ backup_script	/opt/share/backups/backup-git	git/
 backup_script	/opt/share/backups/backup-pgsql	pgsql/
 
 # remote servers
+backup	ocfbackups@hal:/etc/libvirt/qemu/	servers/vm_xml/hal/
+backup	ocfbackups@jaws:/etc/libvirt/qemu/	servers/vm_xml/jaws/
+backup	ocfbackups@pandemic:/etc/libvirt/qemu/	servers/vm_xml/pandemic/
+backup	ocfbackups@riptide:/etc/libvirt/qemu/	servers/vm_xml/riptide/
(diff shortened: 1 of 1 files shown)


cg505:
Actually can you make this automatically grab everything that is a hypervisor in case we forget to add new hypervisors here?

jvperrin:
I do think that would be good, but after revisiting this, the only way I can think of doing that apart from hardcoding another list in hieradata somewhere (similar to [this](https://github.com/ocf/puppet/blob/master/hieradata/common.yaml#L10-L19)) is to use [exported resources](https://puppet.com/docs/puppet/6.7/lang_exported.html) (each host that uses `ocf_kvm` would export something with its hostname that could be used in this file converted to a template), which adds a dependence on puppetdb. I'm not sure it's worth the complexity honestly for what this does (something simple to back up VM XML). Thoughts?

cg505:
We should have one source of truth for what is and is not a hypervisor, IMO. That said, I'm okay with merging this now and creating an issue.

jvperrin:
I definitely agree, but I think the current source of truth is what includes the `ocf_kvm` module, which doesn't have great introspection unless we use puppetdb in some manner

cg505:
sad :( ok

jvperrin:
Ok, I made #718 for making a single source of truth to make this kind of thing better in the future.

PR #714, opened 2019-07-30
Removing obsolete ensure absent statements from PR 684

jxlin123:
As I mentioned in my last PR, https://github.com/ocf/puppet/pull/684, one of my commits actually ensures certain files are absent in Puppet. Now that the PR is filed away, there is no need to continue ensuring that those files are absent.


jxlin123:
As jvperrin mentioned, this should wait a bit before merging, since not all the changes from PR 684 have been deployed across `ocf_desktop` machines. Some of the machines are on different Puppet environments anyway, so it might be better to check in closer to the beginning of the semester to see if things have settled.

cg505:
The desktops will not wake on their own over summer, so unless someone uses `lab-wakeup -f` and/or ssh-list, puppet won't run. We can probably wait until a bit after the start of the semester.

PR #723, opened 2019-08-15
mirrors: disable FTP support @ OCF

abizer:
almost no one uses FTP anymore and it's causing problems on fallingrocks


cg505:
we should notify projects that list us as supporting FTP first.

abizer:
Sure, merge this after that's done.

chriskuehl:
Also remove ftp from https://github.com/ocf/puppet/blob/2c7e5f4b7236617717b5f974d18544e518573fb6/modules/ocf_mirrors/files/README.html#L18-L21 ?

fydai:
This breaks FreeBSD mirroring, though nobody uses that because we aren't listed as an official mirror. We should decide whether or not we want to keep mirroring FreeBSD or remove ftp.

fydai:
Section 2.4.1 here
https://www.freebsd.org/doc/en_US.ISO8859-1/articles/hubs/article.html

PR #724, opened 2019-08-17
irc: minor changes for buster compatibility

dkess:
This ensures we use Buster packages on Buster (instead of stretch-backports) and disables the AppArmor profile for inspircd.

### DIFF
--- a/modules/ocf_irc/manifests/biboumi.pp
+++ b/modules/ocf_irc/manifests/biboumi.pp
@@ -1,7 +1,13 @@
 class ocf_irc::biboumi {
-  ocf::repackage { 'biboumi':
-    backport_on => ['stretch'],
-    require     => Package['prosody'],
+  if $::lsbdistcodename != 'buster' {
+    ocf::repackage { 'biboumi':
+      backport_on => ['stretch'],
+      require     => Package['prosody'],
+    }
+  } else {
+    package { 'biboumi':
+      require => Package['prosody'],
+    }
   }
 
   service { 'biboumi':
--- a/modules/ocf_irc/manifests/xmpp.pp
+++ b/modules/ocf_irc/manifests/xmpp.pp
@@ -1,19 +1,30 @@
 class ocf_irc::xmpp {
   # When upgrading to buster, these can be installed from the regular buster
   # repo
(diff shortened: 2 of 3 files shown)


dkess:
This is tested on `dev-flood`, after this is merged we can upgrade `flood` to buster!

dkess:
I assume someone will ask why I'm not modifying the AppArmor profile to allow this access. It's definitely possible, and probably not a difficult change, but I am not sure what the best way to do it in Puppet would be. Suggestions welcome.

modules/ocf_irc/manifests/biboumi.pp
```diff
@@ -1,7 +1,13 @@
 class ocf_irc::biboumi {
-  ocf::repackage { 'biboumi':
-    backport_on => ['stretch'],
-    require     => Package['prosody'],
+  if $::lsbdistcodename != 'buster' {
```
jvperrin: I don't think you should need this conditional at all, [`ocf::repackage` should do the right thing here](https://github.com/ocf/puppet/blob/2c7e5f4b7236617717b5f974d18544e518573fb6/modules/ocf/manifests/repackage.pp#L31-L33) and just not install the backported one on buster but still install the backport on stretch. Did it not work?

modules/ocf_irc/manifests/xmpp.pp
```diff
@@ -1,19 +1,30 @@
 class ocf_irc::xmpp {
   # When upgrading to buster, these can be installed from the regular buster
   # repo
-  ocf::repackage {
-    # lua-dbi-mysql is needed to connect to MySQL
-    'lua-dbi-common':
-      backport_on => ['stretch'];
+  if $::lsbdistcodename != 'buster' {
```
jvperrin: Same for all these, I don't think they should be actually needed since `ocf::repackage` should just install the package normally on dists not in the `backport_on` list.

dkess:
Looks like you're right, I should have looked at the `repackage` source code before doing this. I'll get rid of those if statements.

dkess:
Gonna make a new PR since the commit is different enough and I don't want to erase all the history.

PR #733, opened 2019-08-22
Add neofetch to extrapackages.pp

64bitpandas:
Why not


encadyma:
I approve

abizer:
In the future, don't use the Github Web Editor. The convenience is not worth the loss of utility from the developer tooling we have, like pre-commit, even if the change is small.

64bitpandas:
> In the future, don't use the Github Web Editor. The convenience is not worth the loss of utility from the developer tooling we have, like pre-commit, even if the change is small.

Sounds good, thanks for the notice.

PR #734, opened 2019-08-22
use fqdn for ipp

cg505:
Without this, printing was failing with `CUPS backend /usr/lib/cups/backend/ipps returned 4.`

@jxlin123 Thoughts?


jxlin123:
@cg505 Interesting, I never thought this would pop up! Thanks for catching this issue...

To confirm, how did you get error code 5? So far I'm only seeing error code 4, and you've mentioned 4 as well in rebuild...

cg505:
... I think that was just a typo. I updated OP

jxlin123:
Cool, cool, thanks for the changes once again!

We may need to also restart CUPS on all the machines in order to make sure these changes apply, since the currently running instances of CUPS may still be using the old addresses.

Issue #735, opened 2019-08-23
Puppet 6.2 seems to be utterly broken


fydai:
Puppetlabs kindly upgraded the version of puppet in their stretch repository to 6.2. This version adds (virtual packages)[https://github.com/puppetlabs/puppet/commit/f33a531d71a960360c519e38e977b37112fc99b3]
which results in beauty such as the following when running puppet.
```
Info: Package code is virtual, defaulting to libperl5.24
Info: Package mysql-client is virtual, defaulting to mariadb-client-core-10.1
Notice: /Stage[main]/Ocf::Packages::Mysql/Package[mysql-client]/ensure: purged (corrective)
Notice: /Stage[main]/Ocf::Packages::Mysql/Package[mariadb-client]/ensure: created (corrective)
Notice: /Stage[main]/Ocf::Packages::Mysql_server/Package[mariadb-server]/ensure: created (corrective)
Notice: /Stage[main]/Ocf::Packages::Mysql_server/Service[mysql]/enable: enable changed 'true' to 'false' (corrective)
Info: Package ack is virtual, defaulting to dconf-gsettings-backend
Info: Package scala is virtual, defaulting to perl-base
Info: Package texlive is virtual, defaulting to texlive-binaries
Info: Package gnome-icon-theme is virtual, defaulting to adwaita-icon-theme
Info: Package xpdf is virtual, defaulting to poppler-utils
Notice: /Stage[main]/Ocf_desktop::Packages/Package[xpdf]/ensure: purged (corrective)
Notice: /Stage[main]/Ocf::Auth/Package[unscd]/ensure: created (corrective)
Notice: /Stage[main]/Ocf::Packages/Package[avahi-daemon]/ensure: purged (corrective)
Info: Package time is virtual, defaulting to default-jre
Info: Package tree is virtual, defaulting to libpython2.7-stdlib
```
I don't know what vscode has to do with perl, or what ack has to do with dconf-gsettings-backend. Another fun benefit is that cups depends on poppler-utils, so this update to puppet completely breaks the OCF printing setup.

jvperrin:
[...]

PR #736, opened 2019-08-23
Disable puppet virtual packages

fydai:
https://github.com/ocf/puppet/issues/735


cg505:
Let's keep #735 open so that we can actually use virtual packages in the future.

cg505:
oh i already did it

PR #737, opened 2019-08-23
Fix printing SSL

fydai:
Printing is currently broken because ipps suddenly validates the hostname.
We don't have a cert for printhost so this just changes it to printhost.ocf.berkeley.edu


cg505:
Duplicate of #734.

PR #742, opened 2019-08-31
Add NOPASSWD when giving root to individual users/groups

fydai:
Individual users/groups don't have a root credential, so there's no point in not having NOPASSWD.

### DIFF
--- a/modules/ocf/templates/sudoers.erb
+++ b/modules/ocf/templates/sudoers.erb
@@ -16,10 +16,10 @@ ocfdeploy ALL=NOPASSWD: /usr/local/sbin/puppet-trigger
 # User specification
 %ocfroot ALL=(ALL) <%= @nopasswd ? 'NOPASSWD:' : '' %> ALL
 <% @gsudo.each do |group| -%>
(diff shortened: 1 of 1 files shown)


cg505:
let this sit for a bit to see if people have comments

kpengboy:
What is the context behind this PR? I'm not sure why in particular we are planning to give groups root access.

fydai:
Currently there is no way of giving somebody sudo access without also enabling ocf::auth::nopasswd (because of not having a root credential). I would prefer to have a lower barrier of entry to giving out local root, to encourage local root being given out when necessary. With this PR, instead of needing to add both ocf::auth::usudo and ocf::auth::nopasswd, you would only need to add ocf::auth::usudo.

We probably shouldn't ever give out root to groups, but the code is there so I just made the change there anyways.

cg505:
usudo/gsudo should not be used permanently ever, so I think setting up a root principle doesn't make much sense.

kpengboy:
@cg505 We currently give `ocfstaff` the ability to sudo on desktops through `gsudo`. This change widens that privilege to all staff, rather than just those with a /root principal.

cg505:
a965251ebf0cc2cb289b7a7079d647e8eb300ac1

PR #745, opened 2019-09-04
Add size checks to raster-filter

fydai:
convert explodes sizes of some scanned PDFs. This causes ipps to fail when confronted with a 0.5gb pdf. This PR adds some checks to make sure that rasterizing doesn't happen in those cases.

### DIFF
--- a/modules/ocf/files/packages/cups/raster-filter
+++ b/modules/ocf/files/packages/cups/raster-filter
@@ -18,6 +18,14 @@ if [ "${mimetype%%;*}" != "application/pdf" ]; then
     exit
 fi
 
+# Don't rasterize if size >5M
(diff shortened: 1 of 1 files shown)


dkess:
I hope this isn't a permanent change... we've hardcoded the 20 page limit here (which could change in the future) and could potentially be blacklisting legitimate jobs if they're too big.

Isn't the whole point of this new system to move rasterization from the single-threaded, slow print server to the more powerful desktops? If the desktops aren't able to rasterize certain PDFs, we should investigate why instead of blaming the PDFs.

cg505:
This from `pdf-open` and has worked fine for a year (although the number of pages before giving up is lower). Also, it won't fail on big things, it will just give up and pass through the initial input

PR #749, opened 2019-09-05
Fix printing

fydai:
~The only functional change in~ this PR is removing the ocfps filter running on printhost. This seems to fix the issue where printing duplicates would shift slowly upwards and to the right.

There is also an internal refactorings (maybe more to come later), removing tea4cups on the desktops (we have a PPD that we can shove a cupsFilter line into).


modules/ocf/files/packages/cups/raster-filter
```diff
@@ -10,34 +10,52 @@ input=$(mktemp --suffix='.pdf')
...
-    cat "$error" | "/usr/local/bin/convert_failure"
-    cat "$input"
-    exit
+    exec cat "$input"
 fi
 
-# Done rasterizing
-cat "$output"
+# Returns a filename of a PDF to be processed further
+make_pdf() {
```
dkess: Why is this a function if it's just getting called once immediately after being defined?

nikhiljha:
I don't think we have this issue anymore, so closing 😄

PR #751, opened 2019-09-09
Add octocatalog-diff

jvperrin:
This pull request adds [octocatalog-diff](https://github.com/github/octocatalog-diff) (forked at https://github.com/ocf/octocatalog-diff to add some patches for puppet 6.x) for creating a catalog on two different branches (one local, one pulled from a repo git repo) and calculating any differences between the two (or an error if the catalog doesn't even compile). This can be run by doing `make all_diffs` ([here's some sample output](https://i.fluffy.cc/NftRNPh8pl0tVdCQngPq5BbjXZr4N2bC.html), it fails for `singularity` currently because this branch isn't fully rebased), and my idea for it would be that it would run on any pull requests and maybe a version could run on master to just check if catalog compilation works or not.

Right now it's calculating against this branch (see https://github.com/jvperrin/puppet/blob/8c2a9dcf74b3307c0c6fc5d70f5fb01e5e5ee0d2/.octocatalog-diff.cfg.rb#L34-L36), but this should be switched to `origin/master` once this branch is merged.

This also introduces a new resource type, `ocf::privatefile`, that is essentially a wrapper around `file` that will set some parameters automatically (`show_diff` and `backup` set to `false` for instance) and works with `octocatalog-diff` to not use the actual secrets, since it does not have access to them.

There's also dummy secrets added in hiera to facilitate this, which is nice because it shows the structure of the secrets we have in a version-controlled way at least.

One thing I'd still like to do before merging is move [the stuff I added to this file](https://github.com/jvperrin/puppet/blob/8c2a9dcf74b3307c0c6fc5d70f5fb01e5e5ee0d2/modules/ocf_ocfweb/manifests/dev_config.pp#L23-L33) somewhere else, but I think in general this is ready to review, even without that change.

I've also spot-checked this on a few machines (my staff VM `fireball`, `dev-flood`, `supernova`), but to actually test it I need to replace the ENC used on `lightning` so that it returns the `use_private_share` parameter. Maybe that's a sign that I should flip the value the other way around so that by default it tries to use the secrets and otherwise doesn't? Anyway, another small thing I should probably change here.

~Probably also need to squash some commits, there's a bunch here that don't have great descriptions or have been rewritten/rebased.~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+WORKSPACE ?= ${HOME}/.cache
+
 .PHONY: all
 all: vendor install-hooks
 
@@ -18,11 +20,19 @@ install-hooks: venv
 vendor: Puppetfile
 	r10k puppetfile install --verbose --color
 
+.PHONY: octocatalog-diff-uncached
+$(WORKSPACE)/.octocatalog-diff-cache:
+octocatalog-diff-uncached:
+	octocatalog-diff \
+		--bootstrap-then-exit \
+		--bootstrapped-from-dir=$(WORKSPACE)/.octocatalog-diff-cache
+
 # Run octocatalog-diff for a particular hostname
 # Add a --debug flag to get much more verbose output
-diff_%:
+diff_%: $(WORKSPACE)/.octocatalog-diff-cache
 	octocatalog-diff \
 		-n $*.ocf.berkeley.edu \
+		--bootstrapped-to-dir=$(WORKSPACE)/.octocatalog-diff-cache \
 		--enc-override environment=production,parameters::use_private_share=false \
 		--ignore 'Ini_setting[puppet.conf/master/storeconfigs]' \
 		--ignore 'Ini_setting[puppet.conf/master/storeconfigs_backend]' \
@@ -33,8 +43,9 @@ diff_%:
 
 # Run octocatalog-diff across all nodes that can be fetched from puppetdb
 # TODO: Make this faster by just selecting a single node from each class we
-# care about (or not selecting all desktops/hozers for example)
-all_diffs:
+# care about (not selecting all desktops/hozers for example)
+#all_diffs: $(WORKSPACE)/.octocatalog-diff-cache
+all_diffs: octocatalog-diff-uncached
 	curl -s --tlsv1 \
 		--cacert /etc/ocfweb/puppet-certs/puppet-ca.pem \
 		--cert /etc/ocfweb/puppet-certs/puppet-cert.pem \
@@ -43,4 +54,4 @@ all_diffs:
 	| jq -r '.[] | .certname' \
 	| cut -d '.' -f 1 \
 	| sort \
-	| xargs -n 1 -P $(shell grep -c ^processor /proc/cpuinfo) -I @ $(MAKE) -i -s diff_@
+	| xargs -n 1 -P $(shell grep -c ^processor /proc/cpuinfo) -I @ $(MAKE) -s diff_@
--- a/Makefile
+++ b/Makefile
@@ -27,24 +27,36 @@ octocatalog-diff-uncached:
 		--bootstrap-then-exit \
 		--bootstrapped-from-dir=$(WORKSPACE)/.octocatalog-diff-cache
 
-# Run octocatalog-diff for a particular hostname
-# Add a --debug flag to get much more verbose output
+# Run octocatalog-diff for a particular hostname. This is meant to be run by
+# humans, which is why it accepts hostnames instead of the whole FQDN and why
+# all_diffs strips the FQDN down to the hostname again
+.PHONY: diff_%
 diff_%: $(WORKSPACE)/.octocatalog-diff-cache
+	# Add a --debug flag to get much more verbose output
+	#
+	# Ignore a few changes that otherwise cause extra noise:
+	#
+	# Ignore changes to puppetserver settings, since these are dependent on the
+	# path that octocatalog-diff sets and thus always differ
+	#
+	# Ignore changes to SSH keys, since these change when any machines are
+	# reprovisioned, etc. and change frequently
 	octocatalog-diff \
 		-n $*.ocf.berkeley.edu \
 		--bootstrapped-to-dir=$(WORKSPACE)/.octocatalog-diff-cache \
-		--enc-override environment=production,parameters::use_private_share=false \
+		--enc-override environment=production,parameters::dummy_secrets=true \
 		--ignore 'Ini_setting[puppet.conf/master/storeconfigs]' \
 		--ignore 'Ini_setting[puppet.conf/master/storeconfigs_backend]' \
 		--ignore 'Ini_setting[puppetdbserver_urls]' \
 		--ignore 'Ini_setting[soft_write_failure]' \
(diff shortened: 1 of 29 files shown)


jvperrin:
```

**Comment by jvperrin** [comment]
I pushed my branch to this repo at the same branch name (`octocatalog-diff-test`), and https://jenkins.ocf.berkeley.edu/job/puppet/job/PR-751/2/ still failed, but https://jenkins.ocf.berkeley.edu/job/puppet/job/octocatalog-diff-test/1/ essentially succeeded (or would have if I rebased this branch anyway, just `singularity` failed). Maybe the git branches are cached and it reused the workspace so it didn't see new changes? Not sure

**Comment by jvperrin** [comment]
Ok, yeah I believe the [PR build failed](https://jenkins.ocf.berkeley.edu/job/puppet/view/change-requests/job/PR-751/3/) and the [branch build succeeded](https://jenkins.ocf.berkeley.edu/job/puppet/job/octocatalog-diff-test/2/) because of a difference in the remote branches they see. That being said, I'm not sure why the `PR-751` one can't just fetch the branch it needs, that's a bit strange:

```
jvperrin@reaper:/opt/jenkins/slave/workspace/puppet_PR-751$ git remote show origin
* remote origin
  Fetch URL: https://github.com/ocf/puppet.git
  Push  URL: https://github.com/ocf/puppet.git
  HEAD branch: master
  Remote branches:
    master             tracked
    refs/pull/751/head tracked
```
versus:

```
jvperrin@reaper:/opt/jenkins/slave/workspace/puppet_octocatalog-diff-test$ git remote show origin
* remote origin
  Fetch URL: https://github.com/ocf/puppet.git
  Push  URL: https://github.com/ocf/puppet.git
  HEAD branch: master
  Remote branch:
    octocatalog-diff-test tracked
```

Anyway, I think this shouldn't be a problem (at least for PRs) when using `origin/master` instead of `origin/octocatalog-diff-test` as the upstream branch, although it might then cause builds to fail on branches pushed to this repo that aren't pull requests?

File: modules/ocf/manifests/rootpw.pp [hunk 1]
```diff
@@ -4,10 +4,10 @@
 # The root password is a last resort that is only used in rare cases where
 # there is no other option.
 #
-# To regenerate the root password, see /opt/share/utils/staff/puppet/gen-rootpw
+# To regenerate the root password, see the gen-rootpw script from utils
```
cg505: This script needs to be updated (somehow), right?

jvperrin: Yeah, this script wouldn't really work any more. Previously it wrote to [a specific file](https://github.com/ocf/utils/blob/13ff417c39eff56a3eebac8d8e8193b7d2113d9d/staff/puppet/gen-rootpw#L5), but this review changes from using that file to putting the data in hiera. The script could be updated to sed something in hierdata, but I think it's probably just better to print out to the terminal and ask the end user to update it themselves. That script is almost never run anyway (hasn't been run to completion in the last 4 years anyway) and updating hiera is generally pretty easy (update a value in a yaml file in the private share) and anyone updating the root password should already know how to do that.

Makefile
```diff
@@ -16,4 +18,40 @@ install-hooks: venv
...
+# care about (not selecting all desktops/hozers for example)
+#all_diffs: $(WORKSPACE)/.octocatalog-diff-cache
+all_diffs: octocatalog-diff-uncached
+	curl -s --tlsv1 \
+		--cacert /etc/ocfweb/puppet-certs/puppet-ca.pem \
+		--cert /etc/ocfweb/puppet-certs/puppet-cert.pem \
+		--key /etc/ocfweb/puppet-certs/puppet-private.pem \
+		https://puppetdb:8081/pdb/query/v4/nodes \
+	| jq -r '.[] | .certname' \
+	| cut -d '.' -f 1 \
```
dkess: Is this cutting out the `.ocf.berkeley.edu`? Why not just leave it in, since you're adding it again in the `diff_%` rule?

jvperrin: Heh, yeah that'd be better, I just cut it out since I was running it as `make diff_supernova` and wanted to match that, but there isn't really much need to split out just the hostname and then include it again like this

jvperrin: In practice I think doing it this way would be about the same complexity, since I'd like to keep a nice way to run individual diffs per host (the `make diff_supernova` interface for instance). However, this means stripping and re-adding the hostname beforehand unless there's a new make target like `diff_fqdn_%` to take in fqdns and then `diff_%` just passes onto that one (this works too, but looks about the same in terms of complexity to me). Thoughts?

dkess: If you expect humans to actually run that `diff` command then feel free to keep it the way it is-- I just thought this would always be automated, in which case it's better to be explicit. It's not worth adding more Makefile rules for this.

jvperrin: Yeah, I didn't make it clear that it's ready for humans to use, so I'll add a comment about that or explaining why the command is set up like it is and not using the fqdn

modules/ocf/manifests/rootpw.pp
jvperrin: I opened https://github.com/ocf/utils/pull/126 for this, so I'm going to resolve this point.

jvperrin:
Hmmm, I should ignore changes to SSH keys too, otherwise this could cause failures all over the place :/
```
I, [2019-09-11T22:09:14.736942 #14106]  INFO -- : Catalogs compiled for fireball.ocf.berkeley.edu
I, [2019-09-11T22:09:15.261326 #14106]  INFO -- : Diffs computed for fireball.ocf.berkeley.edu
diff origin/octocatalog-diff-test/fireball.ocf.berkeley.edu current/fireball.ocf.berkeley.edu
*******************************************
  Sshkey[autocrat] =>
   parameters =>
     host_aliases =>
      - ["kubernetes.ocf.berkeley.edu", "kubernetes", "kubernetes1.ocf.berkeley.edu", "kubernetes1", "autocrat.ocf.berkeley.edu", "169.229.226.76", "2607:f140:8801::1:76"]
      + ["kubernetes.ocf.berkeley.edu", "kubernetes", "kubernetes1.ocf.berkeley.edu", "kubernetes1", "autocrat.ocf.berkeley.edu", "169.229.226.76", "2607:f140:8801::1:79"]
*******************************************
  Sshkey[coup] =>
   parameters =>
     host_aliases =>
      - ["kubernetes.ocf.berkeley.edu", "kubernetes", "kubernetes2.ocf.berkeley.edu", "kubernetes2", "coup.ocf.berkeley.edu", "169.229.226.77", "2607:f140:8801::1:79"]
      + ["kubernetes.ocf.berkeley.edu", "kubernetes", "kubernetes2.ocf.berkeley.edu", "kubernetes2", "coup.ocf.berkeley.edu", "169.229.226.77", "2607:f140:8801::1:77"]
*******************************************
Makefile:35: recipe for target 'diff_fireball' failed
make: *** [diff_fireball] Error 2
```

jvperrin:
I'm not entirely sure this works currently, as it appears to always be saying there's no diff now, which is strange. However, it doesn't run on master (in Jenkins) and having no diff is ok for the time being (it might work when diffing against master anyway, I'm not sure), so I'm going to merge this as-is and iterate from there.

PR #755, opened 2019-09-10
make xfce4 panel span multiple monitors

ja5087:
Currently, the xfce4 panel only stays on one monitor, which means it only displays on the left half on eruption with it's Picture By Picture mode. On other multi-monitor desktops, this is less important but still the panel isn't shown across both monitors. The default should be spanned in my opinion (and appealing to convention, is the default on macOS and Windows) and if users want old behavior they can toggle it in panel settings.

This PR sets the spanned option by default. Tested on cyclone.

### DIFF
--- a/modules/ocf_desktop/files/xsession/xfce4/xfconf/xfce-perchannel-xml/xfce4-panel.xml
+++ b/modules/ocf_desktop/files/xsession/xfce4/xfconf/xfce-perchannel-xml/xfce4-panel.xml
@@ -29,6 +29,7 @@
         <value type="int" value="18"/>
       </property>
       <property name="autohide" type="bool" value="false"/>
+      <property name="span-monitors" type="bool" value="true"/>
     </property>
     <property name="panel-1" type="empty">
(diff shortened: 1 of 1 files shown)


dkess:
Can you explain what exactly this does? Are you sure it's desirable? I don't remember having issues with xfce panels on dual monitors.

ja5087:
Yep, sorry about that. Updated the desc.

dkess:
If the same panel is spanning multiple monitors, I would say that looks weird and unusual on setups that aren't eruption. I would be indifferent if this configuration option were only on eruption, but personally I vote no if this is gonna affect all multi-monitor desktops.

If agree with you that this is a good change, feel free to merge over my objections.

jvperrin:
I feel like a screenshot would be the best to show what this looks like (that or a picture from a camera so you can see the monitor splits)

ja5087:
On Eruption it just looks like it spans the whole monitor. Here are some pictures from Typhoon. 

Here's what it looks like before:


Here's what it will look like after

PR #759, opened 2019-09-16
Add a bingo shorturl

fydai:


ethanhs:
@abizer The problem with /s/ is that it makes being op on IRC a dependency to adding links. I don't see why using regular shorturls is an issue. I think /s/ should get a non-IRC interface before being recommended.

abizer:
using IRC should hardly be an issue for an SM. We want to minimize permanent shorturl use because it still has the unfortunate requirement of us having to maintain a manual synchronization between shorturls and blacklisted usernames.

Baisang:
> I think /s/ should get a non-IRC interface before being recommended.

Do we have a issue/project for creating some sort of web interface for adding short URLs? seems like this would help address issues with people not having op (although I think in the meantime it'd be easier to just grant op)

ethanhs:
>using IRC should hardly be an issue for an SM

I don't disagree with that, but that also isn't my point. My point is that IRC is not a UI, nor should it be shoe-horned into being one. I'd prefer /s/ to be usable as a script, or via ocfweb.

PR #761, opened 2019-09-20
shorturl: add startertasks

dkess:


dkess:
This only removes one item from the username space, the dash doesn't affect it

PR #762, opened 2019-09-20
Post octocatalog-diff results as pull request comments

jvperrin:
This sets octocatalog-diff to run on pull requests and post the results as comments back to the review about what hosts are unaffected, which have diffs, and which errored in the catalog compilation stage. It also groups diffs that are the same together so that it's less verbose than listing every host in the output, but doesn't do the same with errors (after all, they should all be fixed anyway before proceeding). A few examples are shown below in this review

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,3 +1,5 @@
+@Library('shared-ocf-pipeline@add-create-gist') _
+
 pipeline {
   agent {
     label 'slave'
@@ -49,13 +51,13 @@ pipeline {
             // https://issues.jenkins-ci.org/browse/JENKINS-44930), so the
             // output is saved to a file and then used soon after
             def status = sh returnStatus: true, script: 'make all_diffs > all_diffs_output.log'
-            // GitHub has a max comment length of 65536, so truncate and leave
-            // a warning about it at the top if necessary
+            // GitHub has a max comment length of 65536, so create a gist and
+            // link to that if necessary
             def output = readFile('all_diffs_output.log').trim()
             def charLimit = 65536
             if (output.length() > charLimit) {
-              def warning = "**WARNING: Output has been truncated due to comment limit, see Jenkins for full output**\n"
-              pullRequest.comment(warning + output.take(charLimit - warning.length()))
+              def url = createGist(output)
+              pullRequest.comment('**WARNING: Test output is too long for a GitHub comment, posted to a gist instead**: ' + url)
             } else {
               pullRequest.comment(output)
             }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,15 +29,6 @@ pipeline {
     }
 
     stage('octocatalog-diff') {
-      // Don't run this on the master branch yet, since it's really made for
-      // testing PRs and changes, it should always show no diffs on master.
-      // However, it might be useful on master in the future in some kind of
-      // mode to just show that all catalogs actually compile.
-      when {
-        not {
-          branch 'master'
-        }
-      }
       steps {
         // Fetch in the master branch so that octocatalog-diff can diff against
         // it. Jenkins by default only clones in branches that are needed and
@@ -49,16 +40,21 @@ pipeline {
         sh 'git config --add remote.origin.fetch +refs/heads/master:refs/remotes/origin/master'
         sh 'git fetch --no-tags'
 
-        // Don't fail the whole build if octocatalog-diff fails, since it's new
-        // and needs some fixing before it's relied on
         script {
-          try {
-            sh 'make all_diffs'
-          } catch (err) {
-            echo 'make all_diffs failed, but it is being ignored for now'
-            mail to: 'jvperrin@ocf.berkeley.edu',
-                 subject: "all_diffs failed on ${JOB_NAME}/#${BUILD_NUMBER}",
-                 body: BUILD_URL
+          // This should only run for pull requests, so that it is able to post
+          // change/failure comments on the review
+          if (env.CHANGE_ID) {
+            // Don't fail the whole build if octocatalog-diff fails, since it's new
+            // and needs some fixing before it's relied on
+            try {
+              //sh 'make all_diffs'
+              pullRequest.comment('Test comment from jenkins')
+            } catch (err) {
+              echo 'make all_diffs failed, but it is being ignored for now'
+              mail to: 'jvperrin@ocf.berkeley.edu',
+                   subject: "all_diffs failed on ${JOB_NAME}/#${BUILD_NUMBER}",
+                   body: BUILD_URL
+            }
           }
         }
       }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -50,10 +50,10 @@ pipeline {
             // saved at the same time (thanks jenkins, see
             // https://issues.jenkins-ci.org/browse/JENKINS-44930), so the
             // output is saved to a file and then used soon after
-            def status = sh returnStatus: true, script: 'make all_diffs > all_diffs_output.log'
+            def status = sh returnStatus: true, script: './bin/octocatalog-diff > all_diffs_output.md'
             // GitHub has a max comment length of 65536, so create a gist and
             // link to that if necessary
-            def output = readFile('all_diffs_output.log').trim()
+            def output = readFile('all_diffs_output.md').trim()
             def charLimit = 65536
             if (output.length() > charLimit) {
               def url = createGist('octocatalog-diff-results.md', output)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,7 +49,16 @@ pipeline {
             // https://issues.jenkins-ci.org/browse/JENKINS-44930), so the
             // output is saved to a file and then used soon after
             def status = sh returnStatus: true, script: 'make all_diffs > all_diffs_output.log'
-            pullRequest.comment(readFile('all_diffs_output.log').trim())
+            // GitHub has a max comment length of 65536, so truncate and leave
+            // a warning about it at the top if necessary
+            def output = readFile('all_diffs_output.log').trim()
+            def charLimit = 65536
+            if (output.length() > charLimit) {
+              def warning = "**WARNING: Output has been truncated due to comment limit, see Jenkins for full output**\n"
+              pullRequest.comment(warning + output.take(charLimit - warning.length()))
+            } else {
+              pullRequest.comment(output)
+            }
 
             if (status != 0) {
               currentBuild.result = 'FAILURE'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -56,7 +56,7 @@ pipeline {
             def output = readFile('all_diffs_output.log').trim()
             def charLimit = 65536
             if (output.length() > charLimit) {
-              def url = createGist('octocatalog-diff-results.log', output)
+              def url = createGist('octocatalog-diff-results.md', output)
               pullRequest.comment('**WARNING: Test output is too long for a GitHub comment, posted to a gist instead**: ' + url)
             } else {
               pullRequest.comment(output)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -56,7 +56,7 @@ pipeline {
             def output = readFile('all_diffs_output.log').trim()
             def charLimit = 65536
             if (output.length() > charLimit) {
-              def url = createGist(output)
+              def url = createGist('octocatalog-diff-results.log', output)
               pullRequest.comment('**WARNING: Test output is too long for a GitHub comment, posted to a gist instead**: ' + url)
             } else {
               pullRequest.comment(output)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,3 @@
-@Library('shared-ocf-pipeline@add-create-gist') _
-
 pipeline {
   agent {
     label 'slave'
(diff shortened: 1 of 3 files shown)


jvperrin:
Yeah, so this didn't work, but it's not blocking anything. I'm going to sleep on it and see if I can figure out what's wrong tomorrow.

jvperrin:
Ok, I figured out this is something to do with the caching, as removing [the `--bootstrapped-to-dir` argument](https://github.com/ocf/puppet/blob/3413cbf9d7f4e2ded424f47c9874bbc51f399d34/Makefile#L46) from the command does make it show a diff again, however I'm not sure quite yet what is actually wrong with the caching behavior. I'm also thinking that moving this into a script in `bin` would be much better than putting all this logic in the `Makefile` and trying to get the caching behavior just right.

jvperrin:
I'll keep this open until #778 is merged (assuming it is), and then re-run the PR on this and see what it shows, but if all goes well I think I can just close this soon.

jvperrin:
Well, a bit verbose, but that's essentially what I'm going for

ethanhs:
@jvperrin, would it be possible to have the message in a collapsable thing like https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab ?

jvperrin:
@ethanhs Yeah that's what it's in currently, but nesting the collapsible sections or something might be nicer. I'm also looking to deduplicate the diffs, so essentially grouping by hosts that have all the same diff to make the message easier to read. Not planning on doing the same for errors though (which currently aren't reported in a comment because the build step fails when there's an error but then doesn't continue to the end of the step unfortunately)

ethanhs:
@jvperrin oh sorry, my mobile github client doesn't do collapsible sections >.>

>I'm also looking to deduplicate the diffs, so essentially grouping by hosts that have all the same diff to make the message easier to read

This would be really neat! Thank you so much for working on this

cg505:
@jvperrin is this still being worked on?

jvperrin:
@cg505 So I haven't worked on this in the last couple months, but given the amount of work I've put into it so far I'd still like to see it across the finish line of having some usefulness for current staff. I'd say leave this open and I'll try and get to it this weekend or sometime soon (and if I don't then it's fine to close).

cg505:
I think it would be good for someone to pick this up if you can't, so if it looks like you won't be able to finish it, maybe leave some message with info about the current status. If you will be able to finish it, awesome!

jvperrin:
@ethanhs @cg505 (and anyone else who is interested), I believe this is now ready for review and I've tested it enough on this pull request to show its capabilities.

bin/octocatalog-diff
```diff
@@ -47,46 +67,84 @@ def get_hosts_from_puppetdb():
...
+        return (host, 0, stdout)
     else:
         # This means the puppet run has failed in some way (a return code of 1
-        # is pretty usualfor that), so the whole command should error out
-        print('run for {} errored:'.format(host))
-        print(process.stderr.decode('utf-8').rstrip())
-        return 1
+        # is pretty usual for that), so we return an error code and the output
+        # to post in the review
+        return (host, 1, GITHUB_COLLAPSE_CODE.format('**error** for {}'.format(host), 'text', stderr))
```
cg505: does it make sense to include the stdout here as well?

bin/octocatalog-diff
```diff
@@ -47,46 +67,84 @@ def get_hosts_from_puppetdb():
...
+    host, status, cmd_output = output
+    return_code = max(return_code, status)
+
+    if status and cmd_output:
+        errors.append(cmd_output)
+    elif cmd_output:
+        output_without_host = cmd_output.split('\n', 1)[1]
+        diffs[output_without_host] = diffs.get(output_without_host, []) + [host]
+    else:
+        empty_diffs.append('`{}`\n'.format(host))
```
cg505: would be nice to put these in one big code block instead of individual inline code tags

bin/octocatalog-diff
```diff
@@ -95,17 +153,57 @@ def all_diffs():
...
+
     # Get number of cores to parallelize across and create a worker pool of that size
     pool = Pool(os.cpu_count())
 
-    print('Running octocatalog-diff across all hosts...')
-    # Find any status codes that are non-zero (errors encountered in puppet runs)
-    returncode = max([ret for ret in pool.imap(diff, hosts)])
-    return returncode
+    info('Running octocatalog-diff across all hosts...')
+    run_output = pool.imap(diff, hosts)
```
cg505: `imap_unordered`?

ethanhs: It is nice that the hosts are ordered alphabetically e.g. for looking for the output of a particular host.

Jenkinsfile
```diff
@@ -49,16 +40,29 @@ pipeline {
...
-                 body: BUILD_URL
+          // This should only run for pull requests, so that it is able to post
+          // change/failure comments on the review
+          if (env.CHANGE_ID) {
+            // Unfortunately both the output and the status code cannot be
+            // saved at the same time (thanks jenkins, see
+            // https://issues.jenkins-ci.org/browse/JENKINS-44930), so the
+            // output is saved to a file and then used soon after
+            def status = sh returnStatus: true, script: 'make all_diffs > all_diffs_output.log'
+            // GitHub has a max comment length of 65536, so truncate and leave
```
ethanhs: perhaps we could post extra long diffs to fluffy or something? Or perhaps a github gist? (Note I don't think this has to be done immediately, just a follow up idea).

bin/octocatalog-diff
```diff
@@ -47,46 +67,84 @@ def get_hosts_from_puppetdb():
...
-            '--ignore', "'Ini_setting[puppet.conf/master/storeconfigs_backend]'",
-            '--ignore', "'Ini_setting[puppetdbserver_urls]'",
-            '--ignore', "'Ini_setting[soft_write_failure]'",
-            '--ignore', "'File[/tmp/*/routes.yaml]'",
-            # Ignore changes to SSH keys, since these change when any machines
-            # are reprovisioned, etc. and change frequently
-            '--ignore', "'Sshkey[*]'",
+        ' '.join(
+            (
+                'octocatalog-diff',
```
ethanhs: You can (in fact should maybe) pass a list of arguments to `subprocess.run`.

Jenkinsfile
```diff
@@ -49,16 +40,29 @@ pipeline {
...
-          try {
-            sh 'make all_diffs'
-          } catch (err) {
-            echo 'make all_diffs failed, but it is being ignored for now'
-            mail to: 'jvperrin@ocf.berkeley.edu',
-                 subject: "all_diffs failed on ${JOB_NAME}/#${BUILD_NUMBER}",
-                 body: BUILD_URL
+          // This should only run for pull requests, so that it is able to post
+          // change/failure comments on the review
+          if (env.CHANGE_ID) {
```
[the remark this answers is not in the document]
jvperrin: I believe so, but I haven't tested this explicitly since I've only used this branch so far for testing. From https://jenkins.io/doc/book/pipeline/multibranch/#additional-environment-variables it appears to be the pull request ID, so it should exist for non-fork-based PRs too.

bin/octocatalog-diff
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/python3 -u
```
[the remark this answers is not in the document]
jvperrin: Ah, so this was to get unbuffered output since I think I was getting stdout and stderr mixed together (it's been a little while now though so I don't exactly remember). However, given how it's used now, I don't think this is actually necessary for use in Jenkins.

bin/octocatalog-diff
jvperrin: I don't think the stdout has anything particularly useful when errors like this happen (it's blank for instance for all the errors on [build 15](https://jenkins.ocf.berkeley.edu/blue/rest/organizations/jenkins/pipelines/puppet/branches/PR-762/runs/15/nodes/42/steps/47/log/?start=0) where I added a test error), and it is printed shortly above this (line 108), so it'll still be in the jenkins log itself.

bin/octocatalog-diff
jvperrin: Yeah, agreed on that, I'll change this 👍

bin/octocatalog-diff
jvperrin: Yeah, that's why I kept it ordered, having it alphabetical is pretty nice. I'm not sure otherwise what the difference is between the ordered and unordered versions or if `imap` ensures that the results are ordered, but it seems to be working reasonably well right now to return the hosts in sorted order by hostname so I haven't investigated any further.

Jenkinsfile
jvperrin: Hmmm, I do like that, it would be nicer than telling people to just go look at the jenkins log...

I especially like the idea of making a gist out of the results, as then it would have the same collapsable sections, where a fluffy link would not unless each error or diff is uploaded separately or something and then linked.

bin/octocatalog-diff
jvperrin: So I initially used that, but that kept giving junk diffs for `lightning.ocf.berkeley.edu` where it would still include things like `Ini_setting[puppet.conf/master/storeconfigs]` that are meant to be ignored. I couldn't reproduce the same issue when running the command locally in a shell, and switching this to use a shell also made it work (see the comment on lines 99-101). I'm not a fan of using this either where not having a shell would be better, but I also haven't found a better workaround or dug deep enough into what differs between `octocatalog-diff` when it has a shell versus when it does not to figure this one out, so I'm inclined to leave it for the time being and add a TODO or similar note about switching this in the future if a workaround is found.

Jenkinsfile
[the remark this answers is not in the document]
jvperrin: I believe so, but I haven't tested this explicitly since I've only used this branch so far for testing. From https://jenkins.io/doc/book/pipeline/multibranch/#additional-environment-variables it appears to be the pull request ID, so it should exist for non-fork-based PRs too.

bin/octocatalog-diff
ethanhs: You can use shell=True and pass a list (though the Python docs are not very clear here). Unless you mean that passing a string vs a list made a difference?

bin/octocatalog-diff
ethanhs: imap_unordered will usually run faster, but I don't think that is worth it here, and it seems silly to call imap_unordered and then sorted() afterwards.

bin/octocatalog-diff
cg505: >If shell is `True`, it is recommended to pass args as a string rather than as a sequence.
and later
> That is to say, [when `shell=True`,] `Popen` does the equivalent of `Popen(['/bin/sh', '-c', args[0], args[1], ...])`

From the python3 docs on `subprocess.Popen`. So, passing a list will not work.

Jenkinsfile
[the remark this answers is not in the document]
cg505: I think that for non-fork PRs our Jenkins does not run PR tests - it just does the standard branch tests. (Compare Jenkins runs for #821 and #822.) Maybe we should change that?

bin/octocatalog-diff
ethanhs: Ah I missed that part, yeah this is fine then.

Jenkinsfile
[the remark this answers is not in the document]
jvperrin: Ah, I see what you mean, interesting. I'm not sure how to make the branches use the same workflow as pull requests, but the jenkins config for `puppet` does have an option for "Only branches that are also filed as PRs" instead of the current setting of "All branches". I've tried switching it to that and we'll see if that makes a difference or not.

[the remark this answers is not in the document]
jvperrin: Yeah I don't think it makes a difference apart from filtering which branches have tests run, but maybe they'll have the right environment variable set then, that part still need to be tested.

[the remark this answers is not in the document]
jvperrin: Oh never mind there's a "Discover pull requests from origin" option that makes a lot more sense, I'll try that instead.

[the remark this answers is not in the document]
jvperrin: Yeah, this does what we want I believe, so branches that don't have a PR will build and skip this step, but branches that do have a PR or ones that are from a fork should use this workflow and post a comment.

Jenkinsfile
jvperrin: Ok, I changed this to post to a gist instead, and opened https://github.com/ocf/shared-pipeline/pull/7 to add that functionality to jenkins.

Makefile
```diff
@@ -22,4 +22,4 @@ vendor: Puppetfile
 
 .PHONY: all_diffs
 all_diffs:
-	./bin/octocatalog-diff
+	@./bin/octocatalog-diff
```
ja5087: Now that we're not using `make` for the output should we still add the `@` directive?

ethanhs:
I will say I looked at this PR on mobile, through FastHub, and it doesn't collapse the markdown (which in fairness is a problem on their end).

I was thinking regardless that it may just be better to always post to a gist? Could it be something like:

### Errored hosts (46)

### Unaffected hosts (71)

#### Full output: {url}

Then we have a summary of changes (for a quick sanity check) and it keeps the full information out of the comment.

cg505:
This is basically what it looks like on mobile GitHub web, and I think we would be kind of optimizing around FastHub's bug and making everyone else's life more complicated.

jvperrin:
I do like the idea of posting the summary view here even if a gist is used (number of hosts in each category for instance) I was thinking of doing that yesterday but didn't actually end up doing so since it would need some more refactoring. I'll try and get that working today.

Since it would be pretty easy to switch to using a gist for _every_ comment instead of just some (a small change to the `Jenkinsfile` essentially), I'm going to leave it as-is for now and if the output gets to be too annoying or it's too annoying to use in FastHub, then it can be easily changed in the future after this has been used in some actual reviews.

Makefile
jvperrin: Yeah, good point, I'll change this back since this is really only here for people to run manually if they would like to.

ethanhs:
Sounds good, thank you again for working on this!

jvperrin:
I don't think I like the UI of the comment quite as much now, but it does make for a convenient summary at the top even if the comment is too long and needs to be put into a gist. Maybe putting every run into a gist would be better, I'm not sure, but I do think this can be tweaked later too once it's been tested a bit more. Anyway, I think this is good enough for now (and has enough support/feedback/testing) so I'll go ahead and merge it!

PR #763, opened 2019-09-20
Change pages -> sides in paper-genmon

fydai:
opstaff mentioned it was confusing


ethanhs:
@keur, could you explain why you think so?

keur:
@ethanhs it's going to say "10 sides left today". What is a side? It reads weird and probably will confuse people.

ethanhs:
@keur, ah I can see that. Perhaps "10 sides of paper"?

fydai:
Would you like to propose a better alternative?

Issue #744, opened 2019-09-03
Make it easier to access Downloads folder


cg505:
Someone in lab today was confused about how to access the Downloads folder from the desktop. We should consider making an icon on the Desktop, or maybe just putting it in the "favorites" sidebar in thunar.

PR #764, opened 2019-09-22
Add Downloads to GTK bookmarks

zpfeiffer:
As per #744


jvperrin:
Yeah, I'll have a look at that, not sure why there's a difference at all and it shouldn't fail the build on that anyway (should just comment the difference to the PR if it's a simple difference like this)

PR #765, opened 2019-09-22
fixed tab-autocomplete with username

chengxil:
prevents bash/zsh to load up 60k+ user entries when doing tab-autocomplete


jvperrin:
I re-ran the PR build, but it might fail due to some recent octocatalog-diff changes, see #751. Just wanted to see what it does and since this doesn't touch any puppet manifests or anything it wouldn't catch any things there anyway, so the result doesn't matter much (the pre-commit checks are useful though)

dkess:
bless tf up. this is game-changing

PR #767, opened 2019-09-22
Fix paper-genmon in Debian 10

zpfeiffer:
On Debian 10 devices the generic monitor on the top xfce panel that displays the user's printer quota currently fails to display.
This appears to be because setting the panel item's font to `(default)` is no longer supported. Setting a specific font (`Sans 10`) fixes this.


zpfeiffer:
The current font size of `Sans 10` will look the same on Debian 9 but will be too small on Debian 10, so it should be changed to `Sans 13` at some point.

nikhiljha:
I don't think we have this issue anymore, so closing 😄

PR #768, opened 2019-09-22
Restore indicator panel on Debian 10 lock screen

zpfeiffer:
Debian 10 breaks the indicator bar on the greeter/lock screen, this should fix that. This works on Debian 10, but I have not tested it on Debian 9.


dkess:
Haven't tested this, but maybe you can leave both options in until we completely move to buster?

Anyways, tyty for this!

PR #770, opened 2019-09-24
Add zerotier and syncthing to panic

ethanhs:


ethanhs:
>nice copy paste xdxd

Why learn how things work when you can just copy paste :thinking:

PR #774, opened 2019-09-28
Add texlive back to extrapackages

jameslzhu:
15516bc broke hkn's latex resume book generation, so this commit adds texlive back for use on apphost.


dkess:
These shouldn't be in both files, can't we make this a proper `git revert` of the problematic commit?

jameslzhu:
@dkess Oh my bad, I assumed the desktop puppet config was mutually exclusive. I'll do a git revert instead.

abizer:
we don't need to do this, just install the package manually

jameslzhu:
@dkess @abizer updated pr to revert 15516bc.

nattofriends:
Why would you install a package manually in 2019

jameslzhu:
Updating extrapackages means installing texlive on all machines with the config, do we have an apphost-specific manifest?

PR #778, opened 2019-10-08
Move octocatalog-diff from the Makefile to a separate script

jvperrin:
This allows for the workflow to be more easily customized as well as some nicer stuff around managing the output of the commands, since we don't really care if the return code is 0 as that means there's no diff, but if it's a 1 (errored) or 2 (there's a diff) then the output is more important to surface.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,50 +20,6 @@ install-hooks: venv
 vendor: Puppetfile
 	r10k puppetfile install --verbose --color
 
-.PHONY: octocatalog-diff-uncached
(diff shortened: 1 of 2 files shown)


jvperrin:
https://jenkins.ocf.berkeley.edu/job/puppet/job/PR-778/1/console shows generally what the output looks like if there's no changes, specifically https://i.fluffy.cc/PNfZ91cNWwtZHcwD26Cjw1tRsrh8hGnj.html. I'm not sure why it buffered the output like that though, but I suppose for a test run it's not a big deal.

PR #779, opened 2019-10-09
Store ED25519 SSH host keys in Puppet

kpengboy:
Right now, ED25519 keys are generated when sshd is installed, but they are not stored in Puppet, so when the host is rebuilt, the ED25519 keys might change, causing host verification failed warnings for clients which chose to use that key.
Store the keys in Puppet to prevent these kinds of problems.


kpengboy:
I did move them into the private store. Thanks for reviewing.

PR #781, opened 2019-10-10
Fix kerberos not booting up correctly [needs testing]

fydai:
The issue was that kerberos need to be started after the ldap server started, so I modified the systemd unit file to ensure this ordering.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,15 +29,6 @@ pipeline {
     }
 
     stage('octocatalog-diff') {
-      // Don't run this on the master branch yet, since it's really made for
-      // testing PRs and changes, it should always show no diffs on master.
-      // However, it might be useful on master in the future in some kind of
-      // mode to just show that all catalogs actually compile.
-      when {
-        not {
-          branch 'master'
-        }
-      }
       steps {
         // Fetch in the master branch so that octocatalog-diff can diff against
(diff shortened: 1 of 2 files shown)


dkess:
I'd be surprised if this were the problem, since I expect these two Debian packages to have an established and well-tested interaction. But if it works it works 🤷‍♂️

cg505:
stuff still didn't seem to work right with this change

cg505:
wait... was this change present when firestorm was booted from the power outage on Friday?

fydai:
No, the change was not present, on Friday, yes, I did test this when everything was down and it let kerberos start up correctly.

fydai:
I will test this the next time we reboot the hypervisors.

ja5087:
@fydai was this still an issue when we did the unplanned hypervisor rebooting exercise on Sunday? If so we should merge this.

nikhiljha:
@ethanhs If we didn't run into this on Saturday please close this :)

ethanhs:
We had to restart kerberos, but we didn't get to test this. When we reboot riptide next we can add it to the list

ethanhs:
Finally tested this today and it works!

ethanhs:
Should probably rebase on master before merge tho

ethanhs:
Rebased on master in #1206

PR #786, opened 2019-10-16
update entry for bjb

ja5087:
after [https://github.com/ocf/dns/pull/48](https://github.com/ocf/dns/pull/48) is merged this is safe to merge


ja5087:
thanks for that, I might fix up the guide for this in time

PR #788, opened 2019-10-18
Update python3-ldap3 with forced newer version

axmmisaka:
This is to resolve https://github.com/ocf/utils/issues/128 once, but probably not for all for sure; didn't test if version can be correctly fetched but hopefully

### DIFF
--- a/modules/ocf/manifests/packages.pp
+++ b/modules/ocf/manifests/packages.pp
@@ -147,6 +147,9 @@ class ocf::packages {
     backport_on => ['stretch'],
   }
 
+  ocf::repackage { 'python3-ldap3':
+    backport_on => ['stretch'],
+  }
 
   # Packages to only install on Debian (not on Raspbian for example)
   if $::lsbdistid == 'Debian' {


jvperrin:
Also, try running `make test` locally, it'll run the pre-commit tests and show you if it passes or not even before you push (and if you run `make install-hooks` it'll install the pre-commit hooks so any future commits you make should automatically run the checks too).

jvperrin:
Actually, it looks like the [`ensure` parameter doesn't even exist for `ocf::repackage`](https://github.com/ocf/puppet/blob/25b447286e2807f46ce697c53923937c7fe77cdb/modules/ocf/manifests/repackage.pp#L1-L6), so yeah including that definitely won't work. From a desktop (vortex) for instance:

```
Error: Could not call 'find' on 'catalog': Evaluation Error: Error while evaluating a Resource Statement, Ocf::Repackage[python3-ldap3]: has no parameter named 'ensure' (file: /tmp/ocd-ipc-20191018-27124-571ae1/ocd-builddir-20191018-27258-7illv1/environments/production/modules/ocf/manifests/packages.pp, line: 150) on node vortex.ocf.berkeley.edu
```

axmmisaka:
Got it, I’ll fix asap - btw, how do I test puppet without sudo

jvperrin:
That's a tough one, in general you can't test easily (like really test against a host) without sudo, but that's part of why I've been adding `octocatalog-diff` to this project. It's currently running on each build, so you should see it in the build results (https://jenkins.ocf.berkeley.edu/blue/organizations/jenkins/puppet/detail/PR-788/2/pipeline/42 for instance for the latest one on this PR), which then shows if it has any errors and if the changes in this review made any diffs on hosts.

axmmisaka:
BTW, on cyclone `apt --simulate install python3-ldap3` will still install ldap3 1.2 unless I specify the exact version 2.1, do you think using backport can resolve this issue?

jvperrin:
Yeah, that command would install the current version, but [the way `ocf::repackage` works when given the backport option](https://github.com/ocf/puppet/blob/3c7a18026720d28d23bc2046b0b5e359cad0f196/modules/ocf/manifests/repackage.pp#L21) then it'll run something more like `apt -t stretch-backports --simulate install python3-ldap3` which _will_ install the newer version (since it's installing from backports in the command).

axmmisaka:
Sorry folks, I’m too busy recently, it’s now okay

PR #789, opened 2019-10-19
Enable CJK IMEs on the desktops by default

ja5087:
Since this is in `skel` it won't impact users who already have their profile on the desktops, but all new users will have the CJK IMEs enabled by default. This covers the most common reasons people need to switch their keyboard layout, and the default is still English.


dkess:
How was this file generated? Why does it have commented out options?

ja5087:
It's just the profile file located in ~/.config/fcitx after you enable these input methods.

ethanhs:
In general I am against extraneous commented out configs/code/etc. Is there a case where we would want to edit in the commented out options? Is it hard to find docs for this? If the answer to both of those questions is yes then I'm okay with leaving them in. Otherwise I think we should remove them.

ja5087:
So I deleted out the commented out code but I'm not sure how I'd split the lines

fydai:
That's fine, feel free to squash and merge yourself

PR #791, opened 2019-10-24
Kb worker

Vio-Eli:
added Bedbugs to Kubernetes_Worker


ethanhs:
I noticed rootspam like
>Cannot create /opt/share/docker/sourcegraph; parent directory /opt/share/docker does not exist

I think we need to make this a hypervisor or something as well?

PR #793, opened 2019-10-26
add genfile resource definition

dkess:
`ocf::genfile` codifies a somewhat common in the codebase where we run a command to generate a file and thereafter add a cron rule to continuously run this command. This commit adds ocf::genfile and uses it in the necessary manifests.

I promised to do this a year ago in https://github.com/ocf/puppet/pull/386#discussion_r225031819 and am finally following through on the promise. Happy Hacktoberfest :P

This is tested to work everywhere it's used, it's essentially a no-op for the actual resource tree.

### DIFF
--- a/modules/ocf_www/manifests/site/vhosts.pp
+++ b/modules/ocf_www/manifests/site/vhosts.pp
@@ -7,7 +7,7 @@ class ocf_www::site::vhosts {
         Package['python3-ocflib', 'python3-jinja2'],
         File['/opt/share/vhost-web.jinja'],
       ],
-      notify  => Exec['build-vhosts'];
+      notify  => Ocf::Exec_And_Cron['build-vhosts'];
 
     '/opt/share/vhost-web.jinja':
       source  => 'puppet:///modules/ocf_www/vhost-web.jinja';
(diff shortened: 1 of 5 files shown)


dkess:
Sure, I'd be willing to rename it. exec_and_cron seems a bit too verbose for me but if no one can think of anything better then I'll use it.

dkess:
Renamed to `exec_and_cron`.

modules/ocf_www/manifests/site/vhosts.pp
```diff
@@ -36,6 +31,6 @@
       # (The first vhost declared is the fallback vhost.)
       Include /etc/apache2/ocf-vhost.conf
     ",
-    require => Exec['build-vhosts'],
+    require => Ocf::Genfile['build-vhosts'],
```
fydai: You missed one here

modules/ocf_www/manifests/site/vhosts.pp
```diff
@@ -7,7 +7,7 @@
         Package['python3-ocflib', 'python3-jinja2'],
         File['/opt/share/vhost-web.jinja'],
       ],
-      notify  => Exec['build-vhosts'];
+      notify  => Ocf::Genfile['build-vhosts'];
```
fydai: and here

dkess: ahh shame on me for not testing :<

dkess:
Should be all cleaned up now.

PR #800, opened 2019-11-03
prometheus: change DiskWillFillIn5Hours to 3h

dkess:
The 5-hour alert was too noisy, 3 hours should be a lot better.


dkess:
ty

PR #803, opened 2019-11-08
Move redis SSL terminator from hitch to haproxy

fydai:
Until some version of redis in debian supports native SSL.

_Silently prays that this fixes the celery timeout issues_


fydai:
The issue yesterday was using "localhost" rather than "127.0.0.1" in the haproxy forward settings. Fixed, tested, and prayers given.

PR #805, opened 2019-11-09
Added New Redis Printer Channel & Changed Print Notification Channel

snarain01:
I added a new channel to redis for each of the printers. Each printer channel would contain a list of users who used that printer. This would allow me to subscribe to these channels from a raspberry pi which would be connected to a monitor. The monitor would be placed above the printer area of OCF and would allow members to see which printer printed their job. I also changed the current job.user channel to prevent conflicts between the printer name and user name.


fydai:
I worked with snarain in getting this done. It is a minor change and is tested on desktops, so I will quickly merge this.

PR #806, opened 2019-11-10
Update shipwreck.yaml

MrMinos:
add ocf-letsencrypt class


dkess:
Why are you doing this instead of `ocf::ssl::default`?

MrMinos:
> `default`

I'm planning to use `ocf-lets-encrypt` to create non ocf domain certs semi-manually, and not batched into one single cert.

PR #807, opened 2019-11-10
Revert "Update shipwreck.yaml"

MrMinos:
Reverts ocf/puppet#806

Found a better way, rather to obtain wildcard certs.


MrMinos:
Skipping review as this is a revert. (or should I wait?)

PR #811, opened 2019-11-14
add compton, fix screen tearing

ith8:
Watch anime on OCF computers without screen tearing!

*add compton, disable default compositor


nikhiljha:
There don't seem to be issues with tearing anymore, so I think we can close this. Thanks for the PR! :)

PR #814, opened 2019-12-09
remove pgp.pm from keyserver membership

ja5087:
rip their hosting provider


keur:
Should we just remove ourselves from the pool? There is basically no point in
running a PGP server anymore.

Kevin

cg505:
I got hit by the same problem... can't even download shit from my VM because they're so overloaded today as everyone tries to fix their stuff. RIP my custom URL shortener.

PR #815, opened 2019-12-09
remove pgp.pm from sks membership file

cg505:
The operator of this server sent us an email notifying us that it's shutting down.


cg505:
didn't see #814

PR #822, opened 2020-01-04
Irc sasl

keur:
We deliberately do not support the dh_aes (or its predecessor dh_blowfish)
mechanisms. These mechanisms are for users that want to send their
username and passwords encrypted to the server over an insecure
connection. Since the OCF only supports connecting over TLS there is no
point.

### DIFF
--- a/modules/ocf_irc/templates/inspircd.conf.erb
+++ b/modules/ocf_irc/templates/inspircd.conf.erb
@@ -13,6 +13,7 @@
 <module name="m_sha256.so">
 <module name="m_sqloper.so">
 <module name="m_ssl_gnutls.so">
+<module name="m_cap.so">
 
 # Required/Recommended modules for anope
 <module name="m_alias.so">
@@ -23,6 +24,9 @@
 <module name="m_spanningtree.so">
 <module name="m_svshold.so">
 
+# Add SASL authentication (provided by Anope)
+<module name="m_sasl.so">
+<sasl target="services.irc.ocf.berkeley.edu">
 
 <server name="irc.ocf.berkeley.edu"
         description="Open Computing Facility IRC Server"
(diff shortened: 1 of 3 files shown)


keur:
@jvperrin I've tested it on dev-flood. SASL PLAIN is the recommended way to connect to connect to IRC these days. It registers you with the server before the connection is made.

jvperrin:
Ah I see, carry on then, looks great!

dkess:
Can you please explain the PID file changes you made?

keur:
@dkess I moved the location from `/var/run` to `/run`; the latter is mounted as tmpfs and nowadays more standard for storing runtime information. The PIDFile= tells systemd where to look for the PID file and uses that to set $MAINPID. It's not strictly necessary since we specified the service as non-forking, but is good practice to use in services that create PID files (anope will regardless). The PermissionsStartOnly= directive is deprecated so I removed it since the PID file is being created with the correct user/group.

modules/ocf_irc/templates/inspircd.conf.erb
```diff
@@ -13,6 +13,7 @@
 <module name="m_sha256.so">
 <module name="m_sqloper.so">
 <module name="m_ssl_gnutls.so">
+<module name="m_cap.so">
```
cg505: why was this added?

ja5087: Seems to be the capability negotiation module. I was wondering about it too.

keur: SASL won't work otherwise. It's part of the ircv3 spec (SASL is ircv3, although everyone supports it). Your client will do use it to see available ircv3 features the server supports. You can do this in your client with `/quote CAP LS`

https://ircv3.net/specs/extensions/sasl-3.2

cg505: Awesome, sorry I missed that.

PR #824, opened 2020-01-15
ocf_irc: Improve the inspircd service file

keur:
The service file provided by Debian is really dated. Add overrides with
some options that upstream has added. I decided this was worthwhile when
the IRC server on dev-flood went down for a couple days due to a SIGSEV,
and was never brought back automatically since Restart= was not set.

https://github.com/inspircd/inspircd/blob/insp3/make/template/inspircd.service

### DIFF
--- a/modules/ocf_irc/files/inspircd.service.d/override.conf
+++ b/modules/ocf_irc/files/inspircd.service.d/override.conf
@@ -1,0 +1,7 @@
+[Unit]
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+Group=irc
+Restart=always
--- a/modules/ocf_irc/manifests/ircd.pp
+++ b/modules/ocf_irc/manifests/ircd.pp
@@ -56,4 +56,9 @@ class ocf_irc::ircd {
       mode      => '0640',
       show_diff => false;
   }
+
+  ocf::systemd::override { 'inspircd-group-restart':
+    unit   => 'inspircd.service',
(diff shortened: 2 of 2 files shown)


modules/ocf_irc/manifests/ircd.pp
```diff
@@ -55,5 +55,18 @@
...
       mode      => '0640',
       show_diff => false;
+
+    '/etc/systemd/system/inspircd.service.d/':
+      ensure => directory,
+      owner  => root,
+      group  => root,
+      mode   => '0755';
+
+    '/etc/systemd/system/inspircd.service.d/override.conf':
```
kpengboy: Consider using `ocf::systemd::override`?

keur: Good point forgot about that. Thanks.

modules/ocf_irc/files/inspircd.service.d/override.conf
```diff
@@ -0,0 +1,7 @@
+[Unit]
```
kpengboy: Nit: consider naming this file more descriptively, e.g. `inspircd-group-restart.conf`

keur:
@kpengboy these changes are from upstream

cg505:
I would assume he means upstream to Debian.

keur:
Somehow the service file hasn't improved for the buster release, so I'll consider it.

PR #828, opened 2020-01-19
ocf_backups: split backup before encrypting

cg505:
This makes it easier to recover from an encrypted backup. I don't have a machine with 5T, so I would have to put my private keys on an OCF server to recover from backups (since the whole ~5T image is encrypted). This change encrypts each 5G part separately, which allows me to download a chunk, decrypt it, move it to a server, then do the next chunk, without having them all on my machine at the same time.

I have tested the pieces locally but I will try running this on hal once the current backup is finished before merging.


kpengboy:
I wonder if we could make the backups per-user...

cg505:
> I wonder if we could make the backups per-user...

That would be awesome and came up at one point when we were talking about GDPR. It's definitely out-of-scope for this.

modules/ocf_backups/files/upload-to-box
```diff
@@ -142,10 +142,11 @@ def upload_to_box(creds, quiet):
                 '-1',
                 '--disable-epsv',
                 '--ftp-skip-pasv-ip',
-                '--ftp-ssl',
+                '--ssl',
```
dkess: nice catch

modules/ocf_backups/files/upload-to-box
```diff
@@ -217,6 +218,9 @@ def main(args):
     with open('/opt/share/backups/box-creds.json') as f:
         creds = json.load(f)
 
+    # Make sure our refresh token is still active, and fail otherwise.
+    get_access_token(creds)
```
jvperrin: I'd say putting a `try`/`except` or similar around the access token and sharing part below would be enough here, since then the backups would still be getting made but a warning could be put into the email if the token is bad (as mentioned on IRC by dkessler).

modules/ocf_backups/files/create-encrypted-backup
```diff
@@ -37,33 +37,37 @@ umask 077
...
+                 --compress-algo none \
+                 --recipient ckuehl@berkeley.edu \
+                 --recipient willh@ocf.berkeley.edu \
+                 --recipient jvperrin@ocf.berkeley.edu \
+                 --recipient mattmcal@ocf.berkeley.edu \
+                 --recipient kpengboy@ocf.berkeley.edu \
+                 --recipient abizer@ocf.berkeley.edu \
+                 --recipient benzhang4@gmail.com \
+                 --recipient daniel@dkess.me \
+                 --recipient cooperc@ocf.berkeley.edu \
```
jvperrin: Stretch goal: Feel free to prune the list here too, we haven't heard from some of these people in years and probably only current or reasonably recent people need to be on here. That being said, it's not a big deal, but probably good to update periodically. Keep in mind this list is also only used for new backups and any old backups would continue to be accessible, which seems reasonable to me.

modules/ocf_backups/files/create-encrypted-backup
```diff
@@ -37,33 +37,37 @@ umask 077
...
+                 --output \$FILE.gpg -"
+
+split_parts_and_encrypt() {
+    split -d -a 4 -b 5G --filter "$encrypt_cmd" - \
```
jvperrin: TIL `split` takes `--filter`, this is pretty neat

cg505:
okay I finally got this to work

cg505:
I used zerofree on the backups volume and now the encrypted upload is only 2.1T lol

jvperrin:
Nice, is that something that should be run regularly in this script? How long did it take?

Thanks for removing my Box account too, yeah I was notified that was disabled recently

cg505:
> Nice, is that something that should be run regularly in this script? How long did it take?

It took several days IIRC. I think it was only necessary because we added another disk to the LVM LV and that disk was full of stuff. Otherwise, it's only necessary once every few years or something probably.

boomtow:
I think it took me just over a week maybe closer to 2 weeks

PR #830, opened 2020-01-24
change eruption to use new 3-monitor setup

cg505:
This is currently running on eruption. Unfortunately we have to get rid of the panels spanning monitors change because it's totally broken.

### DIFF
--- a/modules/ocf_desktop/files/xsession/fix-displays
+++ b/modules/ocf_desktop/files/xsession/fix-displays
@@ -54,11 +54,26 @@ if __name__ == '__main__':
     else:
         sorted_monitors = list(sorted(monitors))
 
-    prev = None
-    for monitor in sorted_monitors:
-        if prev:
-            subprocess.check_call(['xrandr', '--output', monitor, '--right-of', prev])
-        prev = monitor
+    # special case for eruption: right monitor is rotated left 90 degrees
+    if len(monitors) == 3 and hostname == 'eruption':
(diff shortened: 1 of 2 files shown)


modules/ocf_desktop/files/xsession/fix-displays
```diff
@@ -54,11 +54,26 @@ if __name__ == '__main__':
     else:
         sorted_monitors = list(sorted(monitors))
 
-    prev = None
-    for monitor in sorted_monitors:
-        if prev:
-            subprocess.check_call(['xrandr', '--output', monitor, '--right-of', prev])
-        prev = monitor
+    # special case for eruption: right monitor is rotated left 90 degrees
+    if len(monitors) == 3 and hostname == 'eruption':
```
ja5087: I'm not a fan of encoding special cases in the program like this. Is there a way to encode this info per host that the program could fetch? or even some kind of configuration file that overrides specifically for eruption.

dkess: Could be worth looking into something like [autorandr](https://github.com/phillipberndt/autorandr) (which is in the Debian Buster repos) to dynamically set the monitor configs.

cg505: autorandr looks nice

cg505:
opened #834 to track autorandr and other improvements. I'm gonna merge this right now because that shouldn't block eruption working correctly.

PR #833, opened 2020-01-25
Increase php maximum file size

fydai:
asrb tried for months to valiantly trying to upload a 21mb theme


cg505:
Why do we even have this limit in the first place?

fydai:
/shrug
also octo-catalogdiff is amazing

PR #835, opened 2020-01-27
Move ldap-lint from utils, add groups and kerberos principal checking

jvperrin:
This closes https://github.com/ocf/utils/issues/119 (moving `ldap-lint` into puppet since it's not a script that's run by hand like most of the ones in utils are) and https://github.com/ocf/utils/issues/14 (adding support to `ldap-lint` for checking user groups and kerberos principals).

I tested this on `dev-firestorm`, it's currently live there, but I also copied it to `firestorm` and tested there, with the same results.

Here's the diff in moving it from utils: https://i.fluffy.cc/3lVklzclpMP4znLMz4t1XrTTgf8zwdMc.html
I did do make some other changes, like making it only query the local ldap instead of always the ldap on `firestorm` (otherwise it's not really that useful on `dev-firestorm`, after all), and some structural changes to make it easier to add more check functions in the future.

I've closed #702 in favor this review.

### DIFF
--- a/modules/ocf_ldap/files/ldap-lint
+++ b/modules/ocf_ldap/files/ldap-lint
@@ -1,0 +1,185 @@
+#!/usr/bin/env python3
+"""Check the sanity of our LDAP database.
+
+Currently checks the Hosts OU for:
+    * Duplicate IP addresses
+    * Duplicate MAC addresses
+    * Missing IP address
+    * Existence of MAC address (should be present only on type=desktop)
+    * Invalid or missing type
+    * Host not in DNS or mismatched IP
+    * Unrecognized puppetVar (often typos)
+    * Reverse DNS for IP exists and matches hostname
+"""
+import subprocess
+import sys
+from ipaddress import ip_address
+from operator import itemgetter
+
+import dns
+from ocflib.infra.ldap import ldap_connection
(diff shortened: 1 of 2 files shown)


modules/ocf_ldap/files/ldap-lint
```diff
@@ -0,0 +1,190 @@
...
+            complain(user, 'in ocfroot but has no root principal')
+
+        if user not in admin_users:
+            complain(user, 'in ocfroot but has no admin principal')
+
+
+def main():
+    retval = 0
+
+    def complain(cn, error):
```
cg505: can't we just... define this at the top level to avoid having to awkwardly pass it into `check_hosts` and `check_users`?

jvperrin: The issue with that is that it also sets `retval` along the way. We could just make that a global and that'd work though, so that's probably just worth doing.

jvperrin: From trying this out I'm not sure it's really much cleaner. Yes the function isn't nested, but you still need the `nonlocal retval`, it's just essentially not passing it into two functions and moving it further up in the file...

cg505: That would still be my preference but I see your point.

PR #836, opened 2020-01-27
Bump puppetfile

jvperrin:
I'm curious to see what `octocatalog-diff` does with this one since I suspect there's going to be quite a few changes, and I'd like to see how it differs from what an actual puppet run does for instance (if at all).

Also wow does the diff look messed up on this, but I essentially just bumped all the modules to their latest versions. I suspect this'll break some things or that there might be missing dependencies, so that's why I'd like to see what `octocatalog-diff` has to say

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -6,27 +6,26 @@ moduledir 'vendor'
 # too, since r10k does not manage dependencies and you will be confused as to
 # why they are not working properly.
 
-mod 'alexharvey/firewall_multi',           '1.10.0'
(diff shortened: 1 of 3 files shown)


jvperrin:
Well, looks like this needs some more work, I'll try working on this another day. The `octocatalog-diff` diffs aren't particularly useful here either since they seem to not exactly match in order, making our script think that they are all different changes and not bundling them together at all. Maybe it can do some kind of sort + hash to figure out if they are the same content to combine them, but I'll think about that more when I revisit this.

PR #838, opened 2020-01-27
Use puppetlabs and backports apt repos for buster

jvperrin:
[Puppetlabs now has an apt distribution for buster](http://mirrors.ocf.berkeley.edu/puppetlabs/apt/dists/buster/), and [debian has buster-backports](http://mirrors.ocf.berkeley.edu/debian/dists/buster-backports/), so let's enable both of those and make things a bit more standard when on buster.

Note: This still needs testing on a machine running buster to see how it'll do with the switch in puppet versions. I think it should be fine, but it's worth an actual live test to make sure.

### DIFF
--- a/hieradata/os/buster.yaml
+++ b/hieradata/os/buster.yaml
@@ -1,2 +1,0 @@
-# TODO: Remove this once puppetlabs releases packages for buster
-puppet_agent: false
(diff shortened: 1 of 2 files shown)


jvperrin:
Yeah this'll need a bit of manual work on the hosts that aren't on the puppetlabs version. From testing on dev-flood, they'll essentially need a new bootstrap and cert signed, but that should be pretty quick since it only need to be done on a few hosts to fully switch over.

jvperrin:
I did the necessary cleanup on `dev-fallingrocks` and `dev-flood`, `y2k` still needs to be done but it's currently powered off, and all other hosts on buster are already on the puppetlabs version.

PR #839, opened 2020-01-27
use buster puppet-agent packaging

cg505:


cg505:
sorry, didn't even see that PR before I opened mine!

PR #841, opened 2020-01-28
Add a noop mirrors sync mode and key for dev-fallingrocks

jvperrin:
This adds a noop syncing mode for dev-fallingrocks mirrors where they just print to syslog that they would have tried to sync something, and adds a separate key to sign internal (`ocf_apt`) packages with, since I didn't feel comfortable using the production key with it given it can sign all internal packages if it is improperly released.

I mostly did this while investigating applying #838 on `dev-fallingrocks` and then finding it was out of disk space due to trying to sync a bunch of mirrors with only a 16 GB root drive.

CC @pxhanus as I believe you were working on this last?

### DIFF
--- a/hieradata/nodes/dev-fallingrocks.yaml
+++ b/hieradata/nodes/dev-fallingrocks.yaml
@@ -1,1 +1,3 @@
-ocf::networking::bond: true
+classes:
+    - ocf_apt
+    - ocf_mirrors
--- a/modules/ocf_apt/manifests/init.pp
+++ b/modules/ocf_apt/manifests/init.pp
@@ -10,6 +10,12 @@ class ocf_apt {
 
   package { 'reprepro':; }
 
+  if $::host_env == 'prod' {
+    $pkg_key = 'D72A0AF4'
+  } else {
+    $pkg_key = 'D0BA5B90'
+  }
+
   file {
     default:
       owner => ocfapt,
@@ -23,7 +29,7 @@ class ocf_apt {
       source => 'puppet:///modules/ocf_apt/README.html';
 
     '/opt/apt/etc/distributions':
-      source => 'puppet:///modules/ocf_apt/distributions';
+      content => template('ocf_apt/distributions.erb');
 
     '/opt/apt/bin':
       ensure  => directory,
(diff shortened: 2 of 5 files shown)


jvperrin:
lol the diff for `dev-fallingrocks` so large that it needs to be put into a gist...

Anyway, I've run puppet already on `dev-fallingrocks`, so it doesn't actually have all that as the diff, but that's the diff versus what is currently on `master`, which is how octocatalog-diff works. Maybe a downside of it, but I guess if the goal is to have things on the production environment anyway then it's not too far from the truth.

jvperrin:
@cg505 Hmmm, maybe. I think we could definitely configure projects in hieradata, but I don't think the 16 GB on fallingrocks right now will be enough to sync almost anything. From https://mirrors.ocf.berkeley.edu/sizes for instance, there's only a couple projects that would actually fit within that size and then the root disk will be completely full. Otherwise, I think dev-fallingrocks needs to be reconfigured to have the mirrors on a separate volume or something so that it doesn't impact the rest of the host if it fills up.

cg505:
What if we instead added a `sync_dev` argument to `ocf_mirrors::timer` so that by default we don't sync on `dev-fallingrocks` but we can just enable it by modifying the `timer` resource parameters? Obviously there aren't many projects that can fit on the current `dev-fallingrocks` but historically and in the future there will definitely be projects we want to test syncing with.

jvperrin:
Yeah that might work. I still like your idea about using hiera for this though, it would make sense to use anyway since essentially all of the mirror project manifests are encoding data but have the same structure to them (which is exactly what hiera is good for). Ideally for `dev-fallingrocks` I think we'd have it run rsyncs but do them in `--dry-run` mode or something so that it's clear it can connect and all that but it just doesn't actually download the files. Then I think that'd be a close enough test that a mirror or a project actually works before putting it on the real fallingrocks.

cg505:
> Ideally for `dev-fallingrocks` I think we'd have it run rsyncs but do them in `--dry-run` mode or something so that it's clear it can connect and all that but it just doesn't actually download the files.

that's super tasty

PR #842, opened 2020-01-28
Change ubuntu upstream to official us.archive.ubuntu.

redplanks:
Our previous upstream went out of date; the official US upstream seems to have good enough bandwidth.

Tested--fallingrocks is on my env right now.


cg505:
thank you!

PR #845, opened 2020-01-30
Add Helm 3.0.3

ja5087:
I've been playing with Helm 3 recently and it's actually been very useful. Unfortunately, it's also not packaged in the Debian apt repo and the Puppet module is for 2.x (which has Tiller). 

This might not be the best way to deploy it, so I'm open to comments.

### DIFF
--- a/modules/ocf/manifests/packages.pp
+++ b/modules/ocf/manifests/packages.pp
@@ -10,6 +10,7 @@ class ocf::packages {
   # special snowflake packages that require some config
   include ocf::packages::git
   include ocf::packages::grub
+  include ocf::packages::helm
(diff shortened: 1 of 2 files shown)


Baisang:
Worth noting that we don't need anything installed on our k8s cluster for this; it's literally just a binary that we run to install random crap on our k8s cluster.

As we discussed offline, we'll need to figure out a good way to do "version-controlled deployments" (i.e. we make some git commit to add some helm chart to install on our cluster, and Jenkins handles the actual building and running) for helm stuff later on.

Baisang:
Spoke with @jvperrin offline at the OCF satellite office and he also approved

PR #847, opened 2020-02-03
add authentication to prometheus.ocf.berkeley.edu

cg505:
This requires authentication as ocfstaff, or via an htpasswd file, to access prometheus.ocf.berkeley.edu. This properly restricts access to Alertmanager. The htpasswd file is used for allowing access via external applications such as Grafana.

This depends on ocf/grafana#10 to be merged first, otherwise Grafana will break.


cg505:
Also, before you ask: there's not a way to set the default creator in alertmanager based on a proxy auth header. see prometheus/alertmanager #1196

PR #855, opened 2020-02-09
Shuffle around includes for the base ocf_hpc class

redplanks:
Shouldn't change anything right now, but will help in case we ever want to put just the ocf_hpc class on a node so it can issue slurm commands, but not be a compute or control node.

### DIFF
--- a/modules/ocf_hpc/manifests/compute.pp
+++ b/modules/ocf_hpc/manifests/compute.pp
@@ -1,6 +1,11 @@
 class ocf_hpc::compute {
   require ocf_hpc
 
+  include ocf::ipmi
+  include ocf::firewall::allow_ssh
+
+  include ocf_hpc::singularity
+
   # install proprietary nvidia drivers and CUDA.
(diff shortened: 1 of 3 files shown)


modules/ocf_hpc/manifests/compute.pp
```diff
@@ -1,6 +1,11 @@
 class ocf_hpc::compute {
   require ocf_hpc
 
+  include ocf::ipmi
+  include ocf::firewall::allow_ssh
+
```
ja5087: style: any reason for the space?

redplanks: Thought it looked slightly better to separate the ocf_hpc subclass from the others (though I didn't do it for the other one lol)

modules/ocf_hpc/manifests/compute.pp
```diff
@@ -1,6 +1,11 @@
 class ocf_hpc::compute {
   require ocf_hpc
 
+  include ocf::ipmi
+  include ocf::firewall::allow_ssh
```
cg505: why do we want this?

PR #857, opened 2020-02-09
add prometheus pushgateway

ja5087:
This allows us to push service-level metrics that might not belong to a specific node e.g. suffocation.

### DIFF
--- a/modules/ocf_prometheus/manifests/init.pp
+++ b/modules/ocf_prometheus/manifests/init.pp
@@ -1,5 +1,6 @@
 class ocf_prometheus {
   include ocf_prometheus::alertmanager
   include ocf_prometheus::proxy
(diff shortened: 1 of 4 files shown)


dkess:
It's fine if you want to install the Pushgateway, but metrics from the CO2 meter don't have to go through this. The easiest way to export those is to run a cronjob or systemd timer that writes the metrics to a text file in `/srv/prometheus` and they'll get included in the node exporter. You can run this on every desktop and make it a no-op if the meter isn't connected.

ja5087:
So my take on this is that it's nicer to have the pushgateway cache these metrics instead for the CO2 meter specifically since it'll be an ephemeral but periodic python script. Also it'll avoid them being attached to an instance.

PR #858, opened 2020-02-09
Change to using puppetlabs repo for puppet-agent

ja5087:
Not tested but it's in the repo now and that's the only change from stretch


cg505:
build failing cause IRC is down?

PR #860, opened 2020-02-09
Fix inspircd 2 on Buster

ja5087:
The root cause for this is still unknown, hopefully we can solve it, or it may be solved by moving to inspircd3.


dkess:
What exactly is the issue here?

ja5087:
See commit message. We isolated it to this module (database connection).

PR #861, opened 2020-02-09
InspIRCd v3

ja5087:
Testing on dev-flood. It seems to be working nicely so far.

Thanks @fydai for the assist


modules/ocf_irc/templates/inspircd.conf.erb
```diff
@@ -57,27 +60,22 @@
 
 <uline server="services.irc.ocf.berkeley.edu" silent="yes">
 
-
-# Passwords needed to kill/restart the IRCd while connected to it
-<power diepass="<%= @irc_creds['die_password'] -%>"
-       restartpass="<%= @irc_creds['restart_password'] -%>"
-       pause="5">
```
cg505: this does get used, is there no replacement?

ja5087: See https://docs.inspircd.org/3/commands/#die-ltservergt

/DIE and /RESTART no longer take in a password but instead take the server name. They still require server op status like always.

cg505: in that case, do we need to update modules/ocf_irc/files/reload-ssl.sh?

ja5087: Looking through it, I'm not sure how that uses /DIE or /RESTART

cg505: ah, I was confusing them

modules/ocf_irc/templates/inspircd.conf.erb
```diff
@@ -36,12 +36,15 @@
        nick="ocf"
        email="help@ocf.berkeley.edu">
 
-<gnutls certfile="/etc/ssl/private/<%= @fqdn -%>.bundle"
-        keyfile="/etc/ssl/private/<%= @fqdn -%>.key"
-        priority="NORMAL:-MD5">
+<sslprofile name="clientssl"
+       provider="gnutls"
+       certfile="/etc/ssl/private/dev-flood.ocf.berkeley.edu.bundle"
+       keyfile="/etc/ssl/private/dev-flood.ocf.berkeley.edu.key"
```
cg505: lol

PR #862, opened 2020-02-10
ocf_ns: add also-notify setting for infoblox gridmaster

cg505:
per hostmaster email on 8/6/19


cg505:
I wonder if we should also lock down the ACL for zone transfers a little more. All of UCB seems like it might be a bit too permissive.

PR #868, opened 2020-02-13
GC controller support for desktops

BernardZhao:
Need to test on desktops, but I wanted some help since I'm not too familiar with Puppet.

### DIFF
--- a/modules/ocf/manifests/packages/dolphin_emu.pp
+++ b/modules/ocf/manifests/packages/dolphin_emu.pp
@@ -1,3 +1,0 @@
-class ocf::packages::dolphin_emu {
-  package { 'dolphin-emu': }
-}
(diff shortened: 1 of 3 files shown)


modules/ocf/manifests/packages/dolphin_emu.pp
```diff
@@ -0,0 +1,3 @@
+class ocf::packages::dolphin_emu {
```
dkess: If this package doesn't have a special setup needed to install it, you don't need to make a new Puppet class for it, just add it to the list of desktop packages at https://github.com/ocf/puppet/blob/master/modules/ocf_desktop/manifests/packages.pp

cg505:
isn't dolphin in debian super old or something?

BernardZhao:
> 
> 
> isn't dolphin in debian super old or something?

It definitely is, but honestly I'm not sure how to check what the version in debian even is,

kpengboy:
@BernardZhao packages.debian.org

BernardZhao:
After testing the version in Debian and a version I built myself, the performance is just too poor to justify installing the official version. I guess if people want to play, we can just play off of my nfs.

Baisang:
@BernardZhao Please squash your commits first

PR #872, opened 2020-02-19
add snmp-exporter metrics from switches to prometheus

cg505:
also, refactor our gen-prometheus-* scripts to avoid code duplication.

depends on ocf/snmp_exporter#6


modules/ocf_prometheus/files/gen-prometheus-nodes
```diff
@@ -2,32 +2,65 @@
...
+def get_hosts(types, port):
+    nodefilter = '(|{typefilters})'.format(
+        typefilters=''.join('(type={})'.format(nodetype) for nodetype in types)
+    )
+    hosts = []
+    for vals in hosts_by_filter(nodefilter):
+        cn, ip, host_type = vals['cn'][0], vals['ipHostNumber'][0], vals['type']
+
+        if port is not None:
+            ip = ip + ':{}'.format(port)
```
dkess: Is this output useful at all without a port number? If not, we should require it and not bother with this edge case. My hunch is that it's better to be explicit about this anyways.

cg505: It's useful because of our renumbering technique with the printers and switches.

PR #875, opened 2020-02-23
node_exporter: enable the systemd collector

ja5087:
Lets us tie in services status to prometheus


dkess:
This is fine, but keep an eye on disk usage since this has the potential to blow up in size. Might have to filter out unnecessary units on the scraper side.

ja5087:
That's a good idea. There's also `--collector.systemd.unit-whitelist` on the node_exporter side.

ja5087:
I've enabled just *.service and *.timer which seems to only increase the filesize by a nominal amount.

PR #876, opened 2020-02-23
Delete Ubuntu Sync Script

ja5087:
We haven't synced from MIT in a long time


ja5087:
Ohh that makes sense. Gotcha.

PR #882, opened 2020-02-28
Add file manager to panel

mcint:


ethanhs:
I saw it running on a desktop yesterday, FWIW @jvperrin, looks like the file manager icon on the taskbar as expected :P

mcint:
Now:
image

Proposed: 
image

PR #883, opened 2020-02-28
Hw monitor controls

mcint:
Monitor status reporting and controls of hardware.

Some i2c commands can brick monitors, so this functionality should be limited to root for now.

### DIFF
--- a/modules/ocf_desktop/manifests/displaycontrol.pp
+++ b/modules/ocf_desktop/manifests/displaycontrol.pp
@@ -1,0 +1,16 @@
+class ocf_desktop::displaycontrol {
+
+  file {
+    '/etc/modules-load.d/i2c.conf':
+      mode    => '0644',
+      content => 'i2c_dev\n',
+  }
+
+  exec { 'load-modules':
+    command     => '/lib/systemd/systemd-modules-load',
+    subscribe   => File['/etc/modules-load.d/i2c.conf'],
+    refreshonly => true,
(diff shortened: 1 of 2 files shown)


dkess:
We already have the [`kmod` Puppet module](https://forge.puppet.com/camptocamp/kmod) so maybe use that instead of manually adding the files?

mcint:
Hmm, sweet, thanks.

Not sure what the difference between `kmod::load` and `kmod::install` (more flexible), esp. which I should use. https://forge.puppet.com/camptocamp/kmod

modules/ocf_desktop/manifests/displaycontrol.pp
```diff
@@ -0,0 +1,6 @@
+class ocf_desktop::displaycontrol {
+
+  kmod::load { 'i2c_dev': }
```
cg505: should require the package resource

modules/ocf_desktop/manifests/displaycontrol.pp
```diff
@@ -0,0 +1,6 @@
...
+
+  package { [ 'i2c-tools', 'ddcutil' ]: }
```
cg505: `i2c-tools` is a dep of `ddcutil`. We should only explicitly install the latter.

cg505:
> Does this require root by default, or do we have to do extra work?

I believe it does by default

PR #884, opened 2020-02-28
Added minetest to games

boihem:


modules/ocf_desktop/manifests/packages.pp
```diff
@@ -30,8 +30,8 @@
     # FUSE
     ['fuse', 'exfat-fuse']:;
     # games
-    ['armagetronad', 'gl-117', 'gnome-games', 'redeclipse', 'wesnoth',
-      'wesnoth-music']:;
+    ['armagetronad', 'gl-117', 'gnome-games', 'minetest', 'redeclipse', 'wesnoth',
+      'wesnoth-music',]:;
```
kpengboy: Nit: remove trailing comma, as we don't use it in these sets of lists.

PR #888, opened 2020-03-03
fix node_exporter extra_options

cg505:
The defaults were unintentionally overridden in #875.

### DIFF
--- a/modules/ocf/manifests/node_exporter.pp
+++ b/modules/ocf/manifests/node_exporter.pp
@@ -8,9 +8,7 @@ class ocf::node_exporter {
   }
 
   if $::lsbdistid != 'Raspbian' {
-    class { 'prometheus::node_exporter':
-      collectors_enable => ['systemd'],
-      extra_options     => '--collector.systemd.unit-whitelist ".+\.(service|timer)"',
-    }
+    # Attributes for this class are defined in hieradata
+    include prometheus::node_exporter
   }
 }
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -79,4 +79,6 @@ kubernetes::master_nodes:
 
 # Prometheus config
 prometheus::node_exporter::version: 0.17.0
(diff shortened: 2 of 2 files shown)


ja5087:
We could also move the version

cg505:
for whatever reason it makes more sense to me to keep the version in hiera

jvperrin:
IMO putting all of it in hiera would be better if that's possible

cg505:
@jvperrin don't have a problem with that, but why do you think that's better?

cg505:
I am inclined to move it out of hiera since it's already in the ocf module, and being in hiera decreases discoverability which could cause more bugs like this one to arise in the future.

ja5087:
I think we should do one or the other. Personally I like keeping values local to their usage until there's a need to override/lookup them in other places, but I can see an argument for keeping it all in hiera to specify defaults.

jvperrin:
Hiera is generally the place to put configuration data so it's easy to override, but I do see what you mean since having it in hiera this time caused this issue. I don't have a strong preference, either place seems ok (if it does go into hiera, add a comment about checking hiera in this manifest).

cg505:
Ah, I didn't think about overriding. I'll move to hiera and add a comment.

cg505:
@jvperrin I moved all the config into hieradata (including adding the systemd collector)

modules/ocf/manifests/node_exporter.pp
```diff
@@ -7,10 +7,8 @@
     content => template('ocf/environment.prom.erb'),
   }
 
+  # Attributes for this class are defined in hieradata/common.yaml
```
jvperrin: I'd move this right before the class (inside the conditional), just so it stays closer to the class that it's referring to.

nit: Just saying "in hieradata" is probably better as it could be defined at any level of hiera, not just in `common.yaml` (though that's currently the only place it's defined).

PR #901, opened 2020-03-12
update firewall_multi

cg505:
We are technically running incompatible versions right now.


jvperrin:
I wish puppet had a real dependency manager, but oh well. Maybe we can have a list of base dependencies and then a script to install them with `puppet module install` commands since that apparently does take dependencies into account, but I think we'd still have trouble with multiple modules depending on different versions for instance

PR #905, opened 2020-03-13
update resource types when deploying puppet

cg505:
Today, all puppet builds were failing because firewall_multi was updated and had new type definitions, but `puppet generate types` wasn't run. This will just run that every time we deploy.

See https://puppet.com/docs/puppet/latest/environment_isolation.html for more info. Note that it specifically mentions that we should use `--force` for every new deployment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ WORKSPACE ?= ${HOME}/.cache
 ENVIRONMENT := $(notdir $(realpath $(CURDIR)))
 
 .PHONY: all
-all: vendor install-hooks
+all: vendor .resource_types install-hooks
 
 .PHONY: venv
 venv: bin/venv-update Makefile
@@ -20,6 +20,9 @@ install-hooks: venv
 
 vendor: Puppetfile
 	r10k puppetfile install --verbose --color
+	@echo 'You may also want to run `make .resource_types`'
+
+.resource_types: vendor
 	puppet generate types --environment ${ENVIRONMENT} --force || (\
 		echo '\e[1;31mPlease run' && \
 		echo 'mkdir -p ~/.puppetlabs/etc/puppet/ && cp /etc/skel/.puppetlabs/etc/puppet/puppet.conf ~/.puppetlabs/etc/puppet/\e[0m' && \
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 WORKSPACE ?= ${HOME}/.cache
+ENVIRONMENT := $(notdir $(realpath $(CURDIR)))
 
 .PHONY: all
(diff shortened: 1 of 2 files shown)


cg505:
> Do you not need the environment to be specified if it's for the current directory or something?

It defaults to production. I think (?) if you run it without the --environment flag it will update production and also the default types used by environments that don't have a .resource_types directory.

cg505:
> Oh yeah, I meant to mention this on #904, but I'd also add to the `Makefile` under the `vendor` target to run `puppet generate types --environment $USER` or something like that so that people working on their environment also get updated types

@jvperrin done, can you re-review?

cg505:
Actually I guess if we're doing it this way, we don't need the update to deploy-prod.

cg505:
fuck, this breaks Jenkins.

Makefile
```diff
@@ -19,6 +20,13 @@ install-hooks: venv
 
 vendor: Puppetfile
 	r10k puppetfile install --verbose --color
+	@echo 'You may also want to run `make .resource_types`'
+
+.resource_types: vendor
+	puppet generate types --environment ${ENVIRONMENT} --force || (\
+		echo '\e[1;31mPlease run' && \
+		echo 'mkdir -p ~/.puppetlabs/etc/puppet/ && cp /etc/skel/.puppetlabs/etc/puppet/puppet.conf ~/.puppetlabs/etc/puppet/\e[0m' && \
```
jvperrin: Why not just fix this for people automatically instead of asking them to run this? Then it could also be included as part of the `vendor` target too, right?

cg505: I don't really want to bludgeon config files in their home directory whenever `puppet generate types` fails (which could happen for some other reason). I only really put this message here for old staff that run into this error unexpectedly.

Moving it out of `vendor` is unrelated: Jenkins does `make vendor` and that will fail because it is not run on the puppetserver.

jvperrin: Ah I see, that makes sense, I hadn't considered how jenkins runs this.

I think changing puppet-specific config files even in a home dir would be fine, but it's a not a big deal, it only has to be run at most once by each staff member anyway, so whatever works.

modules/ocf_puppet/files/update-prod
```diff
@@ -17,7 +17,7 @@ else
 fi
 
 # update from remote
-git fetch origin
+git fetch origin master
 git reset --hard origin/master
 
 # Install third-party modules
```
jvperrin: Don't you need a `make .resource_types` in here too now that it's moved to a different target?

cg505: fuck

PR #908, opened 2020-03-14
move keepalived configuration into ocf_lb

cg505:
This allows us to easily create new keepalived VIPs, either as a solo thing or included from other classes. I'm planning to use this for Ceph.

There's also some changes to how SSL certs on the k8s masters are generated. Instead of using the node FQDN as the cert title, we just use lb.ocf.berkeley.edu as the title (which makes more sense logically).

The first commit should essentially be a no-op. The only changes are the dehydrated timeout and the cert changes.


modules/ocf_lb/templates/keepalived.conf.erb
```diff
@@ -6,7 +6,7 @@ vrrp_instance VI_1 {
 
     authentication {
         auth_type PASS
-        auth_pass $ <%= @keepalived_secret %>
+        auth_pass $ <%= @secret %>
```
cg505: @keur why is there a `$` here?

modules/ocf_lb/manifests/init.pp
```diff
@@ -0,0 +1,21 @@
+class ocf_lb (
+  Array $vip_names,
+  String $keepalived_secret_lookup,
+) {
+  $virtual_addresses = $vip_names.map |$vip| {
+    ldap_attr($vip, 'ipHostNumber')
+  }
+  $virtual_addresses_v6 = $vip_names.map |$vip| {
+    regsubst(ldap_attr($vip, 'ipHostNumber'), '^(\d+)\.(\d+)\.(\d+)\.(\d+)$', '2607:f140:8801::1:\4')
```
keur: oh my god this was confusing code. Since this is specific to our IP numbering scheme now, it is worth abstracting this into a function called `OcfV6FromV4`. To simplify things, I would further suggest also storing this information in ldap. 

If you decide to stick with this, doesn't hurt to tighten the regex with `\d{1,3}`

modules/ocf_lb/templates/keepalived.conf.erb
keur: I would double check and see if it is even getting rendered on the hosts. I skimmed the manpage and it seems okay to remove. This was largely ported from the `ocf_mesos` code, so this is a blindly copied from there. And god that diff was fucking hard to find. This is why there was literally no point at all for us to remove the ocf_mesos modules rather than just remove them from hiera. There is lots of context that influences today's design in the old code. Sigh.

https://github.com/ocf/puppet/blob/8b6cb903d287f7544282bfc78355283dce2a0eb6/modules/ocf_mesos/templates/master/load_balancer/keepalived.conf.erb#L9

modules/ocf_lb/manifests/init.pp
dkess: sigh... I guess this is fine

modules/ocf_kubernetes/templates/master/loadbalancer/haproxy.cfg.erb
```diff
@@ -28,7 +28,7 @@ frontend lb-http
   redirect scheme https code 301 if !{ ssl_fc }
 
 frontend lb-https
-  bind 0.0.0.0:443 ssl crt /etc/ssl/private/<%= @fqdn %>.pem
+  bind 0.0.0.0:443 ssl crt /etc/ssl/private/lb.ocf.berkeley.edu.pem
```
keur: I see why you did this, and yeah it makes sense. This is strictly clearer than fqdn, but would slightly prefer a constant. If you left as-is that's OK too.

modules/ocf_lb/manifests/keepalived.pp
```diff
@@ -0,0 +1,18 @@
+class ocf_lb::keepalived(
+  Array $virtual_addresses,
```
keur: can we be more clear here with virtual_addresses_v4

modules/ocf_lb/manifests/init.pp
cg505: copy pasted from https://github.com/ocf/puppet/blob/2623cd0c204aec6f1d0496575c6cb06b5ffdda5c/modules/ocf/manifests/networking.pp#L9 😬

modules/ocf_lb/manifests/keepalived.pp
```diff
@@ -0,0 +1,19 @@
...
+  Array[String] $virtual_addresses_v6,
+  String $secret_lookup,
+  Integer[1, 255] $vrid
+) {
+  $secret = lookup($secret_lookup)
+
+  package { 'keepalived':; } ->
+  # IPv6 addresses have to be specified separately as they cannot be in the vrrp
+  # packet together (keepalived 1.2.20+) so they need to be in a
+  # virtual_ipaddress_excluded block instead.
```
jvperrin: I'd say this comment makes more sense in the actual template itself rather than here

modules/ocf_kubernetes/manifests/master/loadbalancer.pp
```diff
@@ -69,31 +69,10 @@
...
-  service { 'keepalived': }
-
-  $vip = 'lb'
-
-  class { 'ocf_kubernetes::master::loadbalancer::ssl':
-    vip => $vip,
+  class { 'ocf_lb':
+    vip_names                => ['lb'],
+    keepalived_secret_lookup => 'kubernetes::keepalived::secret',
+    vrid                     => 51,
```
jvperrin: Why make this dynamic if only one value is used after all that? I'd just leave it out for now, hardcode it in the template, and refactor later if you do need the complexity of having multiple `vrid`s

cg505: I added this after I tried to add this class in the ceph module and ran into issues. It's necessary to avoid that if we have more than one VIP anywhere (which is kind of the whole point).

modules/ocf_lb/manifests/init.pp
dkess: ++ to storing this in LDAP

cg505: time to figure out ldap-overlay

modules/ocf_kubernetes/manifests/master/loadbalancer.pp
jvperrin: Ah I see, I didn't know you were planning to use this for ceph, that makes more sense then, keep it like this.

modules/ocf_kubernetes/templates/master/loadbalancer/haproxy.cfg.erb
cg505: I'm planning to revamp this config heavily at some point in the future as well. There are a lot of things that can be improved.

modules/ocf_lb/manifests/init.pp
cg505: #912

modules/ocf_kubernetes/templates/master/loadbalancer/keepalived.conf.erb
```diff
@@ -1,23 +0,0 @@
-vrrp_instance VI_1 {
-    interface <%= @networking['primary'] %>
-    state MASTER
-    virtual_router_id 51
-    priority 100
-
-    authentication {
-        auth_type PASS
-        auth_pass $ <%= @keepalived_secret %>
-    }
```
jvperrin: Why is this removed without a replacement? Just curious

cg505: see https://github.com/ocf/puppet/pull/908#discussion_r392554803. It's deprecated and I don't think it's even doing anything right now.

jvperrin: Ah I see:

```
# Note: authentication was removed from the VRRPv2 specification by
# RFC3768 in 2004.
#   Use of this option is non-compliant and can cause problems; avoid
#   using if possible, except when using unicast, where it can be helpful.
```

Well, sounds good to remove then

PR #910, opened 2020-03-14
fix dehydrated permissions problems on lethe

cg505:
This resolves the issues with #669. Apparently the issue was just that we had specified `DEHYDRATED_USER` in the dehydrated config, so if dehydrated was run as root it would just sudo to that user. This removes that config and leaves the specification the the user running dehydrated to puppet. This allows us to revert #669 and remove the lethe puppet spam.

### DIFF
--- a/modules/ocf/files/ssl/dehydrated-config
+++ b/modules/ocf/files/ssl/dehydrated-config
@@ -3,8 +3,6 @@ BASEDIR="/var/lib/lets-encrypt"
 HOOK="/usr/share/dehydrated-hook-ddns-tsig/dehydrated-hook-ddns-tsig.py"
 CHALLENGETYPE="dns-01"
 
-DEHYDRATED_USER="ocfletsencrypt"
-
 # if you update this, run
 # dehydrated --privkey /etc/ssl/lets-encrypt/le-account.key --account
 # with the new config in place to update LE's records
--- a/modules/ocf_postgres/manifests/init.pp
+++ b/modules/ocf_postgres/manifests/init.pp
@@ -32,20 +32,6 @@ class ocf_postgres {
       };
   }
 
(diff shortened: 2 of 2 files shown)


modules/ocf/files/ssl/dehydrated-config
```diff
@@ -3,8 +3,6 @@ BASEDIR="/var/lib/lets-encrypt"
 HOOK="/usr/share/dehydrated-hook-ddns-tsig/dehydrated-hook-ddns-tsig.py"
 CHALLENGETYPE="dns-01"
 
-DEHYDRATED_USER="ocfletsencrypt"
```
jvperrin: I'm concerned that this will mess up existing hosts where they expect this user to be the one doing cert-related things (like death or vampires for instance). I haven't tested this yet though, so I'm not sure.

As I remember this user was originally created because certs on death/vampires were stored on NFS and that plays better with a non-root user accessing them. Maybe this will work fine because that stuff for user/vhost certs is separated out into a separate script and dehydrated isn't running that?

cg505: To be clear, the ocfletsencrypt user will still exist and will still be used in (almost) all invocations of dehydrated, as it's the default in https://github.com/ocf/puppet/blob/7d7b75f783ed856ef53fd6359a5b2f9b92fe0d14/modules/ocf/manifests/ssl/lets_encrypt/dns.pp#L4

death or vampires shouldn't be using dehydrated at all (for vhost certs at least), so this shouldn't affect them. They use ocf-lets-encrypt, which uses HTTP instead of ACME if I understand correctly. That should still be using the ocfletsencrypt user as well.

jvperrin: Ok, yeah if most usages of this keep the existing user, then that seems totally reasonable.

PR #912, opened 2020-03-15
add ip6HostAddress to LDAP host schema

cg505:
As discussed on IRC: https://irclogs.ocf.berkeley.edu/rebuild/2020-03-14#1584221913-1584225880;

This will give us more flexibility moving forward with IP address, and will avoid nasty hacks in Puppet and elsewhere.

I will use [this script](https://i.fluffy.cc/DSTM0wWNmFVbMDCTDlLHpwhxP0k5sP23.html) to update existing records once this is merged. (Note that this depends on ocf/ocflib#206 to not break, but I've tested it with that change on dev-firestorm and it does work.)

### DIFF
--- a/modules/ocf_ldap/files/ocf.schema
+++ b/modules/ocf_ldap/files/ocf.schema
@@ -81,5 +81,5 @@ objectclass ( 1.3.6.1.4.1.41759.2.1.1
         DESC 'Attributes for OCF hosts'
         SUP ( ipHost $ ieee802Device $ puppetClient )
         AUXILIARY
-        MUST ( cn $ ipHostNumber $ type )
+        MUST ( cn $ type )
         MAY ( environment $ dnsCname $ dnsA ) )
--- a/modules/ocf_ldap/files/ocf.schema
+++ b/modules/ocf_ldap/files/ocf.schema
@@ -76,10 +76,26 @@ attributetype ( 1.3.6.1.4.1.41759.2.2.3
         SUBSTR caseIgnoreSubstringsMatch
         SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
 
+# based on ipHostNumber, 1.3.6.1.1.1.1.19
+attributetype ( 1.3.6.1.4.1.41759.2.2.4
+        NAME 'ip6HostNumber'
+        DESC 'IPv6 address, as represented by RFC1884 section 2.2.2'
+        EQUALITY caseIgnoreIA5Match
+        SYNTAX 1.3.6.1.4.1.1466.115.121.1.26{128} )
+
+# based on ipHost, 1.3.6.1.1.1.2.6
+objectclass ( 1.3.6.1.4.1.41759.2.1.2
+        NAME 'ip6Host'
+        DESC 'Host optionally supporting IPv6'
+        SUP top
+        AUXILIARY
+        MUST ( cn )
+        MAY ( ip6HostNumber $ manager $ description $ l ) )
+
 objectclass ( 1.3.6.1.4.1.41759.2.1.1
         NAME 'ocfDevice'
         DESC 'Attributes for OCF hosts'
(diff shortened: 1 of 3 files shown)


modules/ocf_ldap/manifests/init.pp
```diff
@@ -5,6 +5,8 @@
   # ldif files in /etc/ldap/slapd.d, slapd is the ldap server
   package { ['slapd', 'ocf-ldap-overlay', 'libarchive-zip-perl']:; }
   service { 'slapd':
+    ensure    => 'running',
```
jvperrin: Generally this isn't great to have in puppet since it leads towards a pattern where you need puppet to run on boot or something to set up a system. We don't have that, but we've specifically avoided setting this on almost all services in puppet since systemd should handle this. Did `slapd` crash or something that made this necessary?

modules/ocf_ldap/files/ocf.schema
```diff
@@ -76,10 +76,25 @@ attributetype ( 1.3.6.1.4.1.41759.2.2.3
...
+        SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+
+# based on ipHost, 1.3.6.1.1.1.2.6
+objectclass ( 1.3.6.1.4.1.41759.2.1.2
+        NAME 'ip6Host'
+        DESC 'Host supporting IPv6'
+        SUP top
+        AUXILIARY
+        MUST ( ip6HostNumber $ cn )
+        MAY ( manager $ description $ l ) )
```
jvperrin: Why do you need a separate object class for IPv6 hosts? Is this just to enforce the IP is added?

I think we'd just rely on `ldap-lint` for that since we rely on it for a bunch of other checks already that are more complicated than we want to encode into ldap rules.

modules/ocf_ldap/files/ocf.schema
```diff
@@ -76,10 +76,25 @@ attributetype ( 1.3.6.1.4.1.41759.2.2.3
...
+
 objectclass ( 1.3.6.1.4.1.41759.2.1.1
         NAME 'ocfDevice'
         DESC 'Attributes for OCF hosts'
-        SUP ( ipHost $ ieee802Device $ puppetClient )
+        SUP ( ipHost $ ip6Host $ ieee802Device $ puppetClient )
         AUXILIARY
-        MUST ( cn $ ipHostNumber $ type )
+        MUST ( cn $ type )
         MAY ( environment $ dnsCname $ dnsA ) )
```
jvperrin: Add `ipHostNumber` and `ip6HostNumber` here as optional attributes?

modules/ocf_ldap/manifests/init.pp
cg505: It wasn't running on dev-firestorm, I think. I guess `enable => true` should be sufficient?

modules/ocf_ldap/files/ocf.schema
cg505: There isn't a solid reason, but this matches the existing hierarchy of how we include `ipAddress`. It seemed weird to do `SUP ipHost` and then directly `MAY ip6address`.

modules/ocf_ldap/files/ocf.schema
cg505: `ipHostNumber` be mandatory IMO (which it currently is due to the hierarchy with `ipHost`). We could add `ip6HostAddress` here directly as optional, but it seems strange as described above.

kpengboy:
Is there an intended plan for what servers will receive different IPv6 addresses from their IPv4s and in what circumstances?

cg505:
There is no plan and there is no current need to have any machines with different v4 and v6 addresses.

cg505:
going to deploy tonight or tomorrow night and watch carefully

cg505:
Merged and deployed. I've added v6 addresses to all hosts.

PR #914, opened 2020-03-15
Add Folding @ Home to the desktops

ja5087:
While we're idle due to COVID-19 we might as well contribute what
we have.

### DIFF
--- a/modules/ocf/manifests/packages/fahclient.pp
+++ b/modules/ocf/manifests/packages/fahclient.pp
@@ -1,0 +1,18 @@
+class ocf::packages::fahclient {
+  # Folding @ Home
+  package {'fahclient':;}
+
+  file {
+    '/etc/fahclient/config.xml':
+      source  => 'puppet:///modules/ocf/packages/fahclient/config.xml',
+      require => Package['fahclient'],
+      owner   => 'fahclient',
+      mode    => '0664',
+      notify  => Service['FAHClient'];
+  }
+
+  service {
+    'FAHClient':
+      require => Package['fahclient'];
+  }
+}
--- a/modules/ocf/files/packages/fahclient/config.xml
+++ b/modules/ocf/files/packages/fahclient/config.xml
@@ -1,0 +1,15 @@
+<config>
+  <fold-anon v='true'/>
+
+  <team v='50188'/>
+
+  <gpu v='true'/>
+
+  <power v='full'/>
+
+  <slot id='0' type='CPU'/>
+  <slot id='1' type='GPU'/>
+
+  <allow>127.0.0.1</allow>
+  <web-allow>127.0.0.1</web-allow>
+</config>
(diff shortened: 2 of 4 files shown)


modules/ocf/manifests/packages/fahclient.pp
```diff
@@ -0,0 +1,17 @@
+class ocf::packages::fahclient {
+  # Folding @ Home
+  ocf::repackage {'fahclient':}
```
jvperrin: You shouldn't need `ocf::repackage` here as it's not backporting anything (nor is it backporting only on a particular distribution from what I can tell), just a simple `package { 'fahclient':; }` should do just fine.

modules/ocf/manifests/packages/fahclient.pp
```diff
@@ -0,0 +1,18 @@
...
+
+  file {
+    '/etc/fahclient/config.xml':
+      source  => 'puppet:///modules/ocf/packages/fahclient/config.xml',
+      require => Ocf::Repackage['fahclient'],
+      owner   => 'fahclient',
+      mode    => '0664',
+      notify  => Service['FAHClient'];
```
jvperrin: Instead of `require`/`notify` here if you prefer you could use arrows for all this:

```puppet
package { 'fahclient':; }
-> file { '/etc/fahclient/config.xml':
  source  => 'puppet:///modules/ocf/packages/fahclient/config.xml',
  owner   => 'fahclient',
  mode    => '0664',
}
~> service { 'FAHClient':; }
```

Some people prefer them, some don't, idk really

modules/ocf/manifests/packages/fahclient.pp
jvperrin: Oh also you might want to include the package for buster too, although there's no desktops on buster currently, but if it's to be used at all in the future.

redplanks:
There's also the wear and tear on the desktops to think about, if they're gonna be on all the time. (Though if they were only on during normal lab hours it wouldn't make too much difference)

nattofriends:
I wouldn't worry about that if it's only for a few weeks.

2020年3月15日(日) 1:29 Benjamin Zhang <notifications@github.com>:

modules/ocf/manifests/packages/fahclient.pp
ja5087: Yep, included it for both buster and stretch. Thanks!

ja5087:
Seems like sleeping at all is bad if we're going to do GPU folding according to https://foldingforum.org/viewtopic.php?nomobile=1&f=88&t=24737#p247284

It's only probably going to be for a few weeks, so maybe we could disable suspend for now and lower the intensity?

cg505:
> It's only probably going to be for a few weeks, so maybe we could disable suspend for now and lower the intensity?

Just disable suspend. We have the funds to replace parts due to wear and tear, but I don't anticipate big problems.

PR #915, opened 2020-03-15
ocf_mirrors: remove tanglu

cg505:
Their site is down and syncs are failing. I asked on Freenode if they're still active at all, we'll see if I get a response. (Their channel has been dead for more than a month.)


cg505:
yeah, we could leave rsync enabled if we want but I don't think it makes a lot of sense. I'll manually kill off the system services

PR #916, opened 2020-03-16
ocf-netboot: use fqdn for tftp

ja5087:
Somehow domain in resolv.conf is being set wrong during installation, causing the installation to try fetching from `dhcp.berkeley.edu` instead. I'd like to try and figure out why, but the cause is eluding me so far.


cg505:
shouldn't the domain get set... by the DHCP lease?

cg505:
@ja5087 just to be clear: /etc/resolv.conf has `search berkeley.edu`?

ja5087:
Yeah

PR #917, opened 2020-03-16
Add passkey and user to F@H

ja5087:
Apparently this will increase our gains

### DIFF
--- a/modules/ocf/manifests/packages/fahclient.pp
+++ b/modules/ocf/manifests/packages/fahclient.pp
@@ -2,13 +2,17 @@ class ocf::packages::fahclient {
   # Folding @ Home
   package {'fahclient':;}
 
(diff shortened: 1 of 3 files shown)


modules/ocf/manifests/packages/fahclient.pp
```diff
@@ -2,13 +2,17 @@
   # Folding @ Home
   package {'fahclient':;}
 
+  $fah_user = 'Open_Computing_Facility'
+  $fah_passkey = lookup('ocf_desktop::fah_passkey')
```
jvperrin: Ah, so for any new secrets you add, in order for the build to work, you should add a new placeholder secret to `hieradata/dummy_secrets.yaml`. That's why the build is currently broken on this review, even though I believe the changes would actually work fine.

PR #918, opened 2020-03-17
enable EFI netbooting

cg505:
This should allow us to install Debian on EFI over the network. The OCF-specific options are not added in yet.


modules/ocf_dhcp/files/netboot/ocf-netboot
```diff
@@ -99,6 +99,8 @@ for dist in "${dists[@]}"; do
         #
         # TODO: Figure out why domain is set wrong in /etc/resolv.conf during
         # installation, hence the FQDN for dhcp.
+	# TODO: Add these options to GRUB as well, so they're accessible
+	# on EFI boots.
```
cg505: sry

cg505:
I know I opened this PR, but I want to hold off merging because I'm not sure it's working and we should really add the menu options to GRUB for this to be at all useful.

PR #919, opened 2020-03-17
ocftv MPD configuration

### LINKED ISSUE #887: Write proper buster config for ocftv
@fydai and I worked on this one night when ocf-tv was moved to buster.
A list of changes we could come up with:

- [x] remove nodm dependency (nodm is deprecated)
- [x] add lightdm
- [ ] create group name autologin, add ocftv to group
- [x] install mpd + mpc
- [x] make mpd listen to anything (change mpd.conf)
- [ ] make sure x11 vnc service starts after (not before) lightdm
- [x] solve volume problems on mpd in some way (UPDATE: use pulseaudio sink, see mpd.conf)
- [ ] change ocf-tv script to remove hardcoded sink value

lightdm.conf:
```
[Seat:*]
pam-service=lightdm
pam-autologin-service=lightdm-autologin
autologin-user=ocftv
autologin-user-timeout=0
session-wrapper=/etc/X11/Xsession
greeter-session=lightdm-greeter
```

BernardZhao:
Should fix #887. However, I don't know if these changes don't cover all the things @encadyma mentions in the issue (yet). Could use some advice with:
*  create group name autologin, add ocftv to group (how should I go about doing this)
*  change ocf-tv script to remove hardcoded sink value (where is this and what is the alternative - since the sink value is also hardcoded in `mpd.conf`)

Also, will the systemd require make sure lightdm starts before X11?

### DIFF
--- a/modules/ocf_tv/files/mpd.conf
+++ b/modules/ocf_tv/files/mpd.conf
@@ -11,13 +11,7 @@ bind_to_address     "any"
 input {
   plugin            "curl"
 }
-decoder {
-  plugin            "hybrid_dsd"
-  enabled           "no"
-}
 audio_output {
-	type              "pulse"
-	name              "Sink #1"
+  type              "pulse"
+  name              "Sink #1"
 }
-filesystem_charset  "UTF-8"
-
--- a/modules/ocf_tv/files/mpd.conf
+++ b/modules/ocf_tv/files/mpd.conf
@@ -1,0 +1,23 @@
+music_directory     "/var/lib/mpd/music"
+playlist_directory  "/var/lib/mpd/playlists"
+db_file			        "/var/lib/mpd/tag_cache"
+log_file            "/var/log/mpd/mpd.log"
+pid_file            "/run/mpd/pid"
+state_file          "/var/lib/mpd/state"
+sticker_file        "/var/lib/mpd/sticker.sql"
+user				        "mpd"
+bind_to_address		  "any"
+
+input {
+  plugin            "curl"
+}
+decoder {
+  plugin            "hybrid_dsd"
+  enabled           "no"
+}
+audio_output {
+	type		          "pulse"
+	name		          "Sink #1"
+}
+filesystem_charset	"UTF-8"
+
--- a/modules/ocf_tv/files/mpd.conf
+++ b/modules/ocf_tv/files/mpd.conf
@@ -1,12 +1,12 @@
 music_directory     "/var/lib/mpd/music"
 playlist_directory  "/var/lib/mpd/playlists"
-db_file			        "/var/lib/mpd/tag_cache"
+db_file             "/var/lib/mpd/tag_cache"
 log_file            "/var/log/mpd/mpd.log"
 pid_file            "/run/mpd/pid"
 state_file          "/var/lib/mpd/state"
 sticker_file        "/var/lib/mpd/sticker.sql"
-user				        "mpd"
-bind_to_address		  "any"
+user                "mpd"
+bind_to_address     "any"
 
 input {
   plugin            "curl"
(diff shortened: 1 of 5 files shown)


modules/ocf_tv/files/mpd.conf
```diff
@@ -0,0 +1,17 @@
...
+user                "mpd"
+bind_to_address     "any"
+
+input {
+  plugin            "curl"
+}
+audio_output {
+  type              "pulse"
+  name              "Sink #1"
+}
```
encadyma: I wonder if the encoding field should still be kept so that it’s ensured it is operating in UTF-8

modules/ocf_tv/manifests/init.pp
```diff
@@ -18,7 +18,9 @@
       'ffmpeg',
       'feh',
       'i3',
-      'nodm',
+      'lightdm'
+      'mpc',
```
encadyma: [2020-03-17T09:15:13.098Z] Error: Could not parse for environment production: Syntax error at 'mpc' at /opt/jenkins/slave/workspace/puppet_PR-919/modules/ocf_tv/manifests/init.pp:22:7
?

jvperrin: Yeah it needs a comma added at the end of the previous line

BernardZhao: oops lol

modules/ocf_tv/manifests/init.pp
```diff
@@ -41,6 +43,12 @@
       group   => ocftv,
       require => User['ocftv'];
 
+    '/etc/mpd.conf':
+      source => 'puppet:///modules/ocf_tv/mpd.conf';
```
cg505: you should also require the mpd package here

modules/ocf_tv/manifests/init.pp
```diff
@@ -41,6 +43,12 @@
...
+
+    '/etc/lightdm/lightdm.conf':
+      source => 'puppet:///modules/ocf_tv/lightdm.conf';
```
cg505: same, require the lightdm package here

BernardZhao:
> Full stability from this point would mean ensuring ocftv is in the autologin group and lightdm.conf is present, so good next steps

Am I missing something to complete this? The configuration should handle all of this right

encadyma:
actually, if we're fixing #887 , there's a few items missing:

- create group name autologin, add ocftv to group (add `autologin` to groups attribute of ocftv)
- make sure x11 vnc service starts after (not before) lightdm
- change ocf-tv script to remove hardcoded sink value

BernardZhao:
Just clarifying about these points @encadyma:
* Shouldn't `autologin-user=ocftv` in `lightdm.conf` make ocftv autologin? Why is the autologin group needed?
* To make sure that x11 vnc service starts after lightdm, does that mean we want to start it as a systemd service after lightdm starts as a systemd service? I'm not too familiar with how to do this.
* I'll change the script after I confirm that this PR completely works. However, wouldn't we just change it to the new hardcoded sink value? Or is there some different approach you have in mind.

encadyma:
1. Adding `autologin-user=ocftv` is only the first part. For a user with a password (ocftv should have one), the user needs to be added to the `autologin` group as well for passwordless login [(source)](https://wiki.archlinux.org/index.php/LightDM#Enabling_autologin).
2. That's how it should work. I'm not completely confident in this either, but a peek at [this documentation](https://puppet.com/docs/puppet/latest/lang_relationships.html) might suggest declaring the service, which you can add a `require` property to?
3. There could be a way to fetch a listing of PulseAudio sinks, which Puppet could pull from and use to write the mpd.conf file. I agree that it's not necessary for this PR, and the current value seems generic enough that it'll be stable for future puppetting.

I don't want to call it complete because there might be some components I'm missing... we should test out this config once campus opens again.

cg505:
>does that mean we want to start it as a systemd service after lightdm starts as a systemd service

If you are trying to do this, Puppet "require" will not do what you want. You need to edit https://github.com/ocf/puppet/blob/master/modules/ocf_tv/files/x11vnc.service to be `After=lightdm.service` instead.

PR #923, opened 2020-03-28
add 1 day timeout to sync scripts

### LINKED ISSUE #531: Add timeout to mirror healthcheck script
Sometimes healthchecks freeze up (usually because of upstream network problems). Since they use systemd timers instead of cronjobs, this means future healthchecks won't run.

Luckily this triggers our alerting, so someone can go into fallingrocks and manually kill the healthcheck. But this should happen automatically.

We could set the timeout inside the Python script, but I think it would be better to hangle it within systemd. [This thread](https://unix.stackexchange.com/questions/227017/how-to-change-systemd-service-timeout-value) could be useful.

dkess:
Fixes #531.

### DIFF
--- a/modules/ocf_mirrors/templates/systemd/sync.service.erb
+++ b/modules/ocf_mirrors/templates/systemd/sync.service.erb
@@ -8,3 +8,5 @@ Type=simple
 Environment="<%= env %>=<%= val %>"
 <% end -%>
 ExecStart=<%= @exec_start %>
+# one day should be good enough for anyone
+TimeoutStartSec=86400


modules/ocf_mirrors/templates/systemd/sync.service.erb
```diff
@@ -8,3 +8,4 @@ Type=simple
 Environment="<%= env %>=<%= val %>"
 <% end -%>
 ExecStart=<%= @exec_start %>
+TimeoutStartSec=86400 # one day should be good enough for anyone
```
cg505: My understanding is that comments need to be on their own lines to be syntactically correct. Just put it above.

keur: nice. you are correct `system.syntax(7)`.

jvperrin:
(note that this is changing to having a timeout for both sync scripts and for the healthcheck scripts since they both use the same timers)

boomtow:
How was that did it work

PR #926, opened 2020-03-29
rsyncd performance improvements

Hexhu:
- By default rsyncd compresses files it serves. Skipping already-compressed files saves CPU and reduces latency.
- For a read-only rsync server, logging visitors' reverse DNS addresses is unnecessary.

### DIFF
--- a/modules/ocf_mirrors/files/rsyncd.conf
+++ b/modules/ocf_mirrors/files/rsyncd.conf
@@ -7,7 +7,7 @@ list = yes
 timeout = 60
 max connections = 100
 socket options = SO_KEEPALIVE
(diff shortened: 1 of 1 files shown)


cg505:
@Hexhu have you had a chance to test this?

Hexhu:
@cg505 I tested the "read only = yes" option, and in rsyncd log it will show `UNDETERMINED (<v4 or v6 address>)` instead of `<resolved rDNS>  (<v4 or v6 address>)`. I am not sure weather the performance increase will be significant though, but it should do no harm.

cg505:
cool!

cg505:
looks good in prod!

PR #928, opened 2020-04-03
Added logging to kubernetes

BernardZhao:
Putting what I set up earlier into writing. Check out [github.com/ocf/loki](https://github.com/ocf/loki) to see the Loki setup that goes along with this.


BernardZhao:
Just finished testing, it works perfectly after the tweaks. Thanks to @cg505 for letting me run puppet.

modules/ocf_kubernetes/files/logging.yaml
```diff
@@ -0,0 +1,181 @@
...
+    # <source>
+    #   @type tail
+    #   path /var/log/*.log
+    #   pos_file /var/log/fluentd/tmp/access.log.pos
+    #   tag foo.*
+
+    #   <parse>
+    #     @type json
+    #   </parse>
+    # </source>
```
cg505, line 118: can we clean this up

modules/ocf_kubernetes/files/logging.yaml
```diff
@@ -0,0 +1,181 @@
...
+          pod
+        </label>
+        flush_interval 10s
+        flush_at_shutdown true
+        buffer_chunk_limit 1m
+      </store>
+
+      # <store>
+      #   @type stdout
+      # </store>
```
cg505, line 180: yeah remove this too imo

modules/ocf_kubernetes/files/logging.yaml
BernardZhao: Ima keep it real with you chief idek what this is supposed to do

ja5087: It seems to be a comment? If it's not needed we can just remove it.

PR #932, opened 2020-04-09
add calibre to extrapackages

### LINKED ISSUE #809: Calibre
I want Calibre to be on the desktop computers at OCF. Calibre will allow me to read legally downloaded materials for my English classes. I'm sure there are plenty of other users who wish they could use Calibre and read epubs at the lab. Thanks.

ja5087:
closes #809 

I haven't really tested this yet, and it didn't work when I try to run it over X-Forwarding. Let's merge when we can test that it works either in lab or otherwise.


cg505:
Did we look into alternatives? There were concerns raised with Calibre, as it doesn't seem to be designed with multi-user security in mind.

ja5087:
I'm curious as to if that's something we could work around since it is pretty useful for converting between ebooks in addition to reading.

ja5087:
closing since it's gotten kind of stale

PR #933, opened 2020-04-09
Add ocfenforcer as user

PotatoParser:


PotatoParser:
Just tested on dev-whiteout; it seems to work!
Screenshot (258)

ja5087:
oh nice, good catch. weird that the script doesn't error at all

ja5087:
Ah, I see why the error didn't show up now:

```
mail.info:Apr  9 19:13:05 anthrax postfix/smtpd[24860]: NOQUEUE: reject: RCPT from dev-whiteout.ocf.berkeley.edu[2607:f140:8801::1:47]: 550 5.1.1 <ocfenforcer@ocf.berkeley.edu>: Recipient address rejected: User unknown; from=<> to=<ocfenforcer@ocf.berkeley.edu> proto=ESMTP helo=<dev-whiteout.ocf.berkeley.edu>
```

ja5087:
So I deleted a bunch of messages but from some investigation this will cause error reports to fail also (since ocfenforcer doesn't exist on `mail`). I think adding it as an alias for root to https://github.com/ocf/puppet/blob/master/modules/ocf_mail/files/site_ocf/aliases might work.

See the following log:

```
Apr  9 19:34:34 dev-whiteout postfix/smtp[32503]: 73AF480E9E: to=<root@ocf.berkeley.edu>, relay=smtp.ocf.berkeley.edu[2607:f140:8801::1:35]:25, delay=5.9, delays=5.9/0.01/0.01/0.04, dsn=5.1.7, status=bounced (host smtp.ocf.berkeley.edu[2607:f140:8801::1:35] said: 550 5.1.7 <ocfenforcer@ocf.berkeley.edu>: Sender address rejected: User unknown (in reply to RCPT TO command))
```

PotatoParser:
Will look into it!

modules/ocf_printhost/manifests/cups.pp
```diff
@@ -58,7 +58,16 @@
         content => epp('ocf_printhost/cups/ppd/m806.ppd.epp', { 'double' => true });
     }
   }
-
+  file { '/var/spool/cups/':
+    ensure => 'directory',
+    owner  => 'ocfenforcer',
+    mode   => '0700';
+  }
+  file { '/usr/lib/cups/backend/tea4cups':
```
cg505, line 66: pls leave a comment saying why this is necessary

PotatoParser:
tea4cups saves files based on the owner (switched from root -> ocfenforcer)

cg505:
sorry, I meant like add a comment to the file

modules/ocf_printhost/manifests/cups.pp
```diff
@@ -58,7 +58,16 @@
         content => epp('ocf_printhost/cups/ppd/m806.ppd.epp', { 'double' => true });
     }
   }
-
+  file { '/var/spool/cups/':
```
ja5087, line 61: You should modify the mount point a few lines below instead (add `uid=ocfenforcer`)

PotatoParser: Thanks for the heads up!

modules/ocf_printhost/manifests/cups.pp
```diff
@@ -59,9 +59,16 @@
     }
   }
 
+  #Tea4cups saves files based on its owner
+  file { '/usr/lib/cups/backend/tea4cups':
+    ensure => 'file',
+    owner  => 'ocfenforcer',
+    mode   => '0700';
```
jvperrin, line 66: Probably also good to include `require => Package['cups-tea4cups']` here since this is normally provided by that package so you'd want it to exist first before modifying it's permissions

ja5087:
I think the code looks good but let's wait to merge until we get lab access again

PR #934, opened 2020-04-09
Delete old staffvm definitions

ja5087:
Don't merge until we've shutdown all of them and sent images out properly


jvperrin:
Please remove mine too (`fireball`), I've moved off it now. It's still running but not doing anything currently after the upgrade to buster.

raymondnumbergenerator:
By the way, I'm still using my staff vm (fukushima) so I'd like to keep it.

PR #935, opened 2020-04-10
bump websocket timeouts in k8s lb

cg505:
I still need to test this, but this should keep the mastodon websocket from timing out.


cg505:
This is tested and it works. Please review.

PR #944, opened 2020-04-16
staffvm: release nyx

abizer:
so long, and thanks for all the fish

### DIFF
--- a/hieradata/nodes/nyx.yaml
+++ b/hieradata/nodes/nyx.yaml
@@ -1,5 +1,0 @@
-classes:
-    - ocf_ocfweb::dev_config
-#   - ocf::packages::docker
-
-owner: abizer


dkess:
o7

glee-:
o7

jvperrin:
There's probably some delay after that too, since the tests are currently getting host lists from puppetdb, so it would need to be absent from there to actually not run

cg505:
hmm, I guess we may not have removed any hosts since octocatalogdiff has been rolled out

cg505:
I ran `sudo puppet node deactivate nyx.ocf.berkeley.edu` and it seemed to fix the problem.

jvperrin:
Nice, good to know that works 👍

BernardZhao:
Is there anything left to be done here?

PR #946, opened 2020-04-23
Setup alpine and devuan mirror

MandiZhao:
Tested on dev-fallingrock, helped by jaw !!!

### DIFF
--- a/modules/ocf_mirrors/files/project/alpine/sync-archive
+++ b/modules/ocf_mirrors/files/project/alpine/sync-archive
@@ -1,0 +1,3 @@
+#!/bin/sh -eu
+/usr/local/bin/rsync-no-vanished -az --delete --safe-links \
+	rsync://rsync.alpinelinux.org/alpine/ /opt/mirrors/ftp/alpine
--- a/modules/ocf_mirrors/files/project/alpine/sync-archive
+++ b/modules/ocf_mirrors/files/project/alpine/sync-archive
@@ -1,3 +1,3 @@
 #!/bin/sh -eu
-/usr/local/bin/rsync-no-vanished -az --delete --safe-links \
+/usr/local/bin/rsync-no-vanished -auzH --delete --delete-after --delay-updates --safe-links \
 	rsync://rsync.alpinelinux.org/alpine/ /opt/mirrors/ftp/alpine
--- a/modules/ocf_mirrors/manifests/projects/alpine.pp
+++ b/modules/ocf_mirrors/manifests/projects/alpine.pp
@@ -1,0 +1,24 @@
+class ocf_mirrors::projects::alpine {
+  file {
+    '/opt/mirrors/project/alpine':
+      ensure  => directory,
+      source  => 'puppet:///modules/ocf_mirrors/project/alpine',
+      owner   => mirrors,
+      group   => mirrors,
+      mode    => '0755',
+      recurse => true;
+  }
+
+  ocf_mirrors::monitoring { 'alpine':
+    type          => 'unix_timestamp',
+    upstream_host => 'mirrors.alpinelinux.org',
+    ts_path       => 'TIME',
+  }
+
+  ocf_mirrors::timer {
+    'alpine':
+      exec_start => 'opt/mirrors/project/alpine/sync-archive',
+      hour       => '0/3',
+      minute     => '22';
+  }
+}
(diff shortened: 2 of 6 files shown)


modules/ocf_mirrors/manifests/projects/alpine.pp
```diff
@@ -0,0 +1,25 @@
...
+    type          => 'unix_timestamp',
+    upstream_host => 'mirrors.alpinelinux.org',
+    ts_path       => 'TIME',
+  }
+
+  ocf_mirrors::timer {
+    'alpine':
+      exec_start => '/opt/mirrors/project/alpine/sync-archive',
+      hour       => '0/3',
+      minute     => '22',
```
cg505, line 22: This will run every 3 hours, but they seem to want us to sync every hour, which is probably fine. We can use `'*'` instead of `'0/3'`for `hour`. Also, centos already syncs at minute 22, so maybe we can choose another value random value for that.

ja5087:
I forgot to mention but we should also add an entry in rsyncd.conf to enable rsync access.

https://github.com/ocf/puppet/blob/master/modules/ocf_mirrors/files/rsyncd.conf

modules/ocf_mirrors/files/project/alpine/sync-archive
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -eu
+/usr/local/bin/rsync-no-vanished -az --delete --safe-links \
```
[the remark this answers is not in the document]
MandiZhao: Gotcha, updating to add `--delete-after --delay-updates` and -uH` for `--update --hard-links`

modules/ocf_mirrors/manifests/projects/alpine.pp
MandiZhao: Ah Mb, didn't know minute should be random :( In that case we should probably change here too https://github.com/ocf/puppet/blob/4e37cbedd228d89bc2f32234dbb4fc54114faa9d/modules/ocf_mirrors/manifests/projects/archlinux.pp#L21

MandiZhao:
@ja5087 @cg505  Making a new commit to reflect your suggestions above (rsync flags; upstream_host; timer hour/minute;  rsyncd.conf) Could jaw maybe grant me testing access to dev-fallingrocks again to test these changes? Thank u both for bearing with me! xdd

ja5087:
Done. Be careful though since the disk will fill up pretty fast.

MandiZhao:
> Done. Be careful though since the disk will fill up pretty fast.

`sudo: PAM authentication error: User not known to the underlying authentication module` :( 
I changed init.pp to comment out other projects, and changed dev-fallingrocks.yaml like u suggested earlier

MandiZhao:
I think it's working: 
```
mannndy@dev-fallingrocks:/opt/mirrors/ftp/alpine$ sudo puppet-trigger -t
Info: Using configured environment 'mannndy'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Retrieving locales
Info: Loading facts
Info: Caching catalog for dev-fallingrocks.ocf.berkeley.edu
Info: Applying configuration version '1587654438'
Warning: Augeas[/etc/postfix/master.cf](provider=augeas): Loading failed for one or more files, see debug for /augeas//error output
Notice: Applied catalog in 92.62 seconds
```

MandiZhao:
Ok messed up a little, should've create a new branch to setup devuan mirror, but committed here again, I'll make this branch about both devuan and alpine then? I'm just trying to replicate everything to do devuan.

modules/ocf_mirrors/manifests/projects/devuan.pp
```diff
@@ -9,11 +9,7 @@
       recurse => true;
   }
 
-  ocf_mirrors::monitoring { 'devuan':
-    type          => 'unix_timestep',
-    upstream_host => 'https://www.devuan.org/#download',
-    ts_path       => 'TIME';
-  }
+# Looks like devuan doesn't have a timestamp file, so omitted ocf_mirrors::monitoring
```
cg505, line 12: it doesn't need to block this PR, but we may want to contact them and ask for this to be added. we've done this with other distros in the past IIRC.

modules/ocf_mirrors/manifests/projects/devuan.pp
```diff
@@ -0,0 +1,20 @@
...
+      group   => mirrors,
+      mode    => '0755',
+      recurse => true;
+  }
+
+# Looks like devuan doesn't have a timestamp file, so omitted ocf_mirrors::monitoring
+
+  ocf_mirrors::timer { 'devuan':
+    exec_start => '/opt/mirrors/project/devuan/sync-archive',
+    hour       => '*',
```
cg505, line 16: we can probably do `0/6` for the as they recommend in https://files.devuan.org/MIRRORS.txt

PR #948, opened 2020-04-23
add ripgrep to extrapackages

nikhiljha:
I made this bar graph to illustrate why we need ripgrep:

IMG_0022

As we can see, the bar for rg is bigger, so we need it.

### DIFF
--- a/modules/ocf/manifests/extrapackages.pp
+++ b/modules/ocf/manifests/extrapackages.pp
@@ -190,6 +190,7 @@ class ocf::extrapackages {
     'r10k',
     'rails',
     'rbenv',
+    'ripgrep',
     'ruby-build',
     'ruby-dev',
     'ruby-fcgi',


nattofriends:
ripgrep is not available in debian before buster.

nikhiljha:
lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 9.12 (stretch)
Release:	9.12
Codename:	stretch
```

nikhiljha:
I'll let this PR sit until we get buster then :(

cg505:
you could move the package into the if statement lower down so it will install on any buster machines

nikhiljha:
@cg505 none of the hosts are on buster?

cg505:
we can probably upgrade supernova

dkess:
Pretty sure supernova/tsunami/death/vampires all have to be upgraded at the same time (since they all share nfs and could be symlinking to packages) but @jvperrin please confirm

ja5087:
Is backporting it an option?

cg505:
> Is backporting it an option?

we would have to manually, it's not in stretch-backports

ja5087:
Ok, I threw the latest .deb onto our apt repo (for stretch and buster). Not sure if there's any downside to doing that in general but this should work now.

jvperrin:
> Pretty sure supernova/tsunami/death/vampires all have to be upgraded at the same time (since they all share nfs and could be symlinking to packages) but @jvperrin please confirm

Sorta, tsunami/death/vampires all have to be done like this since they're all public and people get confused if they're editing files and testing with one PHP version or something but it runs with a different one. That's the rationale I've heard anyway so I've always done these upgrades in lockstep.

For supernova it can be done whenever, it's for staff only and I don't think it has any services on it now? (used to have redis before it moved to gridlock)

PR #952, opened 2020-04-29
quarantine -> njha

nikhiljha:
it is I, quarantine


hieradata/nodes/quarantine.yaml
```diff
@@ -1,5 +1,4 @@
 # Enable outside operator users to login
 ocf::auth::ulogin: [["%{lookup('owner')}", 'ALL']]
```
cg505, line 2: you can delete this as well

PR #953, opened 2020-04-29
Matrix Federation 

expedited:
## Changes 
* Added JSON response to be served in puppet file
* Updated Apache configuration to specify path for puppet file and created an alias for this path. 

Addresses https://github.com/ocf/matrix/issues/1

### DIFF
--- a/modules/ocf_www/manifests/site/www.pp
+++ b/modules/ocf_www/manifests/site/www.pp
@@ -26,6 +26,11 @@ class ocf_www::site::www {
   include ocf_www::mod::php
   include ocf_www::mod::suexec
 
+  file {
+    '/var/www/html/.well-known/matrix/server':
+      source => 'puppet:///modules/ocf_www/matrix-server';
+  }
+
   # TODO: dev-death should add a robots.txt disallowing everything
   apache::vhost { 'www':
     servername          => 'www.ocf.berkeley.edu',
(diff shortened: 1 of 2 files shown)


modules/ocf_www/manifests/site/www.pp
```diff
@@ -56,6 +69,8 @@
           '%{REQUEST_URI} !^/icons/',
           # ...hide ocfweb metrics
           '%{REQUEST_URI} !^/metrics',
+    # ...and not if it's the matrix well-known file
+    '%{REQUEST_URI} !^/\.well-known/matrix',
```
nikhiljha, line 73: The indentation looked right when you were in vim but now it doesn't? Estoy confundido.

modules/ocf_www/manifests/site/www.pp
```diff
@@ -42,6 +49,12 @@
     request_headers     => ['set X-Forwarded-Proto https'],
     proxy_preserve_host => true,
 
+    aliases             => [
+      { alias => '/.well-known/matrix/server',
```
cg505, line 53: should probably put alias on its own line to be consistent with the other sections

modules/ocf_www/manifests/site/www.pp
cg505: ^^

expedited: I think the linter did that? I wonder if I try to fix it if it will do it again

expedited:
Thanks for handling @cg505 :)

cg505:
it didn't work tho :(

PR #954, opened 2020-04-30
Enable cgiirc support

jessicadong101:
Allows us to add WEBIRC support to lounge

### DIFF
--- a/hieradata/nodes/dev-flood.yaml
+++ b/hieradata/nodes/dev-flood.yaml
@@ -1,3 +1,2 @@
-ocf::auth::usudo: ['jessicadong', 'joyceguo']
 classes:
     - ocf_irc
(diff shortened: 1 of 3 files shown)


hieradata/nodes/dev-flood.yaml
```diff
@@ -1,2 +1,3 @@
+ocf::auth::usudo: ['jessicadong', 'joyceguo']
```
ja5087, line 1: Please remove this line

modules/ocf_irc/templates/inspircd.conf.erb
```diff
@@ -118,6 +119,12 @@
 #          port="3306"
 #          id="irc">
 
+<cgiirc opernotice="yes">
+
+<cgihost type="webirc"
+	 mask="169.229.226.0/24"
```
ja5087, line 125: Can you align this with the rest of the block?

PR #959, opened 2020-05-11
setup bookstack domain

nikhiljha:
This is open to suggestions, but I think `help` is fine.


ja5087:
You will also need the corresponding PR in ocf/dns

nikhiljha:
@ja5087 thanks

Looks like that's root only `ldap_modify: Insufficient access (50)` though :(

Gotta wait for someone with root to hop in discord again asdhghad

PR #962, opened 2020-05-12
log outbound request on death

ja5087:
this was approved by the board of directors on 2020-05-04

god bless if you run dmesg without `-l` after this

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -29,6 +29,8 @@ internal_zone_exceptions:
     - dev-death
 
 # IPv4 and IPv6 src ranges for types of servers
+ocf_ipv4_mask: 169.229.226.0/24
+ocf_ipv6_mask: 2607:f140:8801::/48
 internal_zone_range_4: 169.229.226.3-169.229.226.90
 desktop_src_range_4: 169.229.226.100-169.229.226.139
 staffvm_src_range_4: 169.229.226.200-169.229.226.252
--- a/modules/ocf_www/files/iptables-log.conf
+++ b/modules/ocf_www/files/iptables-log.conf
@@ -1,0 +1,4 @@
+# Send iptables logs to a separate file
+# Must come before others to prevent other rules from recording first
+:msg,contains,"[iptables] " /var/log/iptables.log
+& ~
--- a/modules/ocf_www/files/iptables-log.conf
+++ b/modules/ocf_www/files/iptables-log.conf
@@ -1,5 +1,5 @@
 # Send iptables logs to a separate file
 # Must come before others to prevent other rules from recording first
 :msg, contains, "[iptables-outbound] " /var/log/iptables.log
-# Clear all messages already recorded
+# Prevent other rules from matching messages already recorded
 & ~
--- a/modules/ocf_www/files/iptables-log.conf
+++ b/modules/ocf_www/files/iptables-log.conf
@@ -1,4 +1,5 @@
 # Send iptables logs to a separate file
 # Must come before others to prevent other rules from recording first
-:msg,contains,"[iptables] " /var/log/iptables.log
+:msg, contains, "[iptables-outbound] " /var/log/iptables.log
+# Clear all messages already recorded
 & ~
--- a/modules/ocf_www/files/iptables-logrotate
+++ b/modules/ocf_www/files/iptables-logrotate
@@ -4,5 +4,5 @@
 	missingok
 	rotate 5
 	compress
-	compresscmd xz
+	compresscmd gzip
 }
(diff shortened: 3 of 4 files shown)


modules/ocf_www/manifests/logging.pp
```diff
@@ -7,6 +7,29 @@
     require => Package['nfs-kernel-server'],
   }
 
+  # log outbound requests
+  firewall_multi {
+    '101 log outbound request on death (v4)':
```
ja5087, line 12: this is not nice (I'd like to use firewall46) but ip6tables will complain if you give it ipv4 and vice versa

modules/ocf_www/files/iptables-log.conf
```diff
@@ -0,0 +1,4 @@
+# Send iptables logs to a separate file
+# Must come before others to prevent other rules from recording first
+:msg,contains,"[iptables] " /var/log/iptables.log
```
jvperrin, line 3: It's be nice if you could make this more specific since I could see other messages containing `[iptables] ` potentially, but that's not a big deal if some logs are sent to the wrong file. Looks like you could have a combination of this one plus another filter that it starts with `kernel` and that might work? That or maybe doing `[iptables] IN=` here to be more specific.

Maybe add some spaces in between commas to make it a bit more readable?

Might also need log rotation on this, not sure if that'll be done for you or not. Probably can wait and see on that one though.

modules/ocf_www/manifests/logging.pp
jvperrin: This could be good as a code comment, but anyway thanks for mentioning it here, that was useful

kpengboy: You can at least use a `default` resource to cut down on repetition though.

modules/ocf_www/manifests/logging.pp
```diff
@@ -7,6 +7,29 @@
...
+        provider    => 'iptables',
+        chain       => 'PUPPET-OUTPUT',
+        outiface    => '! lo',
+        jump        => 'LOG',
+        destination => '! 169.229.226.0/24',
```
kpengboy, line 17: Here and below: is it possible to get this netblock from Hiera or the `ocf::networking` class, instead of duplicating it?

modules/ocf_www/files/iptables-log.conf
```diff
@@ -0,0 +1,5 @@
+# Send iptables logs to a separate file
+# Must come before others to prevent other rules from recording first
+:msg, contains, "[iptables-outbound] " /var/log/iptables.log
+# Clear all messages already recorded
```
kpengboy, line 4: Perhaps "prevent other rules from matching messages already recorded"?

Baisang:
Project idea: also send these logs to Kafka, and deploy some stream processing application so that if you see a bunch of outbound requests to the same domain/host/whatever within X minute window, fire an email alert

modules/ocf_www/manifests/logging.pp
cg505: @kpengboy++, or map on an array with the two destinations

ja5087: `default` is really neat, thanks

modules/ocf_www/manifests/logging.pp
ja5087: I don't think we have this stored and while it's in the networking class I'm not sure how to get it out of there

modules/ocf_www/files/iptables-logrotate
```diff
@@ -0,0 +1,8 @@
+# Logrotate configuration for logging iptables
+/var/log/iptables-outbound.log {
+	weekly
+	missingok
+	rotate 5
+	compress
+	compresscmd xz
```
cg505, line 7: is there a reason we're using xz here when everything else uses gz?

hieradata/common.yaml
```diff
@@ -29,6 +29,8 @@ internal_zone_exceptions:
     - dev-death
 
 # IPv4 and IPv6 src ranges for types of servers
+ocf_ipv4_mask: 169.229.226.0/24
+ocf_ipv6_mask: 2607:f140:8801::/48
```
cg505, line 33: nice, we should probably look for other places where we can use this (in another PR)

modules/ocf_www/files/iptables-logrotate
ja5087: I figured we'd get slightly better compression since the logs could blow up. I just tested it and xz was ~30% better than gz which compressed at 11x, although it took a lot more time. I can change it back.

PR #963, opened 2020-05-12
Rename devuan to devuan-cd

ja5087:
I talked to the devuan maintainers and it seems we'd be really useful as a package mirror, so that's definitely a todo for the future.


ja5087:
> 
> 
> You'll probably also have to do a `sudo mv /opt/mirrors/project/devuan /opt/mirrors/project/devuan-cd` unless you want to resync it all? Anyway, seems reasonable

Yeah, that's running in a tmux session right now

PR #967, opened 2020-05-13
change spamassasin scoring and whitelists

cg505:
this needs to be tested still. ~also, internal networks should be updated~


daradib:
The required_score was set aggressively because false negatives are especially painful for mail forwarding: lowering our mail reputation and generating bounces when the provider we forward to (e.g., Gmail) rejects our mail. Since we always reject, at least the sender is aware of false positives. I bet there's enough spam sent to domains hosted by us that even 1% false negatives would result in us sending more spam than ham.

daradib:
By the way, the armresearch.com whitelist can be dropped. We used to pay them for a yearly subscription to MessageSniffer/SNF4SA.

cg505:
> The required_score was set aggressively because false negatives are especially painful for mail forwarding

Reverted that change for now.

PR #968, opened 2020-05-16
mysql buster prep

ja5087:
these variables were deprecated


ja5087:
The three options were deprecated (and promptly reintroduced in 10.4.3 for compatibility, lol): 
https://mariadb.com/kb/en/changes-improvements-in-mariadb-103/

PR #970, opened 2020-05-17
Revert "log outbound request on death"

ja5087:
Reverts ocf/puppet#962

I'm reverting this for now since it filled up the disk on death in no time, and the syslog redirection doesn't seem to work as expected.


cg505:
you hate to see it

PR #972, opened 2020-05-18
install molly-guard everywhere

nikhiljha:
rebooting isn't too common, this intercepts `reboot` and `shutdown` when executed over ssh and asks you to type the hostname to confirm

prevents accidental reboots of the hypervisors, which take a while


ja5087:
there may or may not have been an incident that triggered this

PR #973, opened 2020-05-20
Separate monitoring and sync unit files

ja5087:
Sometimes we see hung connections and this can stall the healthcheck for a long time plus fire alerts. It's probably better to retry a couple of times and then fail.


dkess:
This isn't the best approach.

If the script is hanging, make the timeout shorter and let it fail. The monitoring around this should be robust, and can tolerate the script failing every once in a while.

The real problem is that the timeout duration is shared between the monitoring scripts and the healthcheck scripts, which doesn't make any sense. It's more worthwhile to fix that instead.

ja5087:
Sure, being shared is a problem, but this change will also make the healthcheck script fail after 5 retries of 10 seconds, which sounds similar to the behavior you want? i.e. it won't hang anymore, and it'll also retrya few times before failing hard.

dkess:
Sure, but this doesn't belong in the script. It's simpler and more elegant to let it fail and let systemd handle retrying.

Put another way, the script should fulfill its one purpose of trying to look at mirrors, and the supervisor (systemd) should fulfill its purpose of scheduling execution, determining if it's taking too long (with systemd timeouts), and retrying if necessary. The monitoring service also fulfills its purpose by looking at the state and determining to alert if something is wrong.

The systemd timer naturally does "retrying" by executing again on whatever schedule we set. I don't think additional retries are necessary, but if you really want it, systemd can handle it, see https://stackoverflow.com/a/39284869.

This also means you never have to second-guess if the retry logic in the code is correct. It looks fine now, but a change in the future by someone who doesn't have this background could cause the script to hang again. It's better to be explicit.

dkess:
I'll also add that this change makes the logging less explicit, since each retry is a silent failure that never gets logged. I know the intention was to retry on timeout, but it looks like you're actually retrying on any kind of request failure and throwing away whatever the error was.

You could fix this by printing the exception, but it's better to not have to worry about any of this. Instead, let exceptions propagate naturally, cause a crash, and show up in the journal.

ja5087:
Ok, that sounds pretty convincing to me. I've updated the code to use systemd to manage the execution instead.

modules/ocf_mirrors/manifests/timer.pp
```diff
@@ -12,9 +12,19 @@
   # Any meaningful second precision requires AccuracySec to be set
   $on_calendar = "${year}-${month}-${day} ${hour}:${minute}:00"
 
+  if !($type in ['sync', 'monitor']) {
+    fail('Type parameter of ocf_mirrors:timer must be either sync or monitor')
+  }
+
+  if $type == 'sync' {
```
dkess, line 19: Not a huge deal, but if statements in Puppet are a code smell. The Puppet resource for sync jobs and monitoring jobs should be separate since they do totally different things.

Issue #548, opened 2019-03-10
Debian unattended (automatic) upgrades


dkess:
We currently upgrade all of our systems manually, pretty much whenever we feel like it or whenever we see a Debian security alert. This process is [documented](https://www.ocf.berkeley.edu/docs/staff/procedures/installing-updates/), but it's nonetheless annoying for us to deal with.

Luckily, Debian supports [Unattended Upgrades](https://wiki.debian.org/UnattendedUpgrades), which we should be using instead. That docs page explains how to set it up, including systemd timers. If necessary, we can also include this in our monitoring and/or rootspam.

jvperrin:
One thing we should be careful of here is around upgrading certain packages automatically like marathon, mesos, maybe docker. If we could just upgrade for security releases I think that would be best, but otherwise full automatic upgrades seems decent as long as we can spread the upgrades out enough and detect breakage before it brings down important services.

dkess:
Removing the "good first issue" label since I think this would be best tackled by someone with root or a staffvm. Still definitely worth looking into. I agree that this should start out by only doing security updates.

jvperrin:
It looks like we can configure `unattended-upgrades` to only do security upgrades (or only upgrades from Debian upstream for instance), which would help since most of our third-party upgrades like docker, marathon, jenkins, etc. are more likely to break anyway.

PR #974, opened 2020-05-20
Automatic security updates

ja5087:
Nice: running apt dater everytime a DSA comes out at 3am
Nicer: magic cron man updates them for us

EDIT: I didn't see #548 but this closes that


ethanhs:
All hail the magic cron man \o/

PR #979, opened 2020-07-02
mirrors: Sync devuan packages

ja5087:
There are a few more steps here such as rewrite rules per: https://pkgmaster.devuan.org/devuan_mirror_walkthrough.txt


boomtow:
The change should happen only if the fix isn't there I'm looking for it

fydai:
lgtm, once the checks pass

PR #980, opened 2020-07-04
staffvm: add ocfweb dev config to trojan

Hexhu:


Hexhu:
Edited my environment on `lightning` instead and that worked, closing.

PR #984, opened 2020-07-14
Enable ptrace protection for browser sandbox

daradib:
Only allow ptrace from a parent process to its children or via
CAP_SYS_PTRACE.

To verify sandbox status for Brave, Chrome, Firefox see
brave://sandbox, chrome://sandbox, about:support, respectively.

Also describe disadvantages of enabling unprivileged user namespaces.
Distributions like Debian currently disable unprivileged user namespaces
by default to decrease the kernel attack surface for local privilege
escalation. See [Debian bug #898446][898446]. If kept disabled, Brave 1.2+ and
Chrome will still enforce namespace sandboxing via their setuid-root
helper executable. See brave/brave-browser#3420 and
brave/brave-browser#6247. Firefox does not include a setuid-root binary,
however, so unprivileged user namespaces are useful to have for
defence-in-depth, but not critical. See
<https://www.morbo.org/2018/05/linux-sandboxing-improvements-in_10.html>.

[898446]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=898446

### DIFF
--- a/modules/ocf/manifests/browser_sandbox.pp
+++ b/modules/ocf/manifests/browser_sandbox.pp
@@ -1,0 +1,22 @@
+class ocf::browser_sandbox {
+  # Change kernel settings for the sandbox used by Brave, Chrome, and Firefox.
+  # Verify sandbox status at brave://sandbox, chrome://sandbox, about:support,
+  # respectively.
+  sysctl {
+    # Distributions like Debian currently disable unprivileged user namespaces
+    # by default to decrease the kernel attack surface for local privilege
+    # escalation. See Debian bug #898446. If kept disabled, Brave 1.2+ and
+    # Chrome will still enforce namespace sandboxing via their setuid-root
+    # helper executable. See brave/brave-browser#3420 and
+    # brave/brave-browser#6247. Firefox does not include a setuid-root binary,
+    # however, so unprivileged user namespaces are useful to have for
+    # defense in depth, but not critical. See
+    # <https://www.morbo.org/2018/05/linux-sandboxing-improvements-in_10.html>.
+    'kernel.unprivileged_userns_clone':
+      value => '1';
+    # Enable ptrace protection. Only allow ptrace from a parent process to its
+    # children or via CAP_SYS_PTRACE.
+    'kernel.yama.ptrace_scope':
+      value => '1';
+  }
+}
--- a/modules/ocf/manifests/packages/brave.pp
+++ b/modules/ocf/manifests/packages/brave.pp
@@ -1,5 +1,5 @@
 class ocf::packages::brave {
-  include ocf::userns
+  include ocf::browser_sandbox
 
   $browser_homepage = lookup('browser_homepage')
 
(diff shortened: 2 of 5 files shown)


daradib:
With `sysctl kernel.unprivileged_userns_clone=0`:

|||
-- | --
Layer 1 Sandbox | SUID
PID namespaces | Yes
Network namespaces | Yes
Seccomp-BPF sandbox | Yes
Seccomp-BPF sandbox supports TSYNC | Yes
Ptrace Protection with Yama LSM (Broker) | Yes
Ptrace Protection with Yama LSM (Non-broker) | Yes

With `sysctl kernel.unprivileged_userns_clone=1`:

|||
-- | --
Layer 1 Sandbox | Namespace
PID namespaces | Yes
Network namespaces | Yes
Seccomp-BPF sandbox | Yes
Seccomp-BPF sandbox supports TSYNC | Yes
Ptrace Protection with Yama LSM (Broker) | Yes
Ptrace Protection with Yama LSM (Non-broker) | No

Ptrace Protection with Yama LSM (Non-broker) is expected to be no when using unprivileged user namespaces. See https://chromium.googlesource.com/chromium/src/+/d267fd9917cb1c7494a8067ea3c6f1831bb37e78%5E%21/

> Also, when the Chrome renderers are sandboxed with user namespaces,
any process in the parent namespace with the same UID is able to
ptrace the renderer. However, the chrome://sandbox page displays Yama
LSM as enforcing. This makes it clear that Yama LSM is not protecting
the renderer processes from ptrace by adding "Ptrace Protection with
Yama LSM (Non-broker)" to the webpage.
> [...]
>   // If there is no ptrace protection anywhere, that is bad.
>   // If there is no ptrace protection for nonbroker processes because of the
>   // user namespace sandbox, that is fine and we display as medium.

modules/ocf/manifests/browser_sandbox.pp
```diff
@@ -0,0 +1,22 @@
...
+    # helper executable. See brave/brave-browser#3420 and
+    # brave/brave-browser#6247. Firefox does not include a setuid-root binary,
+    # however, so unprivileged user namespaces are useful to have for
+    # defense in depth, but not critical. See
+    # <https://www.morbo.org/2018/05/linux-sandboxing-improvements-in_10.html>.
+    'kernel.unprivileged_userns_clone':
+      value => '1';
+    # Enable ptrace protection. Only allow ptrace from a parent process to its
+    # children or via CAP_SYS_PTRACE.
+    'kernel.yama.ptrace_scope':
```
kpengboy, line 19: Interesting, I had thought this was already set.

Wonder if we should enable this for all machines?

daradib: Looks like Debian disabled it in [bug #712740](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=712740). Yeah, we could enable it for all machines as Ubuntu does. It obviously won't prevent rogue processes from messing with user files including Kerberos tickets, but the only downside is losing the ability to gdb/strace processes running in the background without root (breaks `gdb --attach` and `strace -p`) and possibly some edge cases like wine. I'll merge this for now and in the meantime maybe one of you root folks can test `sysctl kernel.yama.ptrace_scope=1` on the login server, app host, etc.

PR #986, opened 2020-07-24
Lock down kernel on buster

daradib:
Disable some kernel features: module loading after boot, kexec,
Berkeley Packet Filter (BPF).

Also install cloud kernel image on VMs which removes some hardware
support. Benefits: slightly faster boot and reduced attack surface.


ja5087:
The changes look good to me but I would like to canary this on a VM and physical host first before merging to avoid potential issues. Thanks!

tzhu-yelp:
Why would we want to disable BPF?

nattofriends:
Why would we want to disable BPF?

daradib:
> The changes look good to me but I would like to canary this on a VM and physical host first before merging to avoid potential issues. Thanks!

Most definitely. I also replaced the lockdown package with hardening-runtime. However if the performance impact of the kernel command line options is too great we'll need to revert it.

> Why would we want to disable BPF?

To clarify, this doesn't disable BPF entirely, it disables [unprivileged bpf()](https://lwn.net/Articles/660331/) that has been a source of privilege escalation vulnerabilities, e.g., CVE-2020-8835, CVE-2017-16995, CVE-2016-4557. netfilter and seccomp-bpf predate it and shouldn't be affected.

kpengboy:
Do you know if we currently rely on any after-boot module loading à la `request_module` or similar?

singingtelegram:
Can we take another look at this? Would be good to test on a few dev- VMs.

PR #990, opened 2020-07-28
make ssl restart postgres

ja5087:
Previously we were finding that new certs weren't being applied. Old postgres means we need to manually restart to apply them.

You'd get a run log like this:

```
Info: Class[Ocf::Ssl::Default]: Scheduling refresh of Class[Postgresql::Server]
Info: Class[Postgresql::Server]: Scheduling refresh of Anchor[postgresql::server::start]
Info: Class[Postgresql::Server]: Scheduling refresh of Anchor[postgresql::server::end]
Notice: /Stage[main]/Postgresql::Server/Anchor[postgresql::server::start]: Triggered 'refresh' from 1 event
Notice: /Stage[main]/Postgresql::Server/Anchor[postgresql::server::end]: Triggered 'refresh' from 1 event
```

but the service does not restart:

```
● postgresql.service - PostgreSQL RDBMS
   Loaded: loaded (/lib/systemd/system/postgresql.service; enabled; vendor preset: enabled)
   Active: active (exited) since Mon 2020-07-27 19:12:21 PDT; 5h 18min ago
```

push for octocatalog diff

### DIFF
--- a/modules/ocf_postgres/manifests/init.pp
+++ b/modules/ocf_postgres/manifests/init.pp
@@ -32,12 +32,6 @@ class ocf_postgres {
       };
   }
 
-  service { 'postgresql':
-    ensure  => 'running',
-    enable  => true,
-    require => Class['postgresql::server'],
-  }
-
   Class['Ocf::Ssl::Default'] ~> Class['Postgresql::Server']
   Class['Ocf::Ssl::Default'] ~> Service['postgresql']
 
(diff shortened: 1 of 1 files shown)


jvperrin:
Hmmm, it'd be nice if `octocatalog-diff` ran alongside `make test` so it still shows output even if it fails some linting check or whatever. Not sure why that's not the case currently since I think that should be possible, but maybe I just didn't notice before?

ja5087:
Ideally we'd have the validate puppet manifest checks serve as a smoke test before `octocatalog-diff`, but where linting is more advice and shouldn't fail builds.

dkess:
Recommend adding a comment here saying this can be deleted after upgrading to Buster.

PR #991, opened 2020-07-28
remove node requirement from post action

ja5087:
This may fix the double executor problem

For anyone reading this in the future declaring `agent {label:'slave'}` at the top of the file allocates one executor on the `slave` and runs it there. `node` is a similar declaration, except `node` is a stage level declaration and `agent` is pipeline-level.

Unfortunately, it seems that every time you redeclare this, a new executor is allocated, and all our builds with this post {} block (all of them basically) actually require TWO executors, with one being allocated near the end to send us some IRC notifications.

image

This is a huge problem when we do lots of concurrent builds e.g. full org-wide Jenkins rebuilds since it leads to executor starvation no matter how many executors you allocate, as all of them will be waiting for another executor to be free.

Removing this requirement causes the post actions to stay on the same executor (as per `agent`).

image

Unfortunately this is going to be tricky since Jenkins uses the pipeline defined in each branch but we fix this moving forward and backfill into other branches at some point.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -113,7 +113,7 @@ pipeline {
       emailNotification()
     }
     always {
-      node(label: 'slave') {
+      {
         ircNotification()
       }
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -113,9 +113,7 @@ pipeline {
       emailNotification()
     }
     always {
-      {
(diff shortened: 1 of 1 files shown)


ja5087:
Looks like that did it, I might mass commit this change on all the branches so we can prevent the lockups again

PR #996, opened 2020-08-02
Disable network logging on tsunami

ja5087:
The investigation in rt#9297 has concluded

There's probably a bit of manual cleanup to do to stop it but this won't reapply it.


dkess:
Why not keep netlog around? Could be useful in the future.

ja5087:
I think we could probably roll it out sitewide since so far it's <50MB total in logs on death. At least for tsunami/supernova I'd like to get some consensus from bod before we start logging usage.

PR #1000, opened 2020-08-04
Upgrade pre-commit hooks and add shellcheck

jvperrin:
I was told to do it!
> \<jvperrin-slack\> !8ball steal puppet PR #1000
> \<create\> jvperrin-slack: signs point to yes

Anyway, I noticed that while [utils has shellcheck](https://github.com/ocf/utils/blob/48f1453e84c854a9543f6c2e0d03abaa12c98d04/.pre-commit-config.yaml#L39-L43), this repo does not, so I decided to try enabling it here too. I also tried out the shellcheck hook from https://github.com/gruntwork-io/pre-commit but I liked the simpler one from detailyang instead, especially since newer shellcheck versions do things like check for shebangs automatically instead of doing anything [more custom for that](https://github.com/gruntwork-io/pre-commit/blob/4373c1e6b21003c566c466744ee0d904f6ae81dd/hooks/shellcheck.sh#L39-L47).

I'd like to try and test most of these scripts before just merging this, but I think these all should be safe changes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,13 +3,6 @@ repos:
     rev: v2.1.0
     hooks:
     -   id: puppet-validate
-        # TODO: Remove this version pinning. The problem here seems to be that
-        # the Debian-packaged versions of the semantic_puppet gem
-        # (ruby-semantic-puppet package) cause some kind of error with locale
-        # loading for puppet validation. A bug probably needs to be opened with
-        # the Debian package maintainers or with semantic_puppet depending on
-        # who is actually causing the error.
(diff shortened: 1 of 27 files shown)


ja5087:
Looks like this just needs the file to be deleted on your fork then we can merge this. Sorry I let it get out of sync.

jvperrin:
I believe those failures are all on master and have been fixed in this branch ^

If I run puppet on the `production` environment on `destruction` it fails, but on my branch it runs to completion (well with an error due to the `riot-web` package that's got #1060 open for it, but it does actually compile the catalog at least)

PR #1015, opened 2020-08-24
Adds immutable handling to ocf::privatefile for /opt/passwords

mcint:


mcint:
This could be generalized to support all of the chattr attributes.

> The letters 'aAcCdDeijsStTu' select the new attributes for the files: append only (a), no atime updates (A),  compressed  (c),  no  copy on write (C), no dump (d), synchronous directory updates (D), extent format (e), immutable (i), data journalling (j), project hierarchy (P), secure deletion (s), synchronous updates (S), no tail-merging (t), top of directory hierarchy (T),  and undeletable (u).

However, the current way exposes the minimum of functionality needed in a self-explaining way.

mcint:
The initial chattr should also fail, causing logs noise.

ja5087:
Can you explain what you mean by the log noise? I tried to run `chattr -/+i test` a couple of times and they didn't seem to produce any unusual return codes

ja5087 (approved):

PR #1019, opened 2020-08-28
Upgrade kubernetes to 1.16.14

ja5087:
Next we need to upgrade the actual cluster, but this should be the config changes?

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -69,7 +69,7 @@ kubernetes::kube_api_advertise_address: "%{::ipaddress}"
 kubernetes::api_server_count: 3
 kubernetes::install_dashboard: false
 kubernetes::controller_address: 169.229.226.75:6443
-kubernetes::nginx_version: '0.21.0'
+kubernetes::nginx_version: '0.34.1'
 kubernetes::worker_nodes:
     - pandemic
     - riptide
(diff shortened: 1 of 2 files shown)


hieradata/common.yaml
```diff
@@ -47,10 +47,10 @@ apache::default_vhost: false
 apache::server_tokens: Minor
 
 # Kubernetes config
-kubernetes::kubernetes_version: 1.15.6
-kubernetes::kubernetes_package_version: 1.15.6-00
-kubernetes::docker_version: 18.06.2~ce~3-0~debian # 18.06.02 is the highest supported version to work with kubernetes
-kubernetes::etcd_version: 3.3.10
+kubernetes::kubernetes_version: 1.16.14
+kubernetes::kubernetes_package_version: 1.16.14-00
+kubernetes::docker_version: 18.06.3~ce~3-0~debian # make sure to verify this with the kubernetes release notes
```
dkess, line 52: docker version 18.06.2 is recommended (source: https://v1-16.docs.kubernetes.io/docs/setup/production-environment/container-runtimes/)

ja5087: Sounds good, I was looking at https://v1-16.docs.kubernetes.io/docs/setup/release/notes/

`The current list is 1.13.1, 17.03, 17.06, 17.09, 18.06, 18.09. (#72823, #72831)`

and just assumed that would work

dkess: Ah it's probably fine then ¯\\_(ツ)_/¯

PR #1021, opened 2020-08-30
Install helm from apt repo

ja5087:
this is the recommended way now


nikhiljha (approved):

ja5087:
thanks, it's annoying that I can't delete branches from gh mobile

nikhiljha:
that's weird, my gh mobile has the delete branch button just like on desktop

Issue #906, opened 2020-03-13
Augeas fails on /etc/postfix/master.cf


cg505:
On buster hosts, puppet shows this warning
```
Warning: Augeas[/etc/postfix/master.cf](provider=augeas): Loading failed for one or more files, see debug for /augeas//error output
```
and the resource in question is not deployed.

This is due to an upstream bug. Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=930888 and augeas-lenses: https://github.com/hercules-team/augeas/pull/636.

jvperrin:
Looks like `1.12.0-2` was released today, so whenever that is available for us in `sid` we could backport to make this warning go away.

Thanks for commenting on the debian bug, looks like that helped get things moving on a fix!

cg505:
I am surprised that they do not consider this a bug in Buster (since default augeaus and conf in buster doesn't work).

cg505:
Hopefully they just backport this to buster actually, I guess that could happen.

jvperrin:
They might do, but it's also pretty quick and easy for us to backport it if they don't end up doing that at least.

jvperrin:
I backported the newer version (`1.12.0-2`) for buster and tested installing it on a buster VM, but it did not remove the warning. I also noticed while doing this that augeas appears to have been separated out into a module (https://forge.puppet.com/puppetlabs/augeas_core) as of puppet 6.x. I tried using that but that also made no change in the warning printed, even alongside the backported augeas version installed.

Upon further inspection, it seems the `puppet-agent` package is providing the lens that needs patching (at `/opt/puppetlabs/puppet/share/augeas/lenses/dist/postfix_master.aug`) instead of the `augeas-lenses` package (at `/usr/share/augeas/lenses/dist/postfix_master.aug`), so we can probably just uninstall `augeas-lenses` across all hosts, and I think it's just a holdover from when we used non-puppetlabs puppet packages.

Unfortunately, it'll be harder to patch that lens file and we should probably submit a puppet-agent bug to puppetlabs and try and get them to patch it in some future release.

cg505:
ref https://bugs.debian.org/958503

jvperrin:
Sure, but I think that won't actually fix our issue, as we're not using that package for augeas lenses, we're using the ones bundled within `puppet-agent` instead, if that makes sense

jvperrin:
[...]

PR #1022, opened 2020-08-30
mirrors: exclude temporarily some files for puppetlabs

MCer4294967296:
They have some files with permission 600 and that is causing our rsync
to fail. The files do not show up in their https mirror site, and are
named "work.xxxxx.yyyyyyyyyy" where x is [0-9] and y is [0-9a-zA-Z],
which seem like temp files.


jvperrin:
FWIW, I've also had some success reaching out to https://tickets.puppetlabs.com to raise any mirror or packaging issues in the past, and that might be useful for both this and for #906 for instance.

ja5087:
https://tickets.puppetlabs.com/plugins/servlet/mobile#issue/CPR-759

PR #1026, opened 2020-09-10
Enable slightly prettier Jenkins header

nattofriends:
The current one is dull and depressing.

https://github.com/jenkinsci/jenkins/pull/4463

Also moved the -Xmx to the front


jvperrin (approved):
Looks good to me.

I wonder why we still have `-Djava.awt.headless=true -Djava.net.preferIPv4Stack=true`, seems like IPv6 should be fine by now?

nattofriends:
I think this needs a restart of the service to apply

PR #1028, opened 2020-09-16
Add keyserver.dobrev.eu and keyserver.dobrev.it in SKS membership file

mclueppers:
I'm looking for ways to expand my peering list and provide more reliable SKS keyserver service. All my peers but one terminated their nodes.


ja5087 (approved):
Thanks!

PR #1029, opened 2020-09-16
Add websocket support to app vhosts

ja5087:
Context: HKN is deploying a new app that requires websockets. This is also a good exercise to see how we can support additional reverse-proxying options and enable more types of applications.

```
3381 # hkn-tutoring.eecs.berkeley.edu (user hkn)
3382 server {
3383     listen 443;
3384     listen [::]:443;
3385     server_name "hkn-tutoring.eecs.berkeley.edu";
3386 
3387     location /.well-known/ {
3388         alias /var/lib/lets-encrypt/.well-known/;
3389     }
3390 
3391     location / {
3392 
3393 
3394                 proxy_pass http://unix:/srv/apps/hkn/tutoring.sock;
3395                 proxy_set_header Host $host;
3396 
3397             proxy_set_header X-Forwarded-For $remote_addr;
3398             proxy_set_header X-Real-IP $remote_addr;
3399 
3400     }
3401 
3402 
3403 
3404     location /socket.io/ {
3405             proxy_http_version 1.1;
3406             proxy_set_header Upgrade $http_upgrade;
3407             proxy_set_header Connection "Upgrade";
3408             proxy_set_header Host $host;
3409             proxy_pass http://unix:/srv/apps/hkn/tutoring.sock;
3410     }
3411 
3412 
3413 
3414     access_log /var/log/nginx/vhost_access.log vhost;
3415 
3416 
3417         ssl on;
3418         ssl_certificate /etc/ssl/apphost/hkn-tutoring.eecs.berkeley.edu.bundle;
3419         ssl_certificate_key /etc/ssl/lets-encrypt/le-vhost.key;
3420         add_header Strict-Transport-Security "max-age=31536000";
3421 
3422 }

```

### DIFF
--- a/modules/ocf_www/files/build-vhosts
+++ b/modules/ocf_www/files/build-vhosts
@@ -2,6 +2,7 @@
 import argparse
 import grp
 import os.path
+import re
 import shutil
 import subprocess
 import sys
@@ -76,6 +77,7 @@ class VirtualHost(namedtuple('VirtualHost', (
         'ssl',
         'bind_type',
         'bind_dest',
+        'additional_rules',
 ))):
     """A logical representation of a virtual host directive in one of
     the webserver configs.
@@ -91,6 +93,11 @@ class VirtualHost(namedtuple('VirtualHost', (
     bind_type can be any of 'socket', 'docroot', or 'redirect'.
     bind_dest is then respectively the user socket name, the website
     root directory relative to public_html, or URL to redirect to.
+
+    additional_rules should be a dictionary of
+    additional rules that you'll have to add logic to handle in this
+    class. Currently the only supported one is 'ws', which is meant
+    to add an additional WebSockets path for socket-based app vhosts.
     """
     @property
     def contact_email(self):
@@ -133,6 +140,13 @@ class VirtualHost(namedtuple('VirtualHost', (
         assert self.is_redirect
         return self.bind_dest
 
+    @property
+    def websocket_locations(self):
+        if self.bind_type == 'socket' and 'ws' in self.additional_rules:
+            return self.additional_rules['ws']
+        else:
+            return []
+
     def dev_alias(self, dev_config=False):
         return '{}.{}.ocf.berkeley.edu'.format(
             self.fqdn.replace('.', '-'),
@@ -147,7 +161,7 @@ def report(*args, **kwargs):
 
 
 def build_config(src_vhosts, template, dev_config=False):
-    vhosts = set()
+    vhosts = list()
 
     for domain, vhost in src_vhosts.items():
         user = vhost['username']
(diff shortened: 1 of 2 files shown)


modules/ocf_www/files/build-vhosts
```diff
@@ -147,7 +162,7 @@ def report(*args, **kwargs):
 
 
 def build_config(src_vhosts, template, dev_config=False):
-    vhosts = set()
+    vhosts = list()
```
jvperrin, line 165: I'm curious, why does this need changing from a `set` to a `list`? Are there duplicates now?

modules/ocf_www/files/build-vhosts
```diff
@@ -133,6 +140,14 @@ class VirtualHost(namedtuple('VirtualHost', (
         assert self.is_redirect
         return self.bind_dest
 
+    @property
+    def has_websockets(self):
+        return self.bind_type == 'socket' and len(self.additional_rules.get('ws', [])) > 0
+
+    @property
+    def websocket_locations(self):
+        return self.additional_rules['ws']
```
jvperrin, line 149: Instead of this and using `has_websockets` for a conditional, why not just have `websocket_locations` return an empty list if it's not applicable and then nothing will show up in the template?

So something like:
```python
@property
def websocket_locations(self):
    return self.additional_rules.get('ws', [])
```
(without the `has_websockets` at all)
Or, if you do want to keep the socket check, maybe this (could perhaps be written in a nicer way):
```
@property
def websocket_locations(self):
    if self.bind_type == 'socket' and 'ws' in self.additional_rules:
        return self.additional_rules['ws']
    else:
        return []
```

And then in the template, you can remove the `has_websockets` check entirely:
```
{% for ws_location in vhost.websocket_locations %}
location /{{ws_location}} {
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "Upgrade";
        proxy_set_header Host $host;
        proxy_pass http://unix:{{vhost.socket}};
}
{% endfor %}
```

modules/ocf_www/files/build-vhosts
```diff
@@ -161,6 +176,13 @@ def build_config(src_vhosts, template, dev_config=False):
             bind_type = 'docroot'
         bind_dest = vhost[bind_type]
 
+        # Match and process supported flags
+        additional_rules = {}
+        for flag in vhost['flags']:
+            ws_match = re.match(r'ws\=(.+)', flag)
+            if ws_match:
+                additional_rules.setdefault('ws', []).append(ws_match.group(1))
```
jvperrin, line 184: Heh, this is very clever. Might be more readable on two lines though

jvperrin (approved):

modules/ocf_www/files/build-vhosts
ja5087: Now that we added a dict the tuple isn't hashable. From a cursory reading (also I tested it), it doesn't seem to cause problems although we might be in hot water if we accidentally add duplicate declarations. Not exactly sure how nginx will react either.

modules/ocf_www/files/build-vhosts
ja5087: It didn't look nice when I tried the two line thing so I just left it

modules/ocf_www/files/build-vhosts
jvperrin: Ah I see, didn't realize that was a restriction of sets. Anyway, having this be a list seems reasonable to me, was just curious why this was needed.

Could maybe store the flags or metadata in another list or something instead as a workaround, but that might be over the top.

PR #1030, opened 2020-09-17
Use iptables-legacy when possible

fydai:
This fixes needing to manually change iptables-nft to iptables-legacy on buster.

### DIFF
--- a/modules/ocf/manifests/firewall/chains.pp
+++ b/modules/ocf/manifests/firewall/chains.pp
@@ -1,4 +1,6 @@
 class ocf::firewall::chains {
+  require ocf::firewall::use_legacy
+
   # We put all puppet input/output firewall rules in dedicated chains managed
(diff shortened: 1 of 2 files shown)


ja5087 (approved):
thanks this has been stupid everytime I provision a new buster VM

jvperrin (approved):
Looks good, maybe link in something like https://tickets.puppetlabs.com/plugins/servlet/mobile#issue/MODULES-8543 to track for when this can be removed again?

fydai:
So apparently following that link they merged a PR to fix the issues:
https://github.com/puppetlabs/puppetlabs-firewall/pull/911

So I guess I will close this PR and create a new one to bump the `puppetlabs_firewall` version

cg505:
I believe this is still needed, it looks like the puppetlabs-firewall PR just suppresses the error.

PR #1031, opened 2020-09-18
Allow Puppeting Kustomizations

nikhiljha:
this is probably not the right way to do it but


jvperrin (commented):
Maybe make this an argument to `ocf_kubernetes::apply` instead since it only differs by 2 characters that could be a passed in arg that defaults to `-f` or something

nikhiljha:
didn't do that originally because I thought they functioned differently- they do not

nikhiljha:
closed in favor of https://github.com/ocf/kubernetes

PR #1032, opened 2020-09-19
add chrome to extrapackages

ja5087:
blocker for a some apphosts and sites nowadays that require chrome to run their fancy hack build scripts


ethanhs:
>require chrome to run their fancy hack build scripts

Oh god why

PR #1033, opened 2020-09-19
Upgrade corruption to buster

ja5087:
Pray we do not need to adjust anything else.

Other things to do:
- [x] Backport nvidia-persistenced from snapshots


ethanhs (approved):
🙏🙏

PR #1036, opened 2020-09-21
change slurm.conf template for new version

ja5087:
this one works the old one doesn't


ja5087:
going to merge this since it's working and the previous one is broken

PR #1042, opened 2020-10-10
Redirect to userdirs with numbers

ja5087:
This needs to be patched to allow numbers

### DIFF
--- a/modules/ocf_www/manifests/site/shorturl.pp
+++ b/modules/ocf_www/manifests/site/shorturl.pp
@@ -140,7 +140,7 @@ class ocf_www::site::shorturl {
       {rewrite_rule => '^/zoom$ https://berkeley.zoom.us/j/92953066816 [R]'},
 
       # Otherwise, send a temporary redirect to the appropriate userdir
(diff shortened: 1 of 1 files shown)


modules/ocf_www/manifests/site/shorturl.pp
```diff
@@ -140,7 +140,7 @@
       {rewrite_rule => '^/zoom$ https://berkeley.zoom.us/j/92953066816 [R]'},
 
       # Otherwise, send a temporary redirect to the appropriate userdir
-      {rewrite_rule => '^/~?([a-z]{3,16}(?:/.*)?)$ https://www.ocf.berkeley.edu/~$1 [R]'},
+      {rewrite_rule => '^/~?([a-z0-9]{3,16}(?:/.*)?)$ https://www.ocf.berkeley.edu/~$1 [R]'},
```
kpengboy, line 143: Was `[a-z][a-z0-9]{2,15}` considered?

jvperrin (approved):
Seems good enough, I don't think the regex needs to be that precise since it's the fallback redirect of sorts after all the more specific ones are already tried

cg505:
https://github.com/ocf/ocflib/blob/master/ocflib/account/utils.py#L43 needs updated as well

PR #1044, opened 2020-10-16
Add extrapackages to hpc controller

ja5087:
We recommend stuff like virtualenv for HPC (https://www.ocf.berkeley.edu/docs/services/hpc/), but we don't have it installed. Also, it is a login server so we should keep the packages consistent.

### DIFF
--- a/modules/ocf/manifests/extrapackages.pp
+++ b/modules/ocf/manifests/extrapackages.pp
@@ -217,6 +217,7 @@ class ocf::extrapackages {
     'unison',
     'vagrant',
     'valgrind',
+    'virtualenv',
     'weechat',
     'xvnc4viewer',
     'zlib1g-dev',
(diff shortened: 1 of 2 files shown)


jvperrin (commented):
I think this should already be getting installed as a dependency of `tox`, so I'm surprised it's not present: https://github.com/ocf/puppet/blob/217b3198769ad810d47a28c059d65af716c30df0/modules/ocf/manifests/extrapackages.pp#L215

Seems like `tox` isn't on segfault either though, which is strange. Does it not include `ocf::extrapackages`? I can't actually find a place that it would be included currently

jvperrin:
Yeah it's not included in the list of affected hosts here either, so I think including `ocf::extrapackages` into `ocf_hpc::controller` would be the way to go here instead.

ja5087:
Thanks for the catch, I should have noticed that octocatalog-diff didn't actually show a change on segfault.

jvperrin (approved):

PR #1047, opened 2020-11-07
Give berkeleytime root on hozer-55

ja5087:
This will be their test VM until the situation is more permanent


ja5087:
Going to merge since we discussed this setup and octocatalog-diff isn't screaming

PR #1051, opened 2020-11-23
add powershell to packages

nikhiljha:
the best shell

### DIFF
--- a/modules/ocf/manifests/packages/powershell/apt.pp
+++ b/modules/ocf/manifests/packages/powershell/apt.pp
@@ -4,12 +4,21 @@ class ocf::packages::powershell::apt {
     id     => 'BC528686B50D79E339D3721CEB3E94ADBE1229CF',
     source => 'https://packages.microsoft.com/keys/microsoft.asc';
   }
-
-  apt::source { 'powershell':
-    architecture => 'amd64',
-    location     => '[arch=amd64,arm64,armhf] https://packages.microsoft.com/debian/10/prod',
-    release      => 'buster',
-    repos        => 'main',
-    require      => Apt::Key['powershell repo key'],
+  if $::lsbdistcodename == 'buster' {
+    apt::source { 'powershell':
+      architecture => 'amd64',
+      location     => '[arch=amd64,arm64,armhf] https://packages.microsoft.com/debian/10/prod',
+      release      => 'buster',
+      repos        => 'main',
+      require      => Apt::Key['powershell repo key'],
+    }
+  } else {
+    apt::source { 'powershell':
+      architecture => 'amd64',
+      location     => '[arch=amd64] https://packages.microsoft.com/repos/microsoft-debian-stretch-prod',
+      release      => 'stretch',
+      repos        => 'main',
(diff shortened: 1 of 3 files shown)


modules/ocf/manifests/packages/powershell/apt.pp
```diff
@@ -0,0 +1,15 @@
...
+class ocf::packages::powershell::apt {
+  apt::key { 'powershell repo key':
+    id     => 'BC528686B50D79E339D3721CEB3E94ADBE1229CF',
+    source => 'https://packages.microsoft.com/keys/microsoft.asc';
+  }
+
+  apt::source { 'powershell':
+    architecture => 'amd64',
+    location     => '[arch=amd64,arm64,armhf] https://packages.microsoft.com/debian/10/prod',
+    release      => 'buster',
```
ja5087, line 11: hmm, we still have a number of machines on stretch. consider gating this like https://github.com/ocf/puppet/blob/3567ba2e0dfe17535ab1c5e11ce7f39873150173/modules/ocf_irc/manifests/ircd.pp#L22 ?

modules/ocf/manifests/packages/powershell/apt.pp
```diff
@@ -0,0 +1,15 @@
+# Include PowerShell apt repo
+class ocf::packages::powershell::apt {
+  apt::key { 'powershell repo key':
+    id     => 'BC528686B50D79E339D3721CEB3E94ADBE1229CF',
+    source => 'https://packages.microsoft.com/keys/microsoft.asc';
+  }
+
+  apt::source { 'powershell':
+    architecture => 'amd64',
+    location     => '[arch=amd64,arm64,armhf] https://packages.microsoft.com/debian/10/prod',
```
ja5087, line 10: arm64,armhf?

modules/ocf/manifests/packages/powershell/apt.pp
nikhiljha: this supports stretch but there doesn't seem to be a variable that goes:
buster => 10
stretch => 9

I guess I can use an if statement again?

modules/ocf/manifests/packages/powershell/apt.pp
nikhiljha: future-proofing for the ocf on arm64 switch (I saw another package.pp file like this so I guess it works)

modules/ocf/manifests/packages/powershell/apt.pp
ja5087: I guess if it works it works

jvperrin: Like the `lsbmajdistrelease` fact, or?

ja5087 (approved):

modules/ocf/manifests/packages/powershell/apt.pp
nikhiljha: turns out that won't work anyways, because the repo url for stretch is just entirely different

they changed it for buster :P

PR #1053, opened 2020-11-26
Kubernetes!

fydai:
This is a rewrite of the existing kubernetes setup from puppetlabs-kubernetes to manually configured services and configuration files.


modules/ocf_kube/templates/etcd.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: v1
```
ja5087, line 1: If this is templated by our convention it should end in .erb

modules/ocf_kube/manifests/master.pp
```diff
@@ -0,0 +1,218 @@
...
+      content_path => "${certs_dir}/front-proxy-ca.key";
+
+    # front-proxy client
+    '/etc/kubernetes/pki/front-proxy-client.crt':
+      content_path => "${certs_dir}/front-proxy-client.crt";
+    '/etc/kubernetes/pki/front-proxy-client.key':
+      content_path => "${certs_dir}/front-proxy-client.key";
+  }
+
+  # TODO: disable swap
```
ja5087, line 217: `swapoff` based on `swapon -s` should work?

hieradata/common.yaml
```diff
@@ -79,6 +79,33 @@ kubernetes::master_nodes:
...
+kubectl_package_version: 1.19.4-00
+
+# Kubernetes config (the new setup)
+kube::kubernetes_version: 1.19.4
+# get from https://packages.cloud.google.com/apt/dists/kubernetes-xenial/main/binary-amd64/Packages
+kube::kubernetes_package_version: 1.19.4-00
+# get from https://github.com/kubernetes/k8s.io/
+kube::etcd_version: 3.4.13-0
+# run gen-kube-crt in ocf/utils after editing
+kube::master_nodes:
```
ja5087, line 92: this is going to be super nitpicky but the kubernetes project along with everyone else is transitioning away from this terminology

hieradata/common.yaml
```diff
@@ -79,6 +79,33 @@ kubernetes::master_nodes:
     - autocrat
     - deadlock
 
+# Kubectl version
+kubectl_package_version: 1.19.4-00
+
+# Kubernetes config (the new setup)
+kube::kubernetes_version: 1.19.4
```
ja5087, line 86: This is probably to avoid conflict with the current class name(?). In any case is there a plan for transitioning from the current setup to this?

ja5087 (commented):
wow, this is really impressive in scale

modules/ocf_kube/manifests/master.pp
fydai: You would also want to comment it out in /etc/fstab, otherwise you need to wait for puppet to run, disabling swap before kubelet will even start. That's certainly possible to do in puppet, but I'll leave it for a later PR.

hieradata/common.yaml
fydai: Migrate all existing kubernetes services to the new cluster, and you can just delete all the code for the old configuration.

modules/ocf_kube/templates/etcd.yaml
fydai: It would by the convention, but then you lose syntax highlighting on github and in your editor. I'm willing to change it, but I personally think it's fine as long as people know it's an erb template.

hieradata/common.yaml
fydai: I'll change it.

modules/ocf_kube/templates/etcd.yaml
ja5087: imo it is not clear at all that it's an erb template outside of us in the know right now. You can configure syntax highlighting for different file extensions in your editor and there are cases where that will be wrong due to special characters that overlap between erb and the target language. YAML is pretty simple though so there might not be one in this case.

ja5087 (approved):
Before it went down I think this was working pretty well. Merging it will let other people work on this easier too and avoid us having things on different puppet envs (`cooperc2`).

modules/ocf_kube/templates/etcd.yaml
cg505: ja++

nikhiljha:
🤷

PR #1061, opened 2020-12-12
[WIP] beginnings of bullseye support

nikhiljha:


modules/ocf_apt/files/distributions
```diff
@@ -23,3 +23,9 @@ Codename: buster-backports
 Architectures: amd64 i386 source
 Components: main
 SignWith: D72A0AF4
+
+# The year is 20XX, 32-bit machines no longer exist.
+Codename: bullseye
```
jvperrin, line 28: I'd suggest also adding `bullseye-backports` to this file (in another block), since at some point we'll probably want backports from `sid`/`experimental` or whatever release is next. There's always a few things to backport!

nikhiljha: I read on the Debian wiki that `bullseye-backports` isn't a thing until release(?)

jvperrin: It isn't, but this is for our internal repo (and not the mirror), so this is just to contain any packages we backport from sid or wherever, so it can be created at any time.

jvperrin (approved):

PR #1065, opened 2021-01-30
add newstaff, edit joinstaff

encadyma:


ja5087 (approved):
now that we have /apple and /newstaff, /joinstaff is pretty confusing since we used to use it to link people to the presentation, but it will do for now.

PR #1069, opened 2021-02-10
Remove extra lets encrypt intermediate, use TLSv1.2 only

jvperrin:
I've tested this by running it on `dev-death`, before running it `curl -vi --resolve decal.ocf.berkeley.edu:443:169.229.226.37 https://decal.ocf.berkeley.edu` fails with a cert error (`curl: (60) SSL certificate problem: unable to get local issuer certificate`), and afterwards it succeeds. Essentially all that's happening here is that `SSLCertificateChainFile` is removed and `SSLCertificateFile` is unchanged.

I also tested it on `vampires`, which was a bit more complex since it was building bundle files using the intermediate. I removed that but it was using the `mtime` of the bundle file to determine whether to reload nginx or not to get cert updates, so I replaced that with just using the `mtime` of the cert files themselves compared against the last cert-related reload in `/tmp/last-nginx-reload`. Seems to work well, but suggestions welcome as this is a little bit hacky.

While I was here and looking at SSL reports (from https://www.ssllabs.com/ssltest/index.html), it mentioned that TLSv1 and TLSv1.1 are old and most clients don't need them any more, so I've removed them here and updated the ciphers as per https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28default.29
 (linked in the comment just above the ciphers). When running a SSL Labs check this seems to work just fine, but the versions of nginx we have aren't new enough to support `TLSv1.3` yet unfortunately (needs `1.13.0`+ and we have `1.10.3` unless we use backports).


jvperrin:
Bleh, the failures are due to the same prometheus module issue from before that isn't compatible with hosts that are using facter 4.x I believe. I've upgraded `puppet-prometheus` in this branch too for a better fix than the hotfix I made to the production environment before, but all the failures are on non-production environments that I don't want to switch back from, so I'm going to ignore them for now.

jvperrin:
Tested the new version on `dementors`, since it's the most involved use of prometheus, and it seems to have gone fine (puppet succeeded, https://prometheus.ocf.berkeley.edu seems to be working fine).

ethanhs:
Sadly it looks like this didn't quite work :/

PR #1070, opened 2021-02-13
added all the current staff with root

RonitNath:
@encadyma

### DIFF
--- a/modules/ocf_irc/manifests/services.pp
+++ b/modules/ocf_irc/manifests/services.pp
@@ -7,7 +7,7 @@ class ocf_irc::services {
   }
 
   $irc_creds = lookup('irc_creds')
-  $root_nicks = ['waf', 'nattofriends', 'ckuehl', 'jvperrin', 'mattmcal', 'abizer', 'dkessler', 'cooperc', 'jaw']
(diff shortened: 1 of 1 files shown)


modules/ocf_irc/manifests/services.pp
```diff
@@ -7,7 +7,7 @@
   }
 
   $irc_creds = lookup('irc_creds')
-  $root_nicks = ['waf', 'nattofriends', 'ckuehl', 'jvperrin', 'mattmcal', 'abizer', 'dkessler', 'cooperc', 'jaw']
+  $root_nicks = ['waf', 'nattofriends', 'ckuehl', 'jvperrin', 'mattmcal', 'abizer', 'dkessler', 'cooperc', 'jaw', 'kmo', 'ronitnath', 'njha', 'fydai']
```
dkess, line 10: This nick isn't registered

dkess (changes requested):

modules/ocf_irc/manifests/services.pp
RonitNath: oh hm we have to be registered in irc?

cg505 (approved):
lgtm once dkessler's issue is resolved

modules/ocf_irc/manifests/services.pp
dkess: yes these all have to be nickserv registered names, they aren't related to OCF accounts

nikhiljha (approved):

PR #1076, opened 2021-02-25
Fix apt build of packages

ethanhs:
This was causing approve to fail to build/deploy.

### DIFF
--- a/modules/ocf_apt/files/bin/include-changes-from-stdin
+++ b/modules/ocf_apt/files/bin/include-changes-from-stdin
@@ -21,10 +21,11 @@ unzip -d "$tmpdir" "$tmpfile"
 # re-run a Jenkins job)
 # Use globbing (and also assume that there's exactly one file which matches the
 # glob)
-# shellcheck disable=SC2013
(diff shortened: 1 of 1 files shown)


jvperrin (approved):
I think this might fail pre-commit hooks due to shellcheck? Anyway, thanks for identifying and fixing my mistake on this!

ethanhs:
Hmm the puppet failures appear to be flakes? Weird, as its a different set than the ones I saw previously.

ethanhs:
I'm going to go ahead and merge this since its undoing a change and the failures appear to be unrelated.

jvperrin:
@ethanhs I think the failures in there are due to an issue with the prometheus vendor module (https://tickets.puppetlabs.com/browse/FACT-2880), but it's been patched on `production` so that's why it's not affecting all the hosts and only ones on custom environments. https://github.com/ocf/puppet/pull/1069/commits/486a4850fb416a6b5477cfc0d4d6ade186ac41f6 should fix it, but I haven't merged #1069 that contains that fix yet.

PR #1084, opened 2021-04-07
Unconditionally use restic 0.12.0

ethanhs:
I realized that restic's repository format is not stable, so we should just use a single version of it across our machines. This is mostly a defensive thing, but I don't want anything to break because we are using different versions.


ja5087 (approved):
Once the package is purged we can remove the entire if condition completely

ethanhs:
> Once the package is purged we can remove the entire if condition completely

Yeah I'll make a follow up PR later today/tomorrow.

PR #1089, opened 2021-04-17
Restart shellinabox on Wednesdays

ethanhs:


nikhiljha (approved):
it is wednesday my dudes

kpengboy:
Wonder if there's an option to gracefully restart. Or if we can restart only on not detecting liveness...

ethanhs:
> 
> 
> Wonder if there's an option to gracefully restart. Or if we can restart only on not detecting liveness...

For graceful restarting, I'm not sure. Restart on not detecting liveness is possible, but we really should do more work to do that in general for every service, so I'll push that off to a latter date.

daradib:
Might be fixed by shellinabox/shellinabox@4f0ecc31ac6f985e0dd3f5a52cbfc0e9251f6361

ethanhs:
@daradib interesting, unlikely to get backported on debian though, right?

daradib:
For stretch, yeah. Debian drops "official" support for oldstable after stable release + 1 year. [LTS support](https://wiki.debian.org/LTS/) doesn't include backports. We could backport ourselves, but probably best to focus on upgrading servers to buster and bullseye.

PR #1091, opened 2021-04-18
Install nvtop

TonyLianLong:
NVtop is a tool for displaying the utilization rates in gpus. It will be helpful for HPC users.


ja5087 (approved):

ja5087:
since corruption is on my env this PR won't apply for a while, but I already apt installed it

PR #1100, opened 2021-08-10
Replace zerotier with tailscale

ethanhs:


nikhiljha (approved):
if this is for a staffvm can't you just merge it

ethanhs:
> if this is for a staffvm can't you just merge it

yeah you're probably right, I think I forgot lol

PR #1101, opened 2021-08-16
Remove bad RBL

### LINKED ISSUE #1095: UCEProtect is extortionist
The UCEProtect RBL is a known extortion racket, its lists should not be used.

https://en.wikipedia.org/wiki/Comparison_of_DNS_blacklists#Suspect_RBL_providers

https://github.com/ocf/puppet/blob/932baeecef59b119008aa17cb2ed89f3625f49ea/modules/ocf_mail/files/spam/policyd-weight/policyd-weight.conf

ethanhs:
See https://en.wikipedia.org/wiki/Comparison_of_DNS_blacklists#Suspect_RBL_providers

Fixes #1095


singingtelegram:
Should we test the removal and see if there's a meaningful difference in spam?

ethanhs:
Yeah we can always revert if there is an increase in spam, its an easy change.

nikhiljha (approved):

PR #1103, opened 2021-08-18
remove brave browser

nikhiljha:
Went to the lab today and opened Brave; there was a lot of cryptocurrency spam in it. We probably should just remove it entirely...

see also: https://en.wikipedia.org/wiki/Brave_(web_browser)#Controversies


nattofriends:
A PR description might be good

ja5087:
what'd brave do now

also this will not remove brave from existing desktops. you need to set the package to ensure => absent

nikhiljha:
edited the original with a description

also I'm in support of pre-installing ublock origin in Firefox/Chrome but that's a separate issue I think

ja5087 (approved):
LGTM pending checks, unless octocatalog-diff is still broken

modules/ocf_desktop/manifests/init.pp
```diff
@@ -1,6 +1,5 @@
 class ocf_desktop {
   include ocf::acct
-  include ocf::packages::brave
```
dkess, line 3: removing this line will mean your `ensure => purged` will never get run

modules/ocf/manifests/packages/brave.pp
```diff
@@ -1,34 +1,11 @@
...
-      '/etc/skel/.config/BraveSoftware/Brave-Browser/',
-      '/etc/skel/.config/BraveSoftware/Brave-Browser/Default'
-    ]:
-      ensure  => directory,
-      require => File['/etc/skel/.config'],
-  } ->
-
-  file {
-    '/etc/skel/.config/BraveSoftware/Brave-Browser/Default/Preferences':
-      content => template('ocf/brave/Preferences.erb'),
```
keur, line 32: Make sure to delete this file `modules/ocf/templates/brave/Preferences.erb`. I don't see it in the diff.

keur (approved):

ethanhs:
> also I'm in support of pre-installing ublock origin in Firefox/Chrome but that's a separate issue I think

We've previously considered this, but it can break some sites and we don't want that, so I think we'd have to have an option to launch with AdBlock (no idea if that is possible)

aaronjanse (approved):

keur:
Since I am the one that added Brave and controversies are being brought up, I just want to say a small piece. 

I agree the removal is a good decision as the crypto stuff being shown to users is confusing and not suited for the lab. There are probably ways to get rid of it (similar to how we stop the browsers from asking to be default with fresh profiles) but it’s not worth it since Firefox and Chrome are good options. I do suggest some sort of adblocker by default in those would be a nice-to-have. 

The context for why we added this was because the CSO of Brave came and spoke to us back in 2018, and it was a show of good faith for the talk. 

Just want to make some counterpoints to some of the controversies linked. For affiliate codes, every major browser with a search engine deal does that. That includes the browsers we have installed. It’s how they get paid. This practice has been the norm since the early 2000s when Safari did the first major search deal with Google. The DNS leaks via Tor was not great but that’s software. Unless you happen to be Daniel Bernstein, you can’t avoid bugs and security issues, only be proactive about resolving them. 

And here are some things that generally aren’t brought up, and are very pro user. 

https://brave.com/privacy-updates-7/
https://brave.com/privacy-updates-4/

ethanhs:
Thanks keur, I really appreciate hearing your views on this, so often on (HN, reddit) I see people pile on to "Brave is Evil" trains pointing to everything they do wrong without really any discussion (as such sites are wont to do when any company does anything perceived as bad)

I do not think Brave is Evil, and having talked with quite a few people who work(ed) on Firefox, it certainly is tough to please browser users. You're right other browsers do send affiliate links for some sites. I would also hope we can all understand bugs happen. Let he who has never written a bug cast the first stone etc, etc.

I think my biggest issues with Brave are:

1. The crypto ads on the home page (which we definitely could disable, but bleh makes me worried about what else Brave may do)

2. The whole BAT system is really weird and until fairly recently was pretty anti-user (it took about 2 years from Tom Scott's video before most of the issues he complained about were satisfactorily rectified)

3. I really don't like the idea of a webkit mono-culture (I realize this isn't really Brave's fault, but ...). This is a rather fringe concern I realize, but things are bad enough for Safari and Firefox users with many sites only testing on Chrome.

4. I also find it hard to choose Brave over other options when the CEO is Brendan Eich. Opposing Prop 8 is one thing, many people changed their minds about gay marriage after 2008, people change. But Eich seems to have never apologized or changed his views, and I don't want to support a company run by someone who thinks gay couples shouldn't be able to marry.

I also think that Firefox with adblock provides a compelling alternative with few downsides. Mozilla has certainly done a few foolish things of their own in the past, but they have been quick to revert those.

To be clear, I bring up these issues not to turn this issue into a discussion of "is Brave bad or good" but to explain why I personally would prefer Firefox over Brave as the default, and I why I think that choice aligns with the OCF.

PR #1107, opened 2021-08-30
Remove brave from xfce panel

64bitpandas:
See #1103 and #308 for more details.

Now that Brave is removed from the desktops, the launcher on the topbar errors when clicked. This edit removes the launcher shortcut.


ethanhs (approved):
The CI error, while concerning, seems unrelated.

PR #1110, opened 2021-09-03
added neovim

AdarshI:
Neovim contains many useful  options for extandability, particularly in its ability to run lua scripts, which makes the extensions available for it much more powerful compared to regular vim.


ethanhs (approved):
The errors seem unrelated. Thanks for the PR!

PR #1113, opened 2021-09-03
irc: increase allowed connections from OCF subnet

dkess:
From 1000 to 5000


dkess:
ugh how does git work

PR #1118, opened 2021-09-09
Add bencuan and longlian to root

TonyLianLong:


encadyma (approved):
root principals added

Issue #726, opened 2019-08-21
update SSL versions/ciphers


cg505:
We should be periodically updating this and it was last updated 3 years ago: https://github.com/ocf/puppet/blob/941b736cdd0791c002a1aa77d0edb579f3499cb3/manifests/site_ssl.pp#L9.

cg505:
Updated ciphers include those for TLSv1.3, however it doesn't seem like puppetlabs/apache supports setting the TLSv1.3 cipher suite at this time. I may open a PR but in the meantime will probably try to set things to TLSv1.2 only with the recommended ciphers for that version from Mozilla.

PR #1131, opened 2021-09-25
Bump SSL protocol and ciphersuite to TLSv1.2

Boomaa23:
Apache 2.4 does not support TLS v1.3 because the version of OpenSSL is not up-to-date. Hence as suggested in #726 this PR updates the ciphersuite and protocols to TLS v1.2 only.


singingtelegram:
The ciphersuites look good, but I'm not sure if we should drop legacy TLS support for all hosts.

singingtelegram:
On second thought, TLS 1.2 was made mainstream back in 2013 so the vast majority of applications and libraries should support it. 

To ensure nothing important breaks, maybe we can do something like [this](https://community.centminmod.com/threads/http-tls-protocol-and-ssl-cipher-usage-statistics-logging.4985/) to log TLS protocol version used.

singingtelegram:
Generally I think supporting only TLSv1.2 (and TLS 1.3 once we migrate to bullseye) is OK, but we'll likely still need to support AES-CBC (otherwise Safari users on OS X 10.8-10.10 and IE 11 users on Windows 7 won't be able to access our sites).

singingtelegram:
Seems like we are TLSv1.2 only now (?)

https://www.ssllabs.com/ssltest/analyze.html?d=www.ocf.berkeley.edu&s=169.229.226.23&latest

singingtelegram:
I'm not sure why as the config files show that TLSv1.0 and TLSv1.1 are still supported, but I just tried to go to https://www.ocf.berkeley.edu on IE 11 on Win7 with TLSv1.2 support disabled and the page failed to load. It seems like we are indeed only supporting TLSv1.2 on prod/death now. Why?

singingtelegram:
```
$ curl --tlsv1.1 https://www.ocf.berkeley.edu
curl: (35) error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version
```

singingtelegram:
```
CONNECTED(00000005)
4304961004:error:1400442E:SSL routines:CONNECT_CR_SRVR_HELLO:tlsv1 alert protocol version:/System/Volumes/Data/SWE/macOS/BuildRoots/38cf1d983f/Library/Caches/com.apple.xbs/Sources/libressl/libressl-56.60.2/libressl-2.8/ssl/ssl_pkt.c:1200:SSL alert number 70
4304961004:error:140040E5:SSL routines:CONNECT_CR_SRVR_HELLO:ssl handshake failure:/System/Volumes/Data/SWE/macOS/BuildRoots/38cf1d983f/Library/Caches/com.apple.xbs/Sources/libressl/libressl-56.60.2/libressl-2.8/ssl/ssl_pkt.c:585:
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 7 bytes and written 0 bytes
---
New, (NONE), Cipher is (NONE)
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.1
    Cipher    : 0000
    Session-ID: 
    Session-ID-ctx: 
    Master-Key: 
    Start Time: 1633071180
    Timeout   : 7200 (sec)
    Verify return code: 0 (ok)
---
```

singingtelegram:
@ethanhs: seems like we are already supporting TLSv1.2 only despite the config files?

singingtelegram:
Closing since we are effectively TLSv1.2 only

singingtelegram:
```
death.ocf.berkeley.edu: [Mon Jan 31 16:15:11 2022] [warn] [pid 30783] ssl_engine_config.c(1402): [client AH02532: SSLProtocol: Protocol 'TLSv1.1' overrides already set parameter(s). Check if a +/- prefix is missing.
death.ocf.berkeley.edu: [Mon Jan 31 16:15:11 2022] [warn] [pid 30783] ssl_engine_config.c(1402): [client AH02532: SSLProtocol: Protocol 'TLSv1.2' overrides already set parameter(s). Check if a +/- prefix is missing.
```

this is probably why death is TLSv1.2 only

PR #1135, opened 2021-09-29
Update init.pp

singingtelegram:
enable .htaccess override for mirrors

TODO: do we require 'all'?


dkess:
What's the reasoning behind this?

This creates a minor security vulnerability, as it gives mirror vendors the ability to run arbitrary code on fallingrocks by supplying a malicious htaccess file.

singingtelegram:
There were _a lot_ of requests to large ISO files from hosts in China with IE 7 / <obsolete Chrome versions> user-agents. I wrote a .htaccess file to deny access from these user-agents, which used the Limit and FileInfo directives. I allowed only these directives at first but ubuntu-releases was broken since it contained a .htaccess file, so I set it to allow all directives for the time being.

singingtelegram:
It seems that we can disable AuthInfo though.

PR #1137, opened 2021-09-30
Debian 11 changes #2

singingtelegram:


ethanhs (commented):
Looks like there is a trailing whitespace but otherwise LGTM, thanks!

ethanhs:
I think matchbox-keyboard could be a good replacement for florence, I believe it is used by the Raspberry Pi.

singingtelegram:
what about onboard? I believe it's the default in Ubuntu. Where is the on screen keyboard used on our systems? I don't see an option to enable virtual keyboard on the logon prompt

ethanhs:
> What about onboard? I think it's the default in Ubuntu.

I think that pulls in gnome which is a lot to install

ethanhs (approved):
LGTM, thanks!

PR #1141, opened 2021-09-30
update Let'sEncrypt certificate chain

ethanwu10:
**THIS IS A STOPGAP SOLUTION.** Update the LE certificate chain in `ocf::ssl` to be the current one served by LE's ACME servers. This resolves TLS errors from DST X3 expiring, since we now serve the R3 -> ISRG X1 chain.

We should instead migrate www/apache to use the bundles created by the ACME client, instead of needing to maintain our own copy.

```
❯ openssl crl2pkcs7 -nocrl -certfile modules/ocf/files/ssl/lets-encrypt.crt | openssl pkcs7 -noout -print_certs
subject=C = US, O = Let's Encrypt, CN = R3
issuer=C = US, O = Internet Security Research Group, CN = ISRG Root X1

subject=C = US, O = Internet Security Research Group, CN = ISRG Root X1
issuer=O = Digital Signature Trust Co., CN = DST Root CA X3

subject=C = US, O = Let's Encrypt, CN = Let's Encrypt Authority X3
issuer=O = Digital Signature Trust Co., CN = DST Root CA X3
```


dongkyunc (approved):

ethanwu10:
Small mistake: `Let's Encrypt Authority X3` is not supposed to be on the chain. It shouldn't break anybody though, and just serves up an extraneous certificate.

PR #1142, opened 2021-09-30
temporarily remove blender on bullseye

singingtelegram:


singingtelegram:
fixed by #1148, no need for workarounds

PR #1145, opened 2021-10-01
Use cert chains provided by LE instead of vendoring

ethanwu10:
Currently, we vendor a copy of the Let'sEncrypt certificate chain at [`modules/ocf/files/ssl/lets-encrypt.crt`](https://github.com/ocf/puppet/blob/e2b02c464fa07be6cace208ba893b726ce4263df/modules/ocf/files/ssl/lets-encrypt.crt), which recently caused minor outages since it was not kept up-to-date (see #1141).

Remove `/etc/ssl/certs/lets-encrypt.crt` and instead rely on the chains provided by Let's Encrypt. For www, the deprecated
`SSLCertificateChainFile` directive was removed from the Apache config template, since the `SSLCertificateFile` points to the full chain obtained from Let's Encrypt. For apphost, the bundle files are still kept since they provide mtime tracking for when to trigger a reload of Nginx; the files themselves are just direct copies of the certificate from Let's Encrypt.

We probably want to patch acme-tiny to explicitly request certificate chains as well (set `Accept: application/pem-certificate-chain` - see RFC8555 §9.1). It currently does not set any `Accept:` header and thus relies on the default being `application/pem-certificate-chain` (see RFC8555 §7.4.2).

This PR is still a draft because I haven't had a chance to test this on real servers. Any guidance on how to test death/vampires's certificate issuance against LE's staging environment would be appreciated.


jvperrin (commented):
I made https://github.com/jvperrin/puppet/commit/aeea1a8885ffb7aa3b406124b76fb04aa98d5a26 for this a while back, but didn't merge it. I think I did test it on apphosts and vhosts though, but probably worth testing again.

nikhiljha:
closing in favor of #1069 (which has `ensure: absent`, seems to solve the issue mentioned in the `TODO` here)

PR #1151, opened 2021-10-03
added miniconda to ocf_hpc

Strequals:
still need to add other useful packages to hpc


modules/ocf_hpc/manifests/miniconda/apt.pp
```diff
@@ -0,0 +1,16 @@
+# Include miniconda apt repo
```
ethanhs, line 1: I would move this file into conda.pp

I also think it probably is better to name it conda.pp since that is the package name.

ethanhs (approved):
Looks good, thanks!

PR #1155, opened 2021-10-04
Removed Python 2 Files

lancemathias:
Removed Python 2 and Python 2 libraries, and Python 2 Matplotlib since Python 2 is end of life.


modules/ocf/manifests/extrapackages.pp
```diff
@@ -15,7 +15,6 @@
   include ocf::packages::chrome
   include ocf::packages::emacs
   include ocf::packages::kubectl
-  include ocf::packages::matplotlib
```
ethanhs, line 18: I don't think this should be removed, otherwise looks great!

lancemathias: Should be fixed!

ethanhs (approved):
If something actually relies on Python 2, god help us.

singingtelegram:
ipython is also based on Python 2 and no longer in the Debian 11 package repositories.

PR #1162, opened 2021-10-09
Enable TLSv1.3 on Bullseye

singingtelegram:
WIP


singingtelegram:
Note that despite the current configuration (support for TLSv1.1 and TLSv1.0), for some reason only TLSv1.2 is supported. Therefore, the new configuration should not affect protocol support other than adding support for TLSv1.3.

manifests/site_ssl.pp
```diff
@@ -6,8 +6,13 @@
...
 #
 # This should be updated from time-to-time.
-$ssl_ciphersuite = 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS'
-$ssl_protocols = 'TLSv1 TLSv1.1 TLSv1.2'
+if $::lsbdistcodename == 'bullseye' {
+  $ssl_ciphersuite = 'ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384'
+  $ssl_protocols = 'all -SSLv3 -TLSv1 -TLSv1.1'
```
ethanhs, line 11: This doesn't seem quite right, its listing ssl, and tls 1.0/1.1 only?

singingtelegram:
There’s a `-` - excluding SSLv3, TLSv1 and v1.1

modules/ocf_mirrors/files/robots.txt
```diff
@@ -0,0 +1,39 @@
+# robots.txt for mirrors.ocf.berkeley.edu
```
kpengboy, line 1: What is the reason for adding these rules? And I wonder if we can autogenerate it, to avoid it going out of sync.

singingtelegram: https://www.google.com/search?q=site%3Amirrors.ocf.berkeley.edu

Mostly for tidiness reasons I guess. Occasionally I see requests from various search engine crawlers crawling actual files, which is not ideal. Some other mirrors also have robots.txt set to disable access to the projects: 

 - https://mirrors.rit.edu/robots.txt
 - https://mirror.clarkson.edu/robots.txt
 - http://debian.csail.mit.edu/robots.txt

singingtelegram: We can (and want) to have our index page indexed, but the crawlers should not be allowed to go into the directories.

ethanhs (approved):
LGTM. You may want to auto-generate the robots.txt, but that can be left as a TODO if you prefer.

PR #1166, opened 2021-10-10
[WIP] epel: mirror epel

singingtelegram:


ethanhs:
I don't think we need to copy all of the doc and tests and such for the mirror scripts. We generally just add what is needed for mirroring.

singingtelegram:
The script itself uses zsh, and apparently Shellcheck doesn’t like that :(

I noticed that ftpsync script is not included in the files but downloaded,
maybe we can do something similar here?

Issue #1163, opened 2021-10-09
Enable HTTP/2


singingtelegram:
We don't support HTTP/2 (!)

https://www.ssllabs.com/ssltest/analyze.html?d=www.ocf.berkeley.edu&s=169.229.226.23

ethanwu10:
Apphost is still not using http/2, everything else seems to be covered? (redirects are not for some reason, but I don't think that's important)

PR #1167, opened 2021-10-11
enable http2 for vhosts and nginx

singingtelegram:
#1163


singingtelegram:
This should be a rather painless patch (if things go smoothly).

HTTP/2's benefit is pretty significantly especially when there are a lot of resources and the user's network is slow/high-latency.

https://imagekit.io/demo/http2-vs-http1

ethanhs:
I haven't brought up dev-death yet but we should do that anyway. (Also looks like this needs a rebase on master)

singingtelegram:
Yes.

Issue #1165, opened 2021-10-10
paper-monitor font size issue


singingtelegram:
10pt appears too small on 2K displays

PR #1170, opened 2021-10-11
Update page indicator font size

singingtelegram:
Update font size to 13pt to match that of the clock and fixed display issue on HiDPI displays


singingtelegram:
Fixes #1165.

PR #1171, opened 2021-10-12
fix collect-mirrors-stats

singingtelegram:


ethanhs (approved):
This looks like it makes sense, thanks!

PR #1172, opened 2021-10-12
Temporarily switch to upstream for debian-cd

singingtelegram:


singingtelegram:
Alternatively, we can sync from mirrors.kernel.org which is closer to us.

ethanhs (approved):
I don't have a strong preference between upstream and kernel.org, but kernel.org probably has less demand, so if they offer the same files it might be nice to use that.

PR #1175, opened 2021-10-14
Add committee signup

RonitNath:


modules/ocf_www/manifests/site/shorturl.pp
```diff
@@ -15,7 +15,7 @@
     ssl_cert      => "/etc/ssl/private/${::fqdn}.crt",
     ssl_chain     => "/etc/ssl/private/${::fqdn}.intermediate",
 
-    rewrites      => [
+rewrites      => [
```
singingtelegram, line 18: modules/ocf_www/manifests/site/shorturl.pp - FIXED: indentation of => is not properly aligned (expected in column 19, but found it in column 15) on line 18 (check: arrow_alignment)

singingtelegram (changes requested):

PR #1178, opened 2021-10-15
Update debian.pp

singingtelegram:
Seems like we are syncing about an hour before Debian archives are updated, so we are almost immediately out of date :(

This should fix that.

https://mirror-master.debian.org/status/mirror-info/mirrors.ocf.berkeley.edu.html
https://mirror-master.debian.org/status/mirror-info/mirrors.wikimedia.org.html


ethanhs:
Only suggestion is that we may want to change to an uneven (i.e. not divisible by 15) minute so we aren't hitting the same parts of the mirror as others syncing.

PR #1193, opened 2021-11-06
Update report-sizes

singingtelegram:
We are using zfs with each project as a dataset, so we can simply process the output of `zfs list` to get the sizes.


ethanhs (approved):
Awesome! I assume you've tested this?

singingtelegram:
Yep - tested.

PR #1196, opened 2021-11-09
Mirrors changes

singingtelegram:
- added: ubuntu-ports, videolan-ftp, gnome
 - added healthcheck for ubuntu-ports and videolan-ftp
 - switched to nginx for better performance (+30% https throughput, lower CPU usage??)
   - bumped puppet-nginx to 3.3.0, there should be no actual changes on other nginx servers
   - this of course breaks the Apache Software Foundation's .htaccess file, but since they discontinued the mirror network, page styling doesn't matter anymore as an unofficial Apache mirror.
   - related changes to collect-mirrors-stats - WIP, functional but need to change "Apache" -> "Nginx"
 - blocked another user-agent
 - new look
 - switched to ftp.usa.openbsd.org from ftp4.usa.openbsd.org for faster sync speeds
 - GNU: sync from ftp.gnu.org

These changes reflect the current state of fallingrocks. (The switch to nginx was tested on dev-fallingrocks before it was deployed to fallingrocks) This PR supersedes all previous unmerged mirrors PRs.

We are now official mirrors of OpenBSD, ubuntu-ports, EPEL, VideoLAN, and GNOME. (traffic 📈)


modules/ocf/manifests/init.pp
```diff
@@ -5,7 +5,9 @@
   include ocf::etc
   include ocf::firewall
   include ocf::groups
-  include ocf::hidepid
+  if ($::hostname != 'fallingrocks') {
```
ethanhs, line 8: Could you leave a comment why we are not including this on fallingrocks?

modules/ocf_mirrors/files/collect-mirrors-stats
```diff
@@ -17,10 +17,10 @@ from ocflib.lab.stats import humanize_bytes
 
 OCFSTATS_PWD = open(os.environ.get('OCFSTATS_PWD__FILE')).read().strip()
 MIRRORS_DATA_PATH = Path('/opt/mirrors/ftp')
-APACHE_LOG_PATH = Path('/var/log/apache2')
+APACHE_LOG_PATH = Path('/var/log/nginx')
```
ethanhs, line 20: Perhaps we should rename these `APACHE_*` variables to reflect the move to Nginx.

ethanhs (commented):
Just a couple of minor things but overall looks excellent, thank you so much for working on all of these changes!

modules/ocf/manifests/init.pp
singingtelegram: I'm running [nginx amplify](https://amplify.nginx.com/login) on fallingrocks for additional insights, since nginx's `stub_status` module is indeed a stub compared to `mod_status` on Apache. The agent runs as the www-data user, and it needs to know the pid of the master nginx process.

modules/ocf_mirrors/files/collect-mirrors-stats
singingtelegram: Good catch!

ethanhs (approved):
Thanks!

PR #1199, opened 2021-11-13
switch to cri-o for kubernetes

nikhiljha:
Here's why we're switching from containerd...

- the NVidia driver only just added support for it because there are weird issues
- other kubernetes distributions seem to be moving to cri-o over containerd https://github.com/poseidon/typhoon/issues/899
- I ran into the issue where java apps just don't work on containerd out of the box one too many times


nikhiljha:
(ethan says he approves this)

dghubble:
I saw you cross-linked my issue as an example of cri-o adoption. However, we adopted containerd just fyi.

PR #1203, opened 2021-11-18
make default netboot option bullseye

nikhiljha:
This way I can netboot something and walk away instead of having to be there to press down arrow 😁


ethanhs (approved):

ethanhs:
Also I think the netboot config for bullseye is missing from puppet, we should make sure to add it.

singingtelegram (approved):

PR #1210, opened 2021-12-01
enable non-root kubectl for ocfroot on newk8s

ethanwu10:
Change permissions on /etc/kubernetes to add execute for ocfroot, and set permissions for all files within to allow read by root only. This allows kubectl to run without root when KUBECONFIG is pointed to /etc/kubernetes/admin.conf.

Still TODO: copy over the profile.d entry from ocf_kubernetes to set KUBECONFIG correctly in shell profiles (and also evaluate the etcd-related access permissions)

<sup>Also why does github default to committing to master in the web UI ;-;</sup>


nikhiljha:
This looks good for now, but we should add in kerberos auth soon(TM)

PR #1090, opened 2021-04-18
Add HPC packages to extrapackages

encadyma:
This is for corruption. Should consider removing jupyterhub from some machines.

PR #1214, opened 2021-12-06
Revert "Add HPC packages to extrapackages"

ethanhs:
Reverts ocf/puppet#1090

This broke puppet/apt on most hosts :/


ethanhs:
cc @encadyma

It looks like a lot of the dependencies are not available on most servers.

PR #1216, opened 2021-12-09
Update firewall rules to include all IPMI hosts

singingtelegram:
Update special devices zone to include all IPMI hosts


singingtelegram:
(note: please ignore tornado as tornado is madcow for now since the NUC died. "tornado" is running under my environment due to ocf_tv and nvidia both include a xorg.conf file)

ethanwu10 (approved):
lgtm

PR #1217, opened 2021-12-10
make ocf-create.conf readable only by root

nikhiljha:


modules/ocf_admin/manifests/create.pp
```diff
@@ -7,9 +7,8 @@
     '/etc/ocf-create':
       ensure => directory;
 
-    # TODO: ideally this file wouldn't be directly readable by staff
     '/etc/ocf-create/ocf-create.conf':
-      group     => ocfstaff,
+      group     => ocfroot,
```
ethanwu10, line 11: `approve` still needs the file to read the celery connection info; did you intend for `approve` to become root-only?

ethanwu10: Also, it seems the proper solution for this is to separate out `celery.broker` and `celery.backend` to a separate config file... is it worth the effort at this point when we're planning to redo secrets management for create when it gets moved into newk8s?

nikhiljha: I think approve is mostly used by rootstaff anyway, so this is probably a good idea?

dkess: If you're going to make approve root-only, don't forget to update the docs. I can think of https://www.ocf.berkeley.edu/docs/staff/scripts/approve/ and https://www.ocf.berkeley.edu/docs/staff/powers/ as pages that mention this.

PR #1221, opened 2021-12-27
Add .well-known for Matrix clients

Kalissaac:
QOL improvement for those using Matrix, allows users to simply type in their full user ID (`@oski:ocf.berkeley.edu`) and compliant clients will auto-fill the homeserver.
The server version of this already exists to allow for federation, this just adds a corresponding portion for clients.

Synapse documentation: https://matrix-org.github.io/synapse/latest/setup/installation.html#client-well-known-uri

Certified to save hundreds of seconds a year!


modules/ocf_www/files/matrix-client
```diff
@@ -0,0 +1,9 @@
+{
+    "m.homeserver": {
+        "base_url": "https://matrix.ocf.berkeley.edu",
+        "server_name": "ocf.berkeley.edu"
+    },
+    "m.identity_server": {
+        "base_url": "https://vector.im"
```
ethanhs, line 7: Doesn't identity come from keycloak? (perhaps this is a matrix thing I don't know about)

modules/ocf_www/manifests/site/www.pp
```diff
@@ -31,6 +31,8 @@
       ensure => 'directory';
     '/var/www/html/.well-known/matrix/server':
       source => 'puppet:///modules/ocf_www/matrix-server';
+    '/var/www/html/.well-known/matrix/client':
```
ethanhs, line 34: The docs say you should add a CORS header for browser clients:

> To work in browser based clients, the file must be served with the appropriate Cross-Origin Resource Sharing (CORS) headers. A recommended value would be Access-Control-Allow-Origin: * which would allow all browser based clients to view it.

ethanhs (approved):
This would be great! I do find it annoying to have to enter this info every time I log in :)

modules/ocf_www/files/matrix-client
Kalissaac: From my understanding, the identity servers in Matrix are so other users can discover you through your phone or email (if connected). I just copied it from the [existing config we use for the hosted Element instance](https://github.com/ocf/matrix/blob/d9db94473f8daa517208e5c6ec1aa9189f9fe4fe/config.riot.json), but I can remove it if you'd like since this portion is optional. Most clients already default to vector.im anyways.

modules/ocf_www/manifests/site/www.pp
Kalissaac: Thanks for pointing that out, forgot about that part. I'll add that right now.

Kalissaac: I'm not too familiar with puppet, do you know of a way to set a header for a single file or directory that is aliased? The only conditions I can see in the [Apache docs](https://httpd.apache.org/docs/current/mod/mod_headers.html#header) is `always` or `onsuccess`, and I'd be concerned with adding allow origin from anywhere to the entire www.

Kalissaac: If we changed the whole `/.well-known/matrix` path to be a rewrite to `matrix.obe` the headers from there would be passed through right? We already serve the client well-known file from there and could just add the server file.

modules/ocf_www/files/matrix-client
ethanhs: Aha, I see, that makes more sense.

Kalissaac:
muchas gracias

PR #1245, opened 2022-02-07
feat: use rocky mirror script

singingtelegram:


modules/ocf_mirrors/manifests/projects/rocky.pp
```diff
@@ -10,8 +10,8 @@
 
   ocf_mirrors::timer {
     'rocky':
-      exec_start => '/opt/mirrors/project/rocky/sync-archive',
-      hour       => '1/3',
+      exec_start => '/opt/mirrors/project/rocky/sync-archive > /dev/null 2>&1',
```
dkess, line 13: why throw out the output? this gets stored in journalctl and could be useful to inspect if something goes wrong

singingtelegram: removed, although the output is redirected to logfile

dkess: Looks like they have two "logs", one is intended to be public facing (saved in `/opt/mirrors/project/rocky`) and the other uses the `logger` command to write directly to the syslog. So I spoke a bit too soon here, it probably doesn't matter if we redirect the `/dev/null` or not, though it certainly doesn't hurt to keep stdout around in case it ends up being useful for debugging purposes later on.

dkess (approved):

PR #1251, opened 2022-02-12
fix: don't redirect on ocf.b.e

singingtelegram:


singingtelegram:
With the matrix .well-known commit, it's now possible to use @oski:www.ocf.berkeley.edu as username. However, since ocf.b.e returns 301 to www.o.b.e, discovery fails for @oski:ocf.berkeley.edu. This PR attempts to fix that by adding o.b.e as an alias of www.o.b.e.

Kalissaac (commented):
i know we haven't tested this, but do we know if content gets served correctly on the root domain?

Kalissaac:
i guess there's only one way to find out anyways..

PR #1253, opened 2022-02-16
Add coffeechats shorturl

64bitpandas:


Kalissaac (approved):
Lgtm!

Kalissaac:
Actually, it looks like Google Sheets doesn't like the url provided here, I'm going to trim some stuff off the end so it behaves
image

PR #1256, opened 2022-02-17
Add bat to puppet

laksith19:
A cat clone with support for syntax highlighting, line numbers, git diff etc..


64bitpandas (approved):
lgtm

PR #1268, opened 2022-03-03
Update shorturl.pp

bryli:
Added 5th staff meeting URL to shorturls at ocf.io/sp22-mt5 (or ocf.io/sp22mt5)


Kalissaac (approved):
very cool

PR #1271, opened 2022-03-10
fix: json (cursed)

singingtelegram:


ethanwu10 (approved):
certified cursed

dkess:
since this is an erb file you could also use template comments. eg:

```erb
<%# this is a comment %>
```

PR #1272, opened 2022-03-11
Treat fallingrocks as internal

singingtelegram:
Fallingrocks is no longer in the OCF subnet


singingtelegram:
VPN somehow makes fallingrocks inaccessible to the hosts inside the OCF network (due to source IP mismatch?)

PR #1278, opened 2022-03-17
use https in mirrors

nikhiljha:
Campus does deep packet inspection on HTTP traffic from the DC to our lab. This makes apt slow. Using https solves this because they don't do DPI on HTTPS.


singingtelegram (approved):
lgtm

PR #1289, opened 2022-04-16
fix: enable ChallengeResponseAuthentication

singingtelegram:


ethanwu10:
Given that it is disabled by default (supposedly - at least according to redhat - for [security reasons](https://access.redhat.com/solutions/336773)), should we only enable this on tsunami (and maybe supernova)?

axmmisaka:
> Given that it is disabled by default (supposedly - at least according to redhat - for [security reasons](https://access.redhat.com/solutions/336773)), should we only enable this on tsunami (and maybe supernova)?

I guess it would be fine if they are prompted to change password right away (given it's a shell login)

PR #1291, opened 2022-04-26
fix: upgrade nginx controller to 0.51.0

singingtelegram:
We are 3 years behind on our controller version. I checked the changelog and the minimum Kubernetes version should remain at 1.16 for this release, with no other breaking changes noted.


ethanwu10:
This broke remote IP passing/parsing, which breaks a lot of APIs which check that the connecting IP is an OCF IP (e.g. session stats).

PR #1303, opened 2022-09-01
fix: only install bat on Debian 11+

ethanwu10:
The package doesn't exist on earlier versions...


modules/ocf/manifests/extrapackages.pp
```diff
@@ -193,6 +192,14 @@
     ]:;
   }
 
+  if Integer($::os['distro']['release']['major']) >= 11 {
```
nikhiljha, line 195: This is cursed but if @singingtelegram thinks its fine too then lgtm i guess

nikhiljha (approved):

PR #1320, opened 2022-10-09
chore: bump nginx module

singingtelegram:


singingtelegram:
As shown in the changes log, the updated module no longer depends on /run/nginx, thus fixing the issue that nginx fails to start on boot.

Kalissaac (approved):

PR #1332, opened 2022-10-23
Elevated priority for ingress nginx pods.

bryli:
Increase priority of Ingress-NGINX controller pods to `system-cluster-critical` in order to prevent pods from being preemption/eviction of pods.

### DIFF
--- a/modules/ocf_kubernetes/manifests/master/ingress/nginx.pp
+++ b/modules/ocf_kubernetes/manifests/master/ingress/nginx.pp
@@ -2,8 +2,6 @@ class ocf_kubernetes::master::ingress::nginx {
 
   $kubernetes_worker_nodes = lookup('kubernetes::worker_nodes')
   $kubernetes_workers_ipv4 = $kubernetes_worker_nodes.map |$worker| { ldap_attr($worker, 'ipHostNumber') }
-  $nginx_version = lookup('kubernetes::nginx_version')
-  $ingress_nginx_url = "https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v${nginx_version}/deploy/static/provider/baremetal/deploy.yaml"
 
   file {
     default:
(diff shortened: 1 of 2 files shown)


singingtelegram (approved):

modules/ocf_kubernetes/manifests/master/ingress/nginx.pp
```diff
@@ -16,11 +14,15 @@
     '/etc/ocf-kubernetes/manifests/ingress/ingress-expose.yaml':
       content => template('ocf_kubernetes/ingress/ingress_expose.yaml.erb'),
       mode    => '0644';
+
+    '/etc/ocf-kubernetes/manifests/ingress/ingress-deploy.yaml':
+      content => 'puppet:///modules/ocf_kubernetes/ingress_deploy.yaml',
```
ethanwu10, line 19: This sets the content of the file to literally `puppet:///modules/ocf_kubernetes/ingress_deploy.yaml`; correct attribute is `source`

modules/ocf_kubernetes/manifests/master/ingress/nginx.pp
```diff
@@ -2,8 +2,6 @@
 
   $kubernetes_worker_nodes = lookup('kubernetes::worker_nodes')
   $kubernetes_workers_ipv4 = $kubernetes_worker_nodes.map |$worker| { ldap_attr($worker, 'ipHostNumber') }
-  $nginx_version = lookup('kubernetes::nginx_version')
-  $ingress_nginx_url = "https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v${nginx_version}/deploy/static/provider/baremetal/deploy.yaml"
```
ethanwu10, line 6: We need some way of actually tying our manifests to upstream...
Since the hiera key is not used, it should be removed, and we should maintain a patch/diff relative to upstream instead of having an opaque modified blob.

ethanwu10:
PR unmerged.
@bryli please re-open the PR

PR #1349, opened 2023-01-30
fix: update chrome and electron apps automatically

singingtelegram:


Kalissaac (commented):
where firefox

Kalissaac (approved):
jk firefox never has cves

singingtelegram:
firefox is in stable-security

PR #1362, opened 2023-04-17
feat: smart monitoring

singingtelegram:


Kalissaac (approved):

Kalissaac:
do we want to run this on all the desktops too or just servers?

singingtelegram:
it should be run on the desktops as well

PR #1365, opened 2023-05-05
fix: remove atom

singingtelegram:
Atom has been deprecated for a while now, and we have VSCode installed. Keeping old Electron-based apps is not great for security.


singingtelegram:
tested on acid

PR #1367, opened 2023-05-08
Use stronger keys for krb5 principals

singingtelegram:
see https://github.com/heimdal/heimdal/issues/988


singingtelegram:
also see https://datatracker.ietf.org/doc/html/rfc8429

PR #1394, opened 2023-09-07
add missing oliverni admin and root principals

oliver-ni:


singingtelegram:
approved retroactively

PR #1395, opened 2023-09-09
Update packages.pp

oski-bear:
🫡 etw


ethanwu10 (changes requested):
terminfo should go on all machines in module `ocf` (since it's needed on the machine being *accessed* for a terminal to work correctly).

PR #1396, opened 2023-09-09
Proxy apphosts through death

Kalissaac:
Still need to figure out how we want to handle groups who want to switch to apphosting, since our current method of adding them to `vhost-app.conf` will cause their development domain to be promoted to production immediately even if they're still working on developing their app or configuring systemd. Maybe we initially add `group-studentorg-berkeley-edu.apphost.ocf.berkeley.edu` as their domain and then edit it to `group.studentorg.berkeley.edu` when they're ready to promote the site to production?

### DIFF
--- a/modules/ocf_www/files/vhost-web.jinja
+++ b/modules/ocf_www/files/vhost-web.jinja
@@ -18,7 +18,6 @@
         # Doesn't matter too much for vhosts.
         RewriteRule ^(.*)$ {{vhost.redirect_dest}}$1 [L,R=302]
     {% elif vhost.is_apphost %}
-        RewriteEngine on
         RequestHeader set X-Forwarded-Proto https
         ProxyPreserveHost On
         SSLProxyEngine on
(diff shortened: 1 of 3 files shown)


modules/ocf_www/files/vhost-web.jinja
```diff
@@ -17,6 +17,13 @@
         # 301 redirects are more correct, but get cached forever by dumb browsers.
         # Doesn't matter too much for vhosts.
         RewriteRule ^(.*)$ {{vhost.redirect_dest}}$1 [L,R=302]
+    {% elif vhost.is_apphost %}
+        RewriteEngine on
```
oliver-ni, line 21: Don't believe this line is necessary

```suggestion
```

oliver-ni (approved):

oliver-ni:
Probably wait on rjz approval not entirely familiar w Apache

singingtelegram (approved):
send it

Kalissaac:
We need to hardcode death's IPs since nginx is too old to support hostnames in `set_real_ip_from` :/

PR #1399, opened 2023-09-13
"maybe you should PR getting foot-terminfo installed on all ocf machi…

oski-bear:
…nes"

### DIFF
--- a/modules/ocf/manifests/packages.pp
+++ b/modules/ocf/manifests/packages.pp
@@ -141,6 +141,13 @@ class ocf::packages {
     }
   }
 
+  # TODO: remove this once we replace vampires
+  if $::os[distro][codename] >= 11 {
+    package {
+      'foot-terminfo':;
+    }
+  }
+
   ocf::repackage { 'python3-attr':
     backport_on =>  ['stretch'],
   }


nikhiljha (changes requested):
Looks like this errored because codename is a string not a number. For an example of how to do this, see...

https://github.com/ocf/puppet/blob/e50094f4966d150f2fb0567e48ee25225aa710e1/modules/ocf/manifests/packages.pp#L136-L142

nikhiljha (changes requested):
Looks like this still fails because of an extra space on line 137...

```
[2023-09-18T20:09:49.871Z] Trim Trailing Whitespace........................................................................Failed
[2023-09-18T20:09:49.871Z] - hook id: trailing-whitespace
[2023-09-18T20:09:49.871Z] - exit code: 1
[2023-09-18T20:09:49.871Z] - files were modified by this hook
[2023-09-18T20:09:49.871Z] 
[2023-09-18T20:09:49.871Z] Fixing modules/ocf/manifests/packages.pp
[2023-09-18T20:09:49.871Z] 
```

To avoid issues like this in the future you should install the pre-commit hooks: https://github.com/ocf/puppet#linting-and-validating-the-puppet-config

nikhiljha:
Also, `foot-terminfo` is only available on bullseye or higher, so you can't put this in the same block.

See https://packages.debian.org/bullseye/foot-terminfo

ethanwu10:
#1351 relevant (I should really get that merged...)

PR #1402, opened 2023-09-21
Fix Force Rasterize m806.ppd.epp

jaysa68:
Okular seems to corrupt certain PDFs when printing on profiles other than ProRes 1200; change the default, since this seems to work.


ethanwu10:
Turns out this prints with colors inverted :(

singingtelegram:
closed

PR #1403, opened 2023-09-21
Replace xournal with xournalpp

jaysa68:
Xournalpp is a more feature-full rewrite of xournal. Potentially, we can make xournalpp our new default pdf viewer instead of Okular if we cannot fix the need to Force Rasterize for certain jobs. Even if we don't... xournalpp is still nice.


ethanwu10 (approved):
lgtm

PR #1410, opened 2023-10-27
Adding pass to ocf desktops

lharriscal:
https://www.passwordstore.org/
Added pass.

Having a simple password manager installed opens up many more potential uses for ocf computers. There is negligible downside.


jaysa68 (approved):
very cool

PR #1416, opened 2023-12-27
Add decal.ocf.io zone key

Kalissaac:


Kalissaac:
not a fan of the secret name but my creative spirit isn't really functioning rn

PR #1417, opened 2023-12-30
Add DNS64 server config

Kalissaac:


hieradata/common.yaml
singingtelegram, line 1: can we allocate a smaller range for decal, say ~180 IPs

PR #1422, opened 2024-02-13
low key oom should restart desktops

lharriscal:
Desktops just hang on a frozen screen if they run out of memory. Chrome likes using a lot of memory. I would argue that restarting the computer automatically is a more desirable behaviour for the end user than just freezing.

Therefore, on out of memory we now kernel panic, and on kernel panics we restart the computer after 5 seconds.

Already tested on firewhirl

etw wanted to test something else though before merging this.

### DIFF
--- a/modules/ocf_desktop/manifests/init.pp
+++ b/modules/ocf_desktop/manifests/init.pp
@@ -14,6 +14,7 @@ class ocf_desktop {
   include ocf_desktop::firewall_output
   include ocf_desktop::grub
   include ocf_desktop::modprobe
+  include ocf_desktop::oom_manage
   include ocf_desktop::packages
   include ocf_desktop::printnotify
   include ocf_desktop::sshfs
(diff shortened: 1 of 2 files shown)


kpengboy:
Question: in OOM situations why doesn't the OOM killer kill the most offending process, thus solving the problem without a reboot? Can it be tuned to kill before the system freezes? What unintended side effects might there be if we kernel panic instantly on OOM?

lharriscal:
> Question: in OOM situations why doesn't the OOM killer kill the most offending process, thus solving the problem without a reboot? Can it be tuned to kill before the system freezes? What unintended side effects might there be if we kernel panic instantly on OOM?

You are definitely right on this so I did some investigation. I tried to replicate computer freezing on OOM, and I was able to do so. I tried on another computer and OOM killer functioned as expected. I am not sure whether it just only sometimes functions as expected or if it only works on some computers. more testing required.

However, none of these cases are hit in practice except on computers with low memory. Once we upgrade the problematic desktop, we likely wont run into this issue often.

lharriscal:
Ok so I did a bit more research. Looks like it is just a few specific desktops that have this issue.

cg505:
banger PR title
is it related to differences in swap?

PR #1440, opened 2025-04-01
ocf_www: fix ocf.io/ofc shorturl

kpengboy:
ocf.io/ofc is supposed to redirect to www.ocf.b.e/~ofc which is supposed to then redirect to https://www.ocf.berkeley.edu/announcements/2016-04-01/renaming-ocf, but the latter redirect is frequently broken because the `ofc` account is usually sorried. Make it an explicit shorturl to avoid this problem.


Storce:
lmao was this a April fool's thing

kpengboy:
Yeah, it was an old April fool

PR #1463, opened 2026-04-06
ocf_www: proxy apphost vhosts directly from nginx

oliver-ni:
Have nginx talk to apphost.ocf.berkeley.edu directly instead of going through Apache. Apache no longer needs SSL support so drop apache::mod::ssl and ocf_www::ssl.


jaysa68 (approved):
sure

PR #1464, opened 2026-04-06
ocf_www: speed up build-vhosts with bulk LDAP query and parallelism

oliver-ni:
speeds it up to 4 seconds from like 10 minutes


jaysa68 (approved):
sure

PR #1484, opened 2026-05-19
ocf_ns: fix BIND9 journal corruption on zone file updates

oliver-ni:
## Summary

Fixes the recurring "journal out of sync with zone" errors on pestilence where BIND9 refuses to load zones after Puppet updates zone files from git.

**Root cause**: Puppet's `vcsrepo` does a `git pull` on `/srv/dns`, overwriting zone files on disk. When BIND has pending journal entries (from dynamic updates to `letsencrypt.ocf.io` or DNSSEC inline signing on other zones), the `.jnl` files become inconsistent with the zone files that git just overwrote. On reload, BIND fails to roll forward the stale journal and refuses to load the zone.

**Fix**: 
1. Add `exec['bind9-sync-clean']` that runs `rndc sync -clean` **before** `vcsrepo` pulls, but only when `.jnl` files are present. This flushes BIND's in-memory journal state to the zone files and removes the `.jnl` files so git can cleanly overwrite them.
2. Replace `notify => Service['bind9']` on `vcsrepo` with `notify => Exec['bind9-reload']` using `rndc reload` for a graceful reload instead of a full service restart.

The `named.conf.options` file still notifies `Service['bind9']` for full restarts on config changes (e.g. key rotation).

## Review & Testing Checklist for Human

- [x] Verify that `rndc sync -clean` is safe to run on pestilence — run it manually and confirm BIND continues serving normally
- [ ] After deploying, confirm that a Puppet run on pestilence with pending `.jnl` files correctly flushes and cleans them before the git pull
- [ ] Trigger a cert renewal (or wait for one) and verify that the `letsencrypt.ocf.io` zone still accepts dynamic updates after a Puppet run
- [ ] Check that `named.conf.options` changes (e.g. key rotation) still trigger a full service restart via `Service['bind9']`

Recommended test: On pestilence, ensure some `.jnl` files exist (e.g. trigger a dynamic update), then run Puppet manually (`puppet agent -t --environment=<your-env>`) and verify no "journal out of sync" errors appear in BIND logs.

### Notes

The `onlyif` guard uses `find ... -print -quit | grep -q .` to efficiently check for any `.jnl` files without scanning the entire directory. This means `rndc sync -clean` is skipped entirely when there are no journals to clean (most Puppet runs).

Link to Devin session: https://app.devin.ai/sessions/8d46bd675e484a7ea82c9112d653a775
Requested by: @oliver-ni

### DIFF
--- a/modules/ocf_ns/manifests/init.pp
+++ b/modules/ocf_ns/manifests/init.pp
@@ -28,7 +28,6 @@ class ocf_ns {
   # with zone" errors on reload.
   exec { 'bind9-sync-clean':
     command => '/usr/sbin/rndc sync -clean',
-    onlyif  => '/usr/bin/find /srv/dns/etc/zones -name "*.jnl" -print -quit | /bin/grep -q .',
     require => Service['bind9'],
     before  => Vcsrepo['/srv/dns'],
   }
--- a/modules/ocf_ns/manifests/init.pp
+++ b/modules/ocf_ns/manifests/init.pp
@@ -22,6 +22,17 @@ class ocf_ns {
       notify  => Service['bind9'];
   }
 
+  # Flush BIND journals to zone files and remove .jnl files before git pull.
+  # Without this, git overwrites zone files that have pending journal entries
+  # (from dynamic updates or inline signing), causing "journal out of sync
+  # with zone" errors on reload.
+  exec { 'bind9-sync-clean':
+    command => '/usr/sbin/rndc sync -clean',
+    onlyif  => '/usr/bin/find /srv/dns/etc/zones -name "*.jnl" -print -quit | /bin/grep -q .',
+    require => Service['bind9'],
+    before  => Vcsrepo['/srv/dns'],
+  }
+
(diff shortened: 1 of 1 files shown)


modules/ocf_ns/manifests/init.pp
```diff
@@ -22,6 +22,17 @@
       notify  => Service['bind9'];
   }
 
+  # Flush BIND journals to zone files and remove .jnl files before git pull.
+  # Without this, git overwrites zone files that have pending journal entries
+  # (from dynamic updates or inline signing), causing "journal out of sync
+  # with zone" errors on reload.
+  exec { 'bind9-sync-clean':
+    command => '/usr/sbin/rndc sync -clean',
+    onlyif  => '/usr/bin/find /srv/dns/etc/zones -name "*.jnl" -print -quit | /bin/grep -q .',
```
oliver-ni: do we need this filter? we can just run all the time right

sophiebsw (approved):

PR #1485, opened 2026-05-25
ocf_www: add per-user fcgid process limits to prevent worker exhaustion

oliver-ni:
## Summary

Adds `FcgidMaxProcessesPerClass 5` and tightens related mod_fcgid limits to prevent a single user's site from exhausting all Apache worker threads.

**Root cause:** With the default `FcgidMaxProcessesPerClass` of 100, one user (e.g. a compromised WordPress site attracting bot traffic) could spawn enough php-cgi backends to consume the entire global pool. Apache workers would block waiting on slow/hung fcgid backends, hitting `MaxRequestWorkers` (300) within seconds and causing 504 Gateway Timeout errors for every other site on the server.

**Fix:** Cap each user (suexec UID = fcgid "class") to 5 PHP backends. When all 5 are busy, mod_fcgid returns 503 quickly instead of tying up an Apache worker thread — other vhosts stay healthy.

Changes to `modules/ocf_www/manifests/mod/fcgid.pp`:
- `FcgidMaxProcessesPerClass 5` — per-user backend cap (was default 100)
- `FcgidMaxProcesses 200` — global backend cap (was default 1000)
- `FcgidBusyTimeout 120` — kill stuck backends after 2 min (was default 300)
- `FcgidIOTimeout 40` — explicit I/O timeout (matches default)
- `FcgidIdleTimeout 300` — reap idle backends after 5 min (was 3600)

## Review & Testing Checklist for Human

- [ ] Verify `FcgidMaxProcessesPerClass 5` is appropriate for the heaviest legitimate user sites — if any user genuinely needs more than 5 concurrent PHP backends, this will throttle them with 503s. Monitor after deploy and bump if needed.
- [ ] After puppet applies, check `ps aux | grep php-cgi | wc -l` on death — count should stay well under 200 instead of spiking to 230+.
- [ ] Confirm no 504 errors during normal traffic after deploy by tailing `/var/log/nginx/access.log` for 5xx responses.

### Notes

The previous `FcgidIdleTimeout 3600` allowed idle backends to accumulate across hundreds of user vhosts. Reducing to 300s (5 min) means backends are reaped sooner when not in use, keeping the global pool healthier. `FcgidProcessLifeTime 3600` is unchanged — backends still get recycled after 1 hour max regardless.

Link to Devin session: https://app.devin.ai/sessions/268d2734e827422a966ae0d294d76377
Requested by: @oliver-ni

### DIFF
--- a/modules/ocf_www/files/vhost-web-nginx.jinja
+++ b/modules/ocf_www/files/vhost-web-nginx.jinja
@@ -20,7 +20,7 @@ server {
     }
 
     location / {
-        limit_req zone=per_vhost burst=100 nodelay;
+        limit_req zone=per_vhost burst=20 nodelay;
         limit_req_status 429;
 
         {% if vhost.is_redirect %}
--- a/modules/ocf_www/files/vhost-web-nginx.jinja
+++ b/modules/ocf_www/files/vhost-web-nginx.jinja
@@ -20,7 +20,7 @@ server {
     }
 
     location / {
-        limit_req zone=per_ip burst=20 nodelay;
+        limit_req zone=per_vhost burst=100 nodelay;
         limit_req_status 429;
 
         {% if vhost.is_redirect %}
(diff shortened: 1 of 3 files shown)


jaysa68 (approved):
sure. give it a shot if youd like

PR #1489, opened 2026-07-17
add ldconfig and nvidia-container-cli paths

Sophon96:
needed for containers that use --nv or --nvccli flags


sophiebsw (approved):
lgtm + tested

PR #1493, opened 2026-09-03
ocf_hpc: add GPU readiness probe and recovery runbook

bluetopazz:
## Summary

- add a read-only GPU readiness probe that checks `nvidia-smi`, NVIDIA/UVM modules and devices, and CUDA Driver API initialization
- install the probe only through `ocf_hpc::compute`
- document coherent NVIDIA/kernel maintenance, failure isolation, rollback, and deployment boundaries
- document safe diagnosis and recovery when GPU memory remains allocated without a Slurm owner or live PID
- add focused tests and run them from pre-commit

## Why

A recent recovery exposed two separate gaps between management-plane health and compute readiness:

1. `nvidia-smi` could enumerate the RTX A6000s while `nvidia_uvm` and `/dev/nvidia-uvm` were absent, causing CUDA initialization to fail.
2. After CUDA was restored, a Slurm-scoped diagnostic found an idle node with no queued jobs while one GPU retained 21,704 MiB against a PID that was absent from `/proc`. `nvidia-smi pmon` and `fuser` found no live owner.

The second observation is evidence for a stale-context candidate, not a proven root cause. The runbook therefore requires evidence capture and administrator verification before reset or reboot.

The readiness probe checks the infrastructure boundary directly through `libcuda.so.1`; it does not depend on PyTorch, Conda, or a user environment.

## Safety / scope

This PR does not change driver versions, package sources, Slurm policy, service ordering, or node state. The probe is manual and read-only.

The stale-memory procedure:

- correlates Slurm ownership, NVIDIA accounting, and live processes before intervention
- drains the node before host-level recovery
- permits targeted reset only after jobs, processes, and peer/display dependencies are excluded
- uses reboot as the fallback when isolated reset is unsupported or ineffective
- requires idle-memory, direct-driver, Slurm CUDA, memory-release, and kernel-log checks before resume

Automatic draining, reset, monitoring integration, and package/version policy remain outside this PR.

## Validation

- `python3 -m unittest discover -s modules/ocf_hpc/tests -p "test_*.py" -v`: 6 tests passed
- `git diff --check`: passed
- applicable added-file, merge-conflict, private-key, EOF, line-ending, and whitespace pre-commit checks: passed
- Flake8 5.0.4 and autopep8 diff checks on the readiness implementation
- Puppet 6.28 parser validation for `compute.pp`
- a Slurm CUDA canary in `torch-gpu` completed forward propagation, backward propagation, synchronization, BF16 execution, and CPU/GPU parity on an RTX A6000

The repository's legacy `make test` bootstrap cannot complete on this macOS host: its pinned pip fails under Python 3.11, and the current Puppet hook dependency resolution exceeds the system Ruby 2.6 runtime. The changed Puppet manifest was syntax-checked separately with Puppet 6.28 and compatible isolated gem dependencies; OCF CI remains the authoritative full-suite check.

## Suggested follow-ups

1. Run the probe and a framework-level Slurm canary after driver/kernel maintenance.
2. Confirm `nvidia-persistenced` service health and collect NVIDIA Xid events.
3. Export readiness and unexplained-memory state to monitoring.
4. Review a separate Slurm `HealthCheckProgram` or startup gate with explicit drain/resume policy.
5. Define and test a coherent driver/toolkit package policy before changing package pins.


.pre-commit-config.yaml
```diff
@@ -47,6 +47,14 @@ repos:
     rev: v1.9.0
     hooks:
     -   id: reorder-python-imports
+-   repo: local
+    hooks:
+    -   id: ocf-hpc-gpu-readiness-tests
+        name: OCF HPC GPU readiness tests
+        entry: python3 -m unittest discover -s modules/ocf_hpc/tests -p test_*.py
```
dotlambda, line 54: This will fail unless the committer's machine has an NVIDIA GPU. It also makes no sense to run it at commit time rather than at deploy time.

dotlambda (changes requested):

24apricots:
this boils down to what can be done with essentially a package version assertion. a lot of the other checks are trying to cover failure modes that are not possible to check as a pre commit hook; i think this is an incorrect approach and introduces a large amount of brittle python.

PR #1494, opened 2026-09-03
Reload nginx after vhost certificate renewal

oliver-ni:
Reloads nginx when certificates are updated so new sites can be served immediately.

Renewal still runs as `ocfletsencrypt`; systemd handles the nginx reload so we don't need to grant sudo.


sophiebsw (approved):
lgtm

